PR #1, opened 2018-11-23
Make the LottieCompositionReader ignore alpha by default. 

simeoncran:
BodyMovin includes an alpha channel in its colors but the renderers ignore the alpha channel (instead Lottie has opacity on transforms). The alpha channel from BodyMovin seems to contain garbage values, so if we don't ignore it some animations look wrong.

The renderer could also ignore alpha, but seeing as the alpha value appears to be garbage, it is handy to get rid of it in the reader. A new option allows the alpha value to be retained if that's what you need for something.


source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -55,6 +55,18 @@ public enum Options
...
 
+            /// <summary>
+            /// Do not ignore the alpha channel when reading color values.
+            /// </summary>
+            /// <description>
+            /// Lottie files produced by BodyMovin include an alpha channel value that
+            /// is ignored by renderers. By default the <see cref="LottieCompositionReader" />
+            /// will ignore the alpha channel. By enabling this option the alpha channel
+            /// will be read. The result will be that the colors will be what was
+            /// generated by BodyMovin, but not what intended to be rendered.
```
simeoncran: Fix the wording here.

PR #15, opened 2018-12-01
[Do Not Merge] Basic review comments

IbraheemOsama:
Hi,

I found some minor comments in the code but wanted to check the validity of those comments before issuing a PR to fix it.

Please check the changes one by one and reply to the comments you don't see it as valid to not include in my next PR (of course after discussions).

I didn't have any logic comments as it would take ages from me to understand then review the whole project :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Please use [GitHub Issues](https://github.com/windows-toolkit/Lottie-Windows/iss
 
 ## Principles
 
+"Contribution should link to the toolkit contribution guide"
 This project has adopted the code of conduct defined by the [Contributor Covenant](http://contributor-covenant.org/)
 to clarify expected behavior in our community.
 For more information see the [.NET Foundation Code of Conduct](http://dotnetfoundation.org/code-of-conduct).
--- a/source/Lottie/Instantiator.cs
+++ b/source/Lottie/Instantiator.cs
@@ -6,7 +6,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.Linq;//unused reference
 using System.Numerics;
 using Microsoft.Graphics.Canvas.Geometry;
 using Wc = Windows.UI.Composition;
@@ -18,6 +18,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie
     /// Creates instances of a <see cref="Windows.UI.Composition.Visual"/> tree from a description
     /// of the tree.
     /// </summary>
+    /// Missing modifier private
     sealed class Instantiator
     {
         readonly Wc.Compositor _c;
--- a/source/Lottie/LottieVisualDiagnostics.cs
+++ b/source/Lottie/LottieVisualDiagnostics.cs
@@ -44,6 +44,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie
         /// <summary>
         /// Gets a value indicating whether the Lottie is compatible with the current operating system.
         /// </summary>
+        /// SHould be IsCompatibleWithCurrentOs
         public bool IsCompatibleWithCurrentOS { get; internal set; }
 
         public TimeSpan Duration => LottieComposition?.Duration ?? TimeSpan.Zero;
--- a/source/Lottie/LottieVisualSource.cs
+++ b/source/Lottie/LottieVisualSource.cs
@@ -56,9 +56,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie
             RegisterDP<Uri>(nameof(UriSource), null,
             (owner, oldValue, newValue) => owner.HandleUriSourcePropertyChanged(oldValue, newValue));
 
+        // RegisterDp
         static DependencyProperty RegisterDP<T>(string propertyName, T defaultValue) =>
             DependencyProperty.Register(propertyName, typeof(T), typeof(LottieVisualSource), new PropertyMetadata(defaultValue));
 
+        // RegisterDp
         static DependencyProperty RegisterDP<T>(string propertyName, T defaultValue, Action<LottieVisualSource, T, T> callback) =>
             DependencyProperty.Register(propertyName, typeof(T), typeof(LottieVisualSource),
                 new PropertyMetadata(defaultValue, (d, e) => callback((LottieVisualSource)d, (T)e.OldValue, (T)e.NewValue)));
--- a/source/LottieReader/Serialization/LottieCompositionReader.cs
+++ b/source/LottieReader/Serialization/LottieCompositionReader.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
 #endif
     sealed class LottieCompositionReader
     {
+        // all static fields here should start with capital letter
         static readonly AnimatableFloatParser s_animatableFloatParser = new AnimatableFloatParser();
         static readonly AnimatableVector2Parser s_animatableVector2Parser = new AnimatableVector2Parser();
         static readonly AnimatableVector3Parser s_animatableVector3Parser = new AnimatableVector3Parser();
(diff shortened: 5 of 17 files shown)


README.md
```diff
@@ -19,6 +19,7 @@ Please use [GitHub Issues](https://github.com/windows-toolkit/Lottie-Windows/iss
 
 ## Principles
 
+"Contribution should link to the toolkit contribution guide"
```
simeoncran: Please do!

source/Lottie/Instantiator.cs
```diff
@@ -6,7 +6,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.Linq;//unused reference
```
simeoncran: Removing any unused references is goodness. Please do it.

source/Lottie/Instantiator.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie
     /// Creates instances of a <see cref="Windows.UI.Composition.Visual"/> tree from a description
     /// of the tree.
     /// </summary>
+    /// Missing modifier private
```
simeoncran: Don't do this yet. The project currently has a policy of not using redundant visibility modifiers. I welcome a discussion about whether that is a good policy, but let's have that discussion before changing things.

source/Lottie/LottieVisualDiagnostics.cs
```diff
@@ -44,6 +44,7 @@ public string SuggestedClassName
         /// <summary>
         /// Gets a value indicating whether the Lottie is compatible with the current operating system.
         /// </summary>
+        /// SHould be IsCompatibleWithCurrentOs
```
simeoncran: Are you sure? I thought it was ok to use 2 letter acronyms all upper case.

source/Lottie/LottieVisualSource.cs
```diff
@@ -56,9 +56,11 @@ public sealed class LottieVisualSource : DependencyObject, IDynamicAnimatedVisua
             RegisterDP<Uri>(nameof(UriSource), null,
             (owner, oldValue, newValue) => owner.HandleUriSourcePropertyChanged(oldValue, newValue));
 
+        // RegisterDp
```
simeoncran: The name of the method already is that. Feels redundant. Maybe the use of the => form is what is jarring here.

source/WinCompData/Mgc/CanvasDevice.cs
```diff
@@ -15,6 +15,8 @@ sealed class CanvasDevice : IDisposable
     {
         public static CanvasDevice GetSharedDevice() => new CanvasDevice();
 
+        // why empty dispose ?
+        // if there is a reason behind that it should be commented, maybe not to break changes!!
```
simeoncran: I assume there was a reason at one time but no longer. The class doesn't need any of its members any more - we never instantiate it. It probably should be given a private constructor, IDisposable removed, and GetSharedDevice removed. Before the latest version of Win2D, we had to have an instance of a CanvasDevice, but now we always specify null for the CanvasDevice in all the places we use it.

source/WinCompData/Serialization/CompositionObjectXmlSerializer.cs
```diff
@@ -183,6 +183,7 @@ IEnumerable<XObject> GetContents()
 
         XElement FromCompositionPropertySet(CompositionPropertySet obj)
         {
+            // typo in property :)
```
simeoncran: Ack. Please fix.

source/WinCompData/Serialization/CompositionObjectXmlSerializer.cs
```diff
@@ -589,6 +590,7 @@ XElement FromIGeometrySource2D(Wg.IGeometrySource2D obj)
         {
             if (obj is Mgcg.CanvasGeometry)
             {
+                // is there is a reason for having an unused canvasGeometry varaible ?
```
simeoncran: Unfinished code. The XML serializers are just used as debugging tools, and they were written very quickly with bits being added only when I wanted to see a particular part of the tree. Looks like this started to get filled out and then I got distracted.  It can be removed.

source/WinCompData/Serialization/CompositionObjectXmlSerializer.cs
```diff
@@ -722,6 +724,7 @@ IEnumerable<XObject> FromAnimatableScalar(string name, IEnumerable<CompositionOb
                 if (initialValue.HasValue)
                 {
                     yield return FromScalar(name, initialValue.Value);
+                    // yield break will never executed, should be removed.
```
simeoncran: Yep.

source/WinCompData/Serialization/CompositionObjectXmlSerializer.cs
```diff
@@ -757,6 +760,7 @@ IEnumerable<XObject> FromAnimatableVector3(string name, IEnumerable<CompositionO
                 if (initialValue.HasValue)
                 {
                     yield return FromVector3(name, initialValue.Value);
+                    // yield break will never executed, should be removed.
```
simeoncran: Yep.

source/WinCompData/Tools/ApiCompatibility.cs
```diff
@@ -32,6 +32,7 @@ public static ApiCompatibility Analyze(CompositionObject graphRoot)
             // RS4 (geometries are flaky in RS4, and CompositionGeometryClip is new in RS5).
             requiresCompositionGeometricClip = true;
 
+            // just a question why not just passing true ?
```
simeoncran: Yes, this can (and should) be simplified to true. Nice find. The code hung around because it was a handy reference to explain "this is how you check for the existence of a particular type in the tree". My thinking was that for the next release we might have to check for some other type so the code would be there ready to use. Maybe we can keep the code within some #if so it's clear that "this is how to do it if actually checking for the existence of CompositionGeometricClip" but "this is our actual policy because of other issues".

source/WinCompData/Tools/ObjectGraph.cs
```diff
@@ -178,6 +178,7 @@ void Reference(T from, CompositionObject obj)
             }
         }
 
+        // if this is always true why not void ?
```
simeoncran: Ummm... I don't remember exactly. The code is mature enough now that I'd be fine with converting all of the handlers in the visitor to void. I think there were 2 motivations behind having the bool: 1) it's a fairly standard way of writing visitors so that the visit can be stopped, 2) sometimes you really need to pass a non-void-returning delegate to something.

I remember trying to convert the ObjectGraph methods to return void, and giving up on it at one point because of some reason that I don't remember.

BTW, I'm answering this without putting much effort into looking at the class. I suggest trying the change in a separate branch.

source/WinCompData/Tools/Stats.cs
```diff
@@ -159,6 +159,7 @@ public Stats(CompositionObject root)
 
         public int PathGeometryCount => _pathGeometryCount;
 
+        //PropertySetPropertyCount not PropertySetProperyCount
```
simeoncran: Yep.

source/Lottie/Instantiator.cs
IbraheemOsama: Personally, I don't mind either way but just to be consistent with the toolkit coding style.
Maybe we can add an exception for this project but it needs to be documented somewhere :)

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -32,6 +32,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
 #endif
     sealed class LottieCompositionReader
     {
+        // all static fields here should start with capital letter
```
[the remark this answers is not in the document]
IbraheemOsama: It's fine my bad.

source/WinCompData/Tools/ApiCompatibility.cs
IbraheemOsama: It's already part of the file history, I will remove it and when its need we can get it from history :) what do you think?

PR #16, opened 2018-12-01
Sensible sln

simeoncran:
Make the solution more conventional for least surprise.

We'd previously been using MSBuild configurations to filter what was built. This was to allow us to _not_ build everything, but it confused first-timers and turned out not to be necessary. So with this change we're back to normal Debug and Release configurations.


Lottie-Windows/README.md
```diff
@@ -0,0 +1,4 @@
+# Lottie-Windows
+The Lottie-Windows project generates a NuGet package for use by UWP apps.
```
sohchatt: Windows apps

LottieViewer/README.md
```diff
@@ -0,0 +1,8 @@
+﻿# Lottie Viewer
+The Lottie Viewer project generates a simple UWP Windows Store app that renders and displays Lottie .json files. The tool is intended to help designers see how Lottie compositions look when rendered using Lottie-Windows.
```
sohchatt: Microsoft Store, or simply Store

sohchatt: The tool is intended to help designers validate the appearance of Lottie composition rendered using Lottie-Windows.

PR #20, opened 2018-12-05
Report parsing and translation issues as warnings rather than errors,…

simeoncran:
… unless in strict mode.

Unless strict mode is enabled, we'll report parsing and translation issues as warnings rather than errors so as not to scare users
Add a script to easily install LottieGen from code you build locally.
Remove the GlobalSuppressions.cs files that were not supposed to be there.
Fix some code analysis warnings that were being disabled in the GlobalSuppressions.cs files.


sohchatt:
Maybe also update the Icon in the LottieViewer to be "Warnings" rather than "Issues"?

PR #27, opened 2018-12-07
Update version numbers in README to 5.0.0.

simeoncran:
Now that we're aligned with the toolkit major version number update the docs.


LottieGen/README.md
```diff
@@ -8,23 +8,23 @@ LottieGen is built as a [.NET Core global tool](https://docs.microsoft.com/en-us
...
-The latest release version can be installed from NuGet:
+The latest release version can be installed from NuGet (https://www.nuget.org/packages/LottieGen):
 
-    dotnet tool install lottiegen -g
+    dotnet tool install -g LottieGen
 
 Prerelease versions can be installed from NuGet:
 
-    dotnet tool install lottiegen -g --version 1.0.0-prerelease01
+    dotnet tool install -g LottieGen --version 5.0.0-prerelease01
```
azchohfi: We already have 5.0.0-prerelease2

simeoncran: I've updated it to remove the 01. dotnet will try to match the string, which causes it to get the latest one.

PR #28, opened 2018-12-09
Give unique colors to nodes in DGML graph.

simeoncran:
Making it easier to set colors  for DGML graph nodes, and giving each node different colors so that it's easier to tell what kind of geometry is attached to shape, and whether a node has any animations. This will make it easier to find opportunities for optimization.

Fix an assertion in the translator discovered during testing.
Make "percent" a first-class concept for properties in the translator to make maintenance easier.


source/WinCompData/Wui/Colors.cs
```diff
@@ -0,0 +1,151 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Wui
```
sohchatt: Wui -- is that intentional?

simeoncran: Yep. These are objects that mimic those in the Windows.UI namespace, thus Wui. It's the pattern we use to refer to namespaces that we are mimicking.

PR #31, opened 2018-12-21
Add globbing support to LottieGen.

simeoncran:
The main motivation for this is to use LottieGen in tests over large numbers of json files. This is much quicker than starting LottieGen for each file.
Previously we supported wildcards, but that only worked for files in a single directory. Now you can specify wildcards and globstars (**) anywhere in the input. For example C:\foo\**\myfiles\*.json

When specifying .json files with globbing, the outputs will be placed in subdirectories of the output folder corresponding to the subdirectories that were expanded relative to the non-globbed part of the input path. This ensures that if the same input filename exists in multiple places, the LottieGen outputs for each will end up in distinct place.

Also refactored the tool to allow parallel processing, and turned on parallel processing.

I also added a .sln file just for LottieGen to make development faster on my Surface Go which struggles with the main .sln fille.


azchohfi:
You can install a new version of .Net Core. Just use the task in the yml file:
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer?view=vsts

PR #38, opened 2019-01-03
Handle some cases of shape groups that contain multiple paths.

simeoncran:
Use of Win2D geometry groups to group paths that are part of the same shape group. Previously we just drew the paths on top of each other, but that's not correct when the paths overlap. The typical case is concentric shapes where the space between them is used like a stroke.

There are many potential cases that we don't yet handle with this change, e.g. we don't handle animation of the shapes, or other shapes, but the concentric non-animated shapes as a stroke is common enough to be worth doing alone.

### DIFF
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -1209,7 +1209,37 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
                         break;
                     case ShapeContentType.Path:
-                        container.Shapes.Add(TranslatePathContent(context, shapeContext, (Shape)shapeContent));
+                        {
+                            var path = (Shape)shapeContent;
+                            List<Shape> paths = null;
+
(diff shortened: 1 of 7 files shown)


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -1209,7 +1209,37 @@ CompositionShape TranslateGroupShapeContent(TranslationContext context, ShapeCon
 
                         break;
                     case ShapeContentType.Path:
-                        container.Shapes.Add(TranslatePathContent(context, shapeContext, (Shape)shapeContent));
+                        {
+                            var path = (Shape)shapeContent;
+                            List<Shape> paths = null;
```
jverral: Nit: you could initialize the paths here and instead of checking for null, check the count.  Less likely to have a exception creep into the code later when someone forgets to check if its null.

simeoncran: I want to allocate the List<Shape> lazily seeing as most of the time it isn't used. It's also being used as a flag to indicate that there are multiple paths.

Issue #30, opened 2018-12-19
lottie issues/warnings html page


sohchatt:
1. Generate the list of known parsing and translation issues as an html page (builds from the source)
2. Host the resulting page in this repo
3. Update uri in LottieViewer's Issues section with the correct link
4. Update store submission to reflect this change

simeoncran:
@sohchatt  - someone (aka you) has to fill in the the description, remarks, and resources for the .sources/Issues/*.md files.

simeoncran:
I've filled in the docs.

PR #44, opened 2019-01-05
Centralize list of warnings and errors. Generate artifacts from it.

simeoncran:
This is to address Issue #30 .

We need to publish an HTML file that has more details about warnings and errors. But there are 3 different lists of issues (parsing, translation, and validation) and it would be difficult to keep everything in sync.
With this commit we move the data into Issues.xml and add scripts to generate the C# classes and the HTML.

Currently the generation has to be done manually, by:
1) Run GenerateIssuesClasses.cmd to update the C# classes. Requires texttransform.exe from the developer command prompt.
2) Run GenerateIssuesHtml.cmd to regenerate the html


source/Issues/Issues.xml
```diff
@@ -0,0 +1,231 @@
...
+    <method>IllustratorLayers</method>
+  </issue>
+  <issue>
+    <id>LP0008</id>
+    <text>Layer effects</text>
+    <method>LayerEffects</method>
+  </issue>
+  <issue>
+    <id>LP0009</id>
+    <text>Mattes</text>
```
MolusMS: There are some single word help text that could be improved. Although I realize it might not be the goal of this specific change.

simeoncran: This is just to get the infrastructure in. I expect Sohum to want to wordsmith this as we generate a nice looking HTML help file.

PR #58, opened 2019-02-11
Matte and additional mask support part1

eliezerpMS:
This change adds functionality that was dependent on VisualSurface. It has the following:
-Adds WinCompData objects for Matte and additional mask support
-Codegen and WUC object instantiation for Matte and additional mask support


source/WinCompData/Tools/ApiCompatibility.cs
```diff
@@ -14,9 +22,10 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Tools
 #endif
     sealed class ApiCompatibility
     {
-        ApiCompatibility(bool requiresCompositionGeometricClip)
+        ApiCompatibility(bool requiresCompositionGeometricClip, bool requiresCompositionVisualSurface)
```
simeoncran: This constructor that is going to grow and grow feels like unnecessary overhead now. Replace it with an empty constructor, make the properties {get; private set;} and use the initializer syntax in the Analyze method, e.g. new ApiCompatibilty { RequiresCompositionGeometricClip = true, RequiresCompositionVisualSurface = whatever, };

source/Lottie/Instantiator.cs
```diff
@@ -4,10 +4,19 @@
 
 #define ReuseExpressionAnimation
 
+// define POST_RS5_SDK if using an SDK that is for a release
+// after RS5
+#if POST_RS5_SDK
```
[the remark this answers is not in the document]
eliezerpMS: As discussed. I will remove the #defines that are not needed for building.

source/WinCompData/CompositionEffectBrush.cs
```diff
@@ -0,0 +1,25 @@
...
+    public
+#endif
+    sealed class CompositionEffectBrush : CompositionBrush
+    {
+        internal CompositionEffectBrush(Mgce.CompositeEffect effect)
+        {
+            this.Effect = effect;
+        }
+
+        public Dictionary<string, CompositionBrush> SourceParameters { get; } = new Dictionary<string, CompositionBrush>();
```
simeoncran: We should avoid putting Dictionaries on public interfaces. And we should try to make the APIs look as similar as possible to the real Comp APIs. Comp has SetSourceParameter and GetSourceParameter methods. Can you do that instead?

source/WinCompData/Compositor.cs
```diff
@@ -58,5 +58,13 @@ sealed class Compositor
...
 
         public CompositionRoundedRectangleGeometry CreateRoundedRectangleGeometry() => new CompositionRoundedRectangleGeometry();
+
+        public SpriteVisual CreateSpriteVisual() => new SpriteVisual();
+
+        public CompositionVisualSurface CreateVisualSurface() => new CompositionVisualSurface();
+
+        public CompositionSurfaceBrush CreateSurfaceBrush(CompositionSurface surface) => new CompositionSurfaceBrush(surface);
+
+        public CompositionEffectBrush CreateEffectBrush(Mgce.CompositeEffect effect) => new CompositionEffectBrush(effect);
```
simeoncran: I don't think this should be on the Compositor class - it isn't on the real compositor.

PR #64, opened 2019-02-15
Add C++/winrt support for code generation

clarkezone:
This PR adds support for c++/winrt code generation.

While the change works, it has a few todo's and will need to be addressed and also will need some refactoring as there are changes in instantiatorgeneratorbase that are c++/winrt specific which will need abstracting / specializing to avoid breaking c# generation.

It also disables c++/CX code generation (since the output filenames are the same for both) but does not remove the code.  My assumption is that c++/winrt should replace c++/CX codegen, please confirm.

Appreciate an early review.


source/WinCompData/CodeGen/CppStringifier.cs
```diff
@@ -43,9 +43,9 @@ public override string CanvasGeometryCombine(Mgcg.CanvasGeometryCombine value)
             }
```
simeoncran: Rather than breaking this for Cx, can you just create a new version. I'd like CX to continue to work side by side.

LottieViewer/MainPage.xaml.cs
```diff
@@ -55,7 +55,10 @@ async void SaveFile_Click(object sender, RoutedEventArgs e)
 
             // Dropdown of file types the user can save the file as
             filePicker.FileTypeChoices.Add("C#", new[] { ".cs" });
-            filePicker.FileTypeChoices.Add("C++ CX", new[] { ".cpp" });
+
+            // TODO: handle Cx generation side by side with CPPWINRT
```
simeoncran: Testing adding a comment from codeflow tool...

sohchatt:
@clarkezone Any updates?

clarkezone:
TodoList

- [x] Fix hard coded changes in InstantiatorGeneratorBase to be target language neutral
- [ ] Confirm that c++/winrt generation still works
- [ ] Confirm that c# generation still works
- [ ] Make c++/CX work again per code review feedback
- [ ] Address rest of Simeon codereview feedback
- [ ] Rebase / merge to Top of Tree

clarkezone:
LMK when you are ready to review this.. I only intend to rebase once and I only want to do that when you're ready to accept it, hence I'm going to stop work until I hear back.

Issue #66, opened 2019-02-20
5.1.0 nuget package fails to install if project MinVersion < 17763


sohchatt:
Installing Lottie 5.1.0 from nuget.org results in [NU1202](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1202?) when minVersion < 17763 even if  targetVersion is 17763

PR #67, opened 2019-02-22
Fix for #66 - cannot install on projects that have a MinVersion < RS5.

simeoncran:


Lottie-Windows/Lottie-Windows.csproj
```diff
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
-    <TargetFramework>uap10.0.17763</TargetFramework>
+    <TargetFramework>uap10.0.16299</TargetFramework>
```
sohchatt: While the toolkit supports version (N-2), I wonder if there is value in lowering this further to 15063 since all that's feasible on OS versions < 17763 is the AnimatedVisualPlayer.FallbackContent case. Thought?

simeoncran: Hopefully not making support too much harder for us. This is untested, but shouldn't be any worse (i.e. projects that had a min version  < 16299 couldn't even build before, so if it builds and crashes it's not really worse).

PR #68, opened 2019-02-23
Support for new Bodymovin schema. Reader hardening.

simeoncran:
Harden the LottieCompositionReader against layers that have no index. And support new Bodymovin schema.


source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -2076,33 +2081,38 @@ IEnumerable<KeyFrame<T>> ReadKeyFrames(LottieCompositionReader reader, JArray js
                 {
                     var lottieKeyFrame = jsonArray[i].AsObject();
 
-                    // "n" is a name on the keyframe. Never seems to be useful.
+                    // "n" is a name on the keyframe. It is not useful and has been deprecated in Bodymovin.
```
sohchatt: Technically the name of the cubic bezier curve used for easing, not the keyframe.

PR #76, opened 2019-03-08
Masks and mattes part1

simeoncran:


source/LottieReader/Serialization/ParsingIssues.cs
```diff
@@ -44,8 +44,6 @@ internal ParsingIssues(bool throwOnIssue)
 
         internal void LayerEffects() => Report("LP0008", "Layer effects");
 
-        internal void Mattes() => Report("LP0009", "Mattes");
-
         internal void TimeRemappingOfPreComps() => Report("LP0011", "Time remapping of precomp layers");
```
simeoncran: Don't forget the .md files and to update these generated files. They are currently out of sync.

source/WinCompData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1808,7 +1949,7 @@ internal string LongComment
             internal bool RequiresStorage { get; set; }
 
             // Set to indicate that the node relies on Win2D / D2D.
-            internal bool RequiresWin2D => Object is CanvasGeometry;
+            internal bool RequiresWin2D => (Object is CanvasGeometry) || (Object is CompositionEffectBrush);
```
simeoncran: I also mentioned this elsewhere - we need more granular flags than this. Replace RequiresWin2D.

source/WinCompData/CodeGen/Optimizer.cs
```diff
@@ -1132,5 +1255,15 @@ CompositionColorBrush GetCompositionColorBrush(CompositionColorBrush obj)
             StartAnimationsAndFreeze(obj, result);
             return result;
         }
+
+        CompositionSurfaceBase GetCompositionSurface(CompositionSurfaceBase obj)
```
simeoncran: This should follow the if GetExisting(...)/CacheAndInitialize pattern. The goal is that it returns a clone of the object that is cached for future calls.

source/WinCompData/CodeGen/Optimizer.cs
```diff
@@ -801,14 +831,79 @@ CompositionViewBox GetCompositionViewBox(CompositionViewBox obj)
...
+            }
+
+            result = CacheAndInitializeCompositionObject(obj, _c.CreateVisualSurface());
+
+            if (obj.SourceVisual != null)
+            {
+                result.SourceVisual = GetVisual(obj.SourceVisual);
+            }
+
+            result.SourceSize = obj.SourceSize;
```
simeoncran: I don't think you should be checking for null SourceOffset - just copy it. At least be consistent with the way you handle SourceSize and SourceOffset.

source/WinCompData/Mgc/CanvasComposite.cs
```diff
@@ -0,0 +1,26 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Mgc
+{
+#if PUBLIC_WinCompData
+    public
+#endif
+    enum CanvasComposite
+    {
+        SourceOver = 0,
```
simeoncran: Now that we don't care about the values, remove the explicit values.

source/WinCompData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -60,7 +60,17 @@ sealed class CSharpInstantiatorGenerator : InstantiatorGeneratorBase
         // Called by the base class to write the start of the file (i.e. everything up to the body of the Instantiator class).
         protected override void WriteFileStart(CodeBuilder builder, CodeGenInfo info)
         {
-            if (info.RequiresWin2d)
+            if (info.UsesCanvas)
+            {
+                builder.WriteLine("using Microsoft.Graphics.Canvas;");
+            }
+
+            if (info.UsesCanvas)
```
simeoncran: Why have 2 ifs for the same test?

source/WinCompData/CodeGen/Optimizer.cs
simeoncran: Still needs to CacheAndInitialize...

source/WinCompData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -113,6 +113,14 @@ namespace AnimatedVisuals
                 builder.WriteLine("#include <wrl.h>");
             }
 
+            if (info.UsesCanvasEffects ||
+                info.UsesCanvas)
+            {
+                // throw an exception in this case for now. In the future the necessary
+                // C++ code gen will be added
+                throw new InvalidOperationException();
```
simeoncran: Seeing as this is going to fail during testing, add a message to make it clear what is going on to the person who is reading the logs from the failed tests.

source/WinCompData/CodeGen/CSharpInstantiatorGenerator.cs
eliezerpMS: Must have been a copy and paste mistake. I will fix this.

source/WinCompData/CodeGen/Optimizer.cs
eliezerpMS: As discussed offline I will change this function to use a switch statement

source/WinCompData/CodeGen/CxInstantiatorGenerator.cs
sohchatt: @eliezerpMS and a github issue

PR #84, opened 2019-03-14
Update Lottie-Windows.csproj

sohchatt:
Update csproj with description and tags for nuget.org

### DIFF
--- a/Lottie-Windows/Lottie-Windows.csproj
+++ b/Lottie-Windows/Lottie-Windows.csproj
@@ -1,6 +1,14 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
+    
+    <!-- Description -->
+    <Title>Lottie-Windows</Title>
+    <Description>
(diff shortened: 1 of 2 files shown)


Lottie-Windows/Lottie-Windows.csproj
```diff
@@ -1,6 +1,14 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
+    
+    <!-- Description -->
+    <Title>Lottie-Windows</Title>
+    <Description>
+      Render Adobe AfterEffects animations natively on Windows: aka.ms/lottie
```
michael-hawker: I'd put the full link so that it's more likely to be clickable depending on client.

```suggestion
      Render Adobe AfterEffects animations natively on Windows: https://aka.ms/lottie
```

michael-hawker:
@sohchatt can you update the version.json [here](https://github.com/windows-toolkit/Lottie-Windows/blob/941ce5e60f369d43547306230395d5e6250ee94c/version.json#L2) to just be `5.1.1` as well?  Then I won't have to do a separate PR for that too. :)

PR #101, opened 2019-04-19
Adding some of the pieces needed for supporting image assets

simeoncran:
Support for embedded images in Lottie files.
Introducing the WinUIXamlMediaData project so we can bring in LoadedImageSource for loading images.


source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -452,6 +452,71 @@ Asset ParseAsset(JsonReader reader)
             throw EofException;
         }
 
+        static ImageAsset CreateImageAsset(string id, double width, double height, string imagePath, string fileName)
```
eliezerpMS: >static [](start = 8, length = 6)

Just curious, but why does this need to be static?

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -452,6 +452,71 @@ Asset ParseAsset(JsonReader reader)
...
+        {
+            // Colon is never valid for a file name. If fileName contains a colon it is probably a URL.
+            var colonIndex = fileName.IndexOf(':');
+            return string.IsNullOrWhiteSpace(imagePath) && colonIndex > 0
```
eliezerpMS: >string.IsNullOrWhiteSpace(imagePath) && colonIndex > 0 [](start = 19, length = 54)

Is this line doing a 'is this a valid filepath or url' check? If so is there a build in API for that?

source/LottieReader/Serialization/LottieCompositionReader.cs
simeoncran: Doesn't need to be, but I like to make methods static whenever they can be. It tells the reader that it does not depend on any object state, which helps local reasoning.

---
In reply to: [277075711](https://github.com/windows-toolkit/Lottie-Windows/pull/101#discussion_r277075711) [](ancestors = 277075711)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -156,6 +156,31 @@ YamlObject FromLayersAsset(LayerCollectionAsset asset)
...
+
+        YamlObject FromEmbeddedImageAsset(EmbeddedImageAsset asset)
+        {
+            return new YamlMap
+            {
+                { "Id", asset.Id },
+                { "Width", asset.Width },
+                { "Height", asset.Height },
+                { "Format", asset.Format},
+                { "Bytes", asset.Bytes.Length},
```
plaiMS: Should these be imageAsset.Id, imageAsset.Width, etc?

plaiMS:
>                 { "Filename", asset.FileName },

Same here -- imageAsset.Id...?

---
Refers to: source/LottieData/Serialization/LottieCompositionYamlSerializer.cs:191 in fe27c03. [](commit_id = fe27c03245f15f61d2f231e852f94d15a1d14a15, deletion_comment = False)

source/WinCompData/CodeGen/Optimizer.cs
```diff
@@ -1253,12 +1253,14 @@ CompositionColorBrush GetCompositionColorBrush(CompositionColorBrush obj)
...
+        ICompositionSurface GetCompositionSurface(ICompositionSurface obj)
         {
-            switch (obj.Type)
+            switch (obj.TypeName)
             {
-                case CompositionObjectType.CompositionVisualSurface:
+                case nameof(CompositionVisualSurface):
                     return GetCompositionVisualSurface((CompositionVisualSurface)obj);
+
+                case "LoadedImageSurface": // Not yet implemented.
```
plaiMS: >"LoadedImageSurface" [](start = 21, length = 20)

why not use nameof(LoadedImageSurface)?

source/WinCompData/Tools/ObjectGraph.cs
```diff
@@ -223,6 +223,20 @@ void Reference(T from, CompositionObject obj)
...
 
+        bool Reference(T from, ICompositionSurface obj)
+        {
+            switch (obj.TypeName)
+            {
+                case nameof(CompositionVisualSurface):
+                    Reference(from, (CompositionObject)obj);
+                    return true;
+
+                case "LoadedImageSurface": // Not yet implemented.
```
plaiMS: >"LoadedImageSurface" [](start = 21, length = 20)

why not use nameof(LoadedImageSurface)?

source/LottieData/Serialization/LottieCompositionXmlSerializer.cs
```diff
@@ -143,14 +143,38 @@ XElement FromLayersAsset(LayerCollectionAsset asset)
...
-                new XAttribute(nameof(asset.Id), asset.Id),
-                new XAttribute(nameof(asset.Width), asset.Width),
-                new XAttribute(nameof(asset.Height), asset.Height),
-                new XAttribute(nameof(asset.Path), asset.Path),
-                new XAttribute(nameof(asset.FileName), asset.FileName));
+                new XAttribute(nameof(imageAsset.Id), imageAsset.Id),
+                new XAttribute(nameof(imageAsset.Width), imageAsset.Width),
+                new XAttribute(nameof(imageAsset.Height), imageAsset.Height),
+                new XAttribute(nameof(imageAsset.Format), imageAsset.Format),
+                new XAttribute(nameof(imageAsset.Bytes), imageAsset.Bytes.Length));
```
plaiMS: >nameof(imageAsset.Bytes), imageAsset.Bytes.Length [](start = 31, length = 49)

Length of byte array is outputted for the "Bytes" attribute. That's misleading.

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
plaiMS: >"Bytes", asset.Bytes.Length [](start = 18, length = 27)

"BytesArrayLength" to be exact?

source/WinCompData/Tools/ObjectGraph.cs
simeoncran: Can't do that yet because it would create a dependency from WinCompData to WinUIXamlMediaData. After the next refactoring we can fix this.

---
In reply to: [277395826](https://github.com/windows-toolkit/Lottie-Windows/pull/101#discussion_r277395826) [](ancestors = 277395826)

source/WinCompData/CodeGen/Optimizer.cs
simeoncran: Can't do that yet because it would create a dependency from WinCompData to WinUIXamlMediaData. After the next refactoring we can fix this.

---
In reply to: [277395773](https://github.com/windows-toolkit/Lottie-Windows/pull/101#discussion_r277395773) [](ancestors = 277395773)

PR #102, opened 2019-04-23
Adding a new project UIData to reference WinUIXamlMediaData.

plaiMS:
This is the first part of a refactoring that will enable code to reference the new WinUIXamlMediaData namespace. Adding a new project UIData to reference WinUIXamlMediaData and will eventually move all code under source/WinCompData that needs to reference WinUIXamlMediaData to UIData.

In this PR, UIData includes ObjectGraph.cs and Graph.cs which is a copy of the ones in source/WinCompData/Tools, except that the UIData/Tools/[Object]Graph.cs has the new UIData namespace and it references WinUIXamlMediaData. 

In the next PR, the original ObjectGraph.cs and Graph.cs under source/WinCompData/Tools will be deleted and all necessary files will be moved to UIData.


LottieGen/LottieGen.csproj
```diff
@@ -37,6 +37,7 @@
   <Import Project="..\source\LottieToWinComp\LottieToWinComp.projitems" Label="Shared" />
   <Import Project="..\source\WinCompData\WinCompData.projitems" Label="Shared" />
   <Import Project="..\source\WinUIXamlMediaData\\WinUIXamlMediaData.projitems" Label="Shared" />
+  <Import Project="..\source\UIData\\UIData.projitems" Label="Shared" />
```
simeoncran: Can you reorder the Imports alphabetically? I know they weren't like that before, but now that we have so many we should do it.

LottieGen/LottieGen.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28315.86
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.572
```
sohchatt: @simeoncran Does this have any actual impact on the sln?

LottieGen/LottieGen.csproj
plaiMS: Reordered.

plaiMS: reordered

---
In reply to: [277889904](https://github.com/windows-toolkit/Lottie-Windows/pull/102#discussion_r277889904) [](ancestors = 277889904)

LottieGen/LottieGen.sln
simeoncran: Don't think so. Should be fine. I think it got bumped to 16 as a result of opening it on VS2019 one time.

PR #105, opened 2019-04-29
Adding first pieces needed in LottieToWinCompTranslator to enable image layer.

plaiMS:
Changes protected under #define EnableImageLayer.


simeoncran:
> //     This code was generated by a tool.

This code is generated.... don't modify it directly. The source is in source\Issues. Add a new .md file there for LT0028.md, then run the GenerateISsuesClasses.cmd script which will regenerate this file.

---
Refers to: source/LottieToWinComp/TranslationIssues.cs:7 in dc91f49. [](commit_id = dc91f49d1e84794ed7cfc938e5b4ee25006c240b, deletion_comment = False)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -651,10 +667,37 @@ CompositionObject TranslateLayer(TranslationContext context, Layer layer)
...
+            var content = CreateSpriteVisual();
+            containerVisualContentNode.Children.Add(content);
+
+            var referencedLayersAsset = _lc.Assets.GetAssetById(layer.RefId);
+            switch (referencedLayersAsset.Type)
+            {
+                case Asset.AssetType.Image:
+                    var imageAsset = (ImageAsset)referencedLayersAsset;
+                    content.Size = new Sn.Vector2((float)imageAsset.Width, (float)imageAsset.Height);
+                    var brush = CreateColorBrush(LottieData.Color.FromArgb(1, 1, 0, 0));
```
sohchatt: maybe a comment that explains the intentions of using a colorbrush?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -694,7 +737,16 @@ Visual TranslatePreCompLayerToVisual(TranslationContext context, PreCompLayer la
                     AddTranslatedLayersToContainerVisual(contentsNode, subContext, referencedLayers, $"{layer.Name}:{layerCollectionAsset.Id}");
                     break;
                 case Asset.AssetType.Image:
+#if EnableImageLayer
```
sohchatt: Why does this need to be done twice? Shouldn't this case be handled by regular translation above?

simeoncran:
> [comment]: # (text:Image layers is not supported)

Remove the s or make it "Image layers are not supported".

---
Refers to: source/Issues/LT0009.md:5 in b43759e. [](commit_id = b43759e861236e133249c0e93320f2271807d372, deletion_comment = False)

simeoncran:
> [comment]: # (text:ReferencedAssetDoesNotExist)

This shouldn't be smoooshed together.

---
Refers to: source/Issues/LT0028.md:5 in b43759e. [](commit_id = b43759e861236e133249c0e93320f2271807d372, deletion_comment = False)

simeoncran:
> [comment]: # (text:{currentLayerType} referenced asset {assetRefId} of type {assetType} which is invalid. Expected asset of type {expectedAssetType}.)

...Expected an asset...

---
Refers to: source/Issues/LT0029.md:5 in b43759e. [](commit_id = b43759e861236e133249c0e93320f2271807d372, deletion_comment = False)

simeoncran:
> # Lottie-Windows Warning LT0004

Wrong number

---
Refers to: source/Issues/LT0029.md:9 in b43759e. [](commit_id = b43759e861236e133249c0e93320f2271807d372, deletion_comment = False)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -250,6 +254,15 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
 
+                        yield return (Visual)item;
+                        break;
+                    case CompositionObjectType.SpriteVisual:
+                        if (shapeVisual != null)
+                        {
+                            yield return shapeVisual;
+                            shapeVisual = null;
+                        }
+
```
simeoncran: I don't understand what this code is intended to achieve. It looks like you'd get the same result if you just added SpriteVisual to the list of cases in the previous case (case ContainerVisual: case ShapeVisual: case SpriteVisual:).

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -273,25 +286,25 @@ void TranslateAndApplyMask(TranslationContext context, Layer layer, Visual visua
...
-                    _unsupported.MaskWithUnsupportedMode(mask.Mode.ToString());
+                    _issues.MaskWithUnsupportedMode(mask.Mode.ToString());
                 }
 
                 // Translation currently does not support having multiple paths for masks.
                 // If possible users should combine masks when exporting to json.
                 if (layer.Masks.Length > 1)
                 {
-                    _unsupported.MultipleShapeMasks();
+                    _issues.MultipleShapeMasks();
```
simeoncran: >MultipleShapeMasks [](start = 28, length = 18)

This is an unsupported case (as per the comments) so that name of the issue should be updated to reflect that. Should be done for all the issue types that are saying that things are unsupported.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -651,10 +664,45 @@ CompositionObject TranslateLayer(TranslationContext context, Layer layer)
...
+                return null;
+            }
+
+            var content = CreateSpriteVisual();
+            containerVisualContentNode.Children.Add(content);
+
+            var referencedLayersAsset = _lc.Assets.GetAssetById(layer.RefId);
+            if (referencedLayersAsset == null)
+            {
+                _issues.ReferencedAssetDoesNotExist();
```
simeoncran: >ReferencedAssetDoesNotExist [](start = 24, length = 27)

Because this is done twice now, it should be moved to a method on this class: GetAssetById(...) that either return the asset or reports the problem and returns null.

source/LottieToWinComp/LottieToWinCompTranslator.cs
plaiMS: Pretty much all of them are issues because they are unsupported. They might look silly if all issues say "NotSupported"?

---
In reply to: [279590031](https://github.com/windows-toolkit/Lottie-Windows/pull/105#discussion_r279590031) [](ancestors = 279590031)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Comment should explain the intention of the brush. There should be comments in this method explaining why you're doing what you're doing - the brush is not the only weird thing in here.

---
In reply to: [279561955](https://github.com/windows-toolkit/Lottie-Windows/pull/105#discussion_r279561955) [](ancestors = 279561955)

plaiMS: What are the other weird things here?

---
In reply to: [279590537](https://github.com/windows-toolkit/Lottie-Windows/pull/105#discussion_r279590537) [](ancestors = 279590537,279561955)

source/LottieToWinComp/LottieToWinCompTranslator.cs
plaiMS: Good call.

---
In reply to: [279589809](https://github.com/windows-toolkit/Lottie-Windows/pull/105#discussion_r279589809) [](ancestors = 279589809)

source/LottieToWinComp/LottieToWinCompTranslator.cs
plaiMS: I went with GetAssetById() without the type checking because I think that makes the code easier to follow.

---
In reply to: [279590241](https://github.com/windows-toolkit/Lottie-Windows/pull/105#discussion_r279590241) [](ancestors = 279590241)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -703,6 +740,18 @@ Visual TranslatePreCompLayerToVisual(TranslationContext context, PreCompLayer la
             return result;
         }
 
+        internal Asset GetAssetById(string assetId)
```
simeoncran: I don't think needs to be internal.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -682,7 +718,8 @@ Visual TranslatePreCompLayerToVisual(TranslationContext context, PreCompLayer la
 #endif
 
             // TODO - the animations produced inside a PreComp need to be time-mapped.
-            var referencedLayersAsset = _lc.Assets.GetAssetById(layer.RefId);
+            var referencedLayersAsset = GetAssetById(layer.RefId);
+
             switch (referencedLayersAsset.Type)
```
simeoncran: >referencedLayersAsset [](start = 20, length = 21)

What if GetAssetById returns null?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -703,6 +740,18 @@ Visual TranslatePreCompLayerToVisual(TranslationContext context, PreCompLayer la
             return result;
         }
 
+        internal Asset GetAssetById(string assetId)
+        {
+            var referencedAsset = _lc.Assets.GetAssetById(assetId);
+            if (referencedAsset == null)
+            {
+                _issues.ReferencedAssetDoesNotExist(assetId);
+                return null;
```
simeoncran: > return null; [](start = 15, length = 13)

This return is redundant, and unnecessarily violates the principle of avoiding multiple exit points in a method.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -651,10 +656,41 @@ CompositionObject TranslateLayer(TranslationContext context, Layer layer)
...
+                // The layer is never visible.
+                return null;
+            }
+
+            var content = CreateSpriteVisual();
+            containerVisualContentNode.Children.Add(content);
+
+            var referencedLayersAsset = GetAssetById(layer.RefId);
+
+            switch (referencedLayersAsset.Type)
```
simeoncran: >referencedLayersAsset [](start = 20, length = 21)

What is GetAssetById returned null?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -651,10 +656,41 @@ CompositionObject TranslateLayer(TranslationContext context, Layer layer)
...
+                    var imageAsset = (ImageAsset)referencedLayersAsset;
+                    content.Size = new Sn.Vector2((float)imageAsset.Width, (float)imageAsset.Height);
+
+                    // For prototyping purpose, adding a red brush as a placeholder for where an image asset should be drawn.
+                    // TODO - remove red brush and support image asset.
+                    var brush = CreateColorBrush(LottieData.Color.FromArgb(1, 1, 0, 0));
+                    content.Brush = brush;
+                    break;
+                default:
+                    _issues.InvalidAssetReferenceFromCurrentLayer(layer.Type.ToString(), layer.RefId, referencedLayersAsset.Type.ToString(), Asset.AssetType.Image.ToString());
```
simeoncran: The word "Current" doesn't belong here. Current only makes sense in the context of this method. When people see the error they don't have a concept of "current".

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -694,7 +731,7 @@ Visual TranslatePreCompLayerToVisual(TranslationContext context, PreCompLayer la
                     AddTranslatedLayersToContainerVisual(contentsNode, subContext, referencedLayers, $"{layer.Name}:{layerCollectionAsset.Id}");
                     break;
                 case Asset.AssetType.Image:
-                    _unsupported.ImageAssets();
+                    _issues.InvalidAssetReferenceFromCurrentLayer(layer.Type.ToString(), layer.RefId, referencedLayersAsset.Type.ToString(), Asset.AssetType.LayerCollection.ToString());
```
simeoncran: >InvalidAssetReferenceFromCurrentLayer [](start = 28, length = 37)

Move this into the default and remove the InvalidOperationException(). We don't need to test for Image here.

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: I don't understand how it makes the code harder to follow. Maybe it's about choosing a better name for GetAssetById. Or maybe it's just a matter of using named parameters to make it clear from the caller: GetAssetById(id: layer.RefId, expectedAssetType: Asset.AssetType.Image);
But I do see that the Image and Precomp layers now have slightly different behavior when given the wrong type of asset, so we already have introduced a bug because we've violated DRY.

---
In reply to: [279607824](https://github.com/windows-toolkit/Lottie-Windows/pull/105#discussion_r279607824) [](ancestors = 279607824,279590241)

simeoncran:
> # Lottie-Windows Warning LT0008

This file is showing up in CodeFlow as a rename of LT0008.md. That means LT0008.md not longer exists. That's suspicious. I think you've done this by accident, and LT00028 is not a new error.

---
Refers to: source/Issues/LT0008.md:9 in b44898e. [](commit_id = b44898ec22e622ee9e31776b4924e3ec56f2eef9, deletion_comment = True)

plaiMS:
> # Lottie-Windows Warning LT0008

LT0008 is the ImageAsset() warning that we said no longer needed and can remove. Somehow Codeflow is treating LT0028 as rename of the LT0008 but it's not.

---
In reply to: [488055213](https://github.com/windows-toolkit/Lottie-Windows/pull/105#issuecomment-488055213) [](ancestors = 488055213)

---
Refers to: source/Issues/LT0008.md:9 in b44898e. [](commit_id = b44898ec22e622ee9e31776b4924e3ec56f2eef9, deletion_comment = True)

PR #106, opened 2019-04-30
Multi-shape mask and matte translation

eliezerpMS:
This change adds the translation logic for supporting masks with multiple shapes and mattes. Both of these use the VisualSurface. The single shape mask case that used GeometryClip has been removed.


source/Lottie/Instantiator.cs
```diff
@@ -879,26 +879,23 @@ Wc.CompositionEffectBrush GetCompositionEffectBrush(Wd.CompositionEffectBrush ob
                 return result;
             }
 
+            var compositeEffect = new Mgce.CompositeEffect();
```
eliezerpMS: I put this back to no longer have the switch statements. The issues was that we need to call CreateEffectFactory after populating the compositeEffect. I would be ok with making this function have 2 switch statements instead. One for before the CreateEffectFactory for the composite effect initialization and one for the compositeEffectBrush initialization

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,226 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+        {
+            Mask.MaskMode maskMode = Mask.MaskMode.None;
 
-                // Add a mask as a clip path if it can support the mask properties
-                if (!mask.Inverted &&
-                    mask.Opacity.InitialValue == 100 &&
-                    !mask.Opacity.IsAnimated &&
-                    mask.Mode == Mask.MaskMode.Additive &&
-                    layer.Masks.Length == 1)
+#if AllowVisualSurface
```
eliezerpMS: Forgot that we by default assumed that we would run on an OS that supports geometric clip and so the "masks are not supported" issue is not used. I will put this back in here in the event "layer.Masks.Any()" and we do not allow for VisualSurface

plaiMS:
> # Lottie-Windows Warning LT0030

Did you pull from Master? I added LT0030 2 weeks ago. 

---
Refers to: source/Issues/LT0030.md:9 in ab660f0. [](commit_id = ab660f07a81b4e686728977ce90a206047e4e629, deletion_comment = False)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3431,6 +3583,21 @@ ExpressionAnimation CreateExpressionAnimation(Expr expression)
...
 
+        CompositionVisualSurface CreateVisualSurface()
+        {
+            return _c.CreateVisualSurface();
+        }
+
+        CompositionSurfaceBrush CreateSurfaceBrush(ICompositionSurface surface)
+        {
+            return _c.CreateSurfaceBrush(surface);
+        }
```
plaiMS: This was added and now in master.

source/Lottie/Instantiator.cs
```diff
@@ -880,19 +880,40 @@ Wc.CompositionEffectBrush GetCompositionEffectBrush(Wd.CompositionEffectBrush ob
...
+                    compositeEffect.Mode = CanvasComposite(((Wd.Mgce.CompositeEffect)obj.GetEffect()).Mode);
+                    var wdCompositeEffect = (Wd.Mgce.CompositeEffect)effect;
+                    foreach (var source in wdCompositeEffect.Sources)
+                    {
+                        compositeEffect.Sources.Add(new Wc.CompositionEffectSourceParameter(source.Name));
+                    }
+
+                    break;
+                default:
+                    throw new InvalidOperationException();
```
plaiMS: > throw new InvalidOperationException(); [](start = 19, length = 39)

Return here, so you don't need to repeat the exact same switch case below.

source/Lottie/Instantiator.cs
```diff
@@ -880,19 +880,40 @@ Wc.CompositionEffectBrush GetCompositionEffectBrush(Wd.CompositionEffectBrush ob
...
+                case Wd.Mgce.GraphicsEffectType.CompositeEffect:
+                    var compositeEffectBrush = effectFactory.CreateBrush();
+                    result = CacheAndInitializeCompositionObject(obj, compositeEffectBrush);
                     var wdCompositeEffect = (Wd.Mgce.CompositeEffect)effect;
-                    var wcCompositeEffect = (Mgce.CompositeEffect)graphicsEffect;
                     foreach (var source in wdCompositeEffect.Sources)
                     {
-                        wcCompositeEffect.Sources.Add(new Wc.CompositionEffectSourceParameter(source.Name));
                         result.SetSourceParameter(source.Name, GetCompositionBrush(obj.GetSourceParameter(source.Name)));
                     }
```
plaiMS: Why don't you move these lines into the above switch case?

simeoncran:
> 

For new .md files, please save them as UTF8 with no BOM so that they look good in CodeFlow.

---
Refers to: source/Issues/LT0031.md:4 in ab660f0. [](commit_id = ab660f07a81b4e686728977ce90a206047e4e629, deletion_comment = False)

source/WinCompData/Compositor.cs
```diff
@@ -66,5 +66,7 @@ sealed class Compositor
         public CompositionSurfaceBrush CreateSurfaceBrush(ICompositionSurface surface) => new CompositionSurfaceBrush(surface);
 
         public CompositionEffectFactory CreateEffectFactory(Mgce.GraphicsEffectBase effect) => new CompositionEffectFactory(effect);
+
+        public CompositionEffectBrush CreateEffectBrush(Mgce.GraphicsEffectBase effect) => new CompositionEffectBrush(effect);
```
simeoncran: This doesn't seem to be a method on the real compositor. Am I missing something?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -19,6 +19,14 @@
 // property without going through a controller. Enable this to prevent flashes.
 #define ControllersSynchronizationWorkaround
 
+// define POST_RS5_SDK if using an SDK that is for a release
+// after RS5
+#if POST_RS5_SDK
```
simeoncran: >POST_RS5_SDK [](start = 4, length = 12)

The 10.0.18362.0 SDK is out now. Do we need this #define any more?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -222,13 +232,41 @@ void Translate()
         }
 
         // Takes a list of Visuals and Shapes and returns a list of Visuals.
-        IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnumerable<CompositionObject> items)
+        IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnumerable<(CompositionObject translatedLayer, Layer layer)> items)
```
simeoncran: Can we change this to ShapeOrVisual to make the intention a bit clearer. Will need updating in the LINQ query that creates this list.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -222,13 +232,41 @@ void Translate()
...
         {
             ShapeVisual shapeVisual = null;
```
simeoncran: This method is really hard for me to read - I don't understand all the magic. It was bad enough before, but at least the rule was simple - share a ShapeVisual if you can. But now I don't know how to explain this code. Can it be refactored or commented to make it clearer?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
             }
         }
 
-        void TranslateAndApplyMask(TranslationContext context, Layer layer, Visual visualForMask)
+        // Helper for shape trees that need a mask applied
```
simeoncran: End comments with a period.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+        Visual TranslateAndApplyMasksOnShapeTree(
```
simeoncran: This method needs a lot more explanation.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+            var contextSize = Vector2(context.Width, context.Height);
+
+            var contentShapeVisual = CreateShapeVisual();
+            contentShapeVisual.Size = contextSize;
+            contentShapeVisual.Shapes.Add(containerShapeToMask);
+
+            return TranslateAndApplyMasks(context, layer, contentShapeVisual);
+        }
+
+        Mask.MaskMode TranslateAndAddMaskPaths(
```
simeoncran: Needs comments.
Does it translate layer and put it in maskContainerShape? Is this a layer that was previously translated and is being translated again?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                        }
+
+                        if (mask.Mode == maskMode)
+                        {
+                            var maskSpriteShape = CreateSpriteShape();
+                            maskSpriteShape.Geometry = compositionPathGeometry;
+
+                            // The mask geometry needs to be colored with something so that it can be used
+                            // as a mask.
+                            maskSpriteShape.FillBrush = CreateColorBrush(LottieData.Color.Black);
```
simeoncran: >Black [](start = 90, length = 5)

Is this color used later on? If so, can it be given a name so it's obvious how they're related?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                            geometry.InitialValue,
+                            SolidColorFill.PathFillType.EvenOdd,
+                            optimizeLines: true);
+
+                        if (geometry.IsAnimated)
+                        {
+                            ApplyPathKeyFrameAnimation(context, geometry, SolidColorFill.PathFillType.EvenOdd, compositionPathGeometry, "Path", "Path", null);
+                        }
+
+                        if (maskMode == Mask.MaskMode.None)
```
simeoncran: Looks like this can change on each iteration. currentMaskMode?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+        // using the VisualSurface. The VisualSurface can take the rendered contents of a visual tree and
+        // use it as a brush. The final masked result is achieved by taking the visual to be masked, putting
+        // it into a VisualSurface, then taking the mask and putting that in a VisualSurface and then combining
+        // the result with a composite effect.
+        Visual TranslateAndApplyMasks(
+            TranslationContext context,
+            Layer layer,
+            Visual visualToBeMasked)
+        {
+            if (layer.Masks != null &&
```
simeoncran: >layer.Masks != [](start = 16, length = 14)

Can we give Layer a default empty list of masks so we never have to check it for null?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+        // A visual tree is turned into a brush by using the CompositionVisualSurface.
+        CompositionObject TranslateMatteLayer(
             TranslationContext context,
-            Layer layer,
-            CompositionContainerShape containerShape,
-            ContainerVisual contentContainerVisual,
-            ContainerVisual rootContainerVisual)
+            Visual matteLayer,
+            Visual mattedLayer,
+            bool invert = false)
```
simeoncran: >false [](start = 26, length = 5)

Avoid optional parameters on private methods. It requires the reader at the callsite to have to guess at what the default behavior is.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
-            //            |
-            //     +----------------+
-            //     | containerShape |
-            //     +----------------+
-            var shapeVisual = CreateShapeVisual();
-            shapeVisual.Shapes.Add(containerShape);
-            shapeVisual.Size = Vector2(context.Width, context.Height);
+            // Calculate the context size which we will use as the size of the images we want to use
+            // for the matte content and the content to be matted.
+            var contextSize = Vector2(context.Width, context.Height);
```
simeoncran: >ector2(context.Width, context.Height); [](start = 31, length = 38)

I've noticed this code a few times. It should be put into the TranslationContext so it doesn't get repeated.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+            var mattedLayerVisualSurface = CreateVisualSurface();
+            mattedLayerVisualSurface.SourceVisual = mattedLayer;
+            mattedLayerVisualSurface.SourceSize = contextSize;
+            var mattedSurfaceBrush = CreateSurfaceBrush(mattedLayerVisualSurface);
+
+            return CompositeVisuals(
+                        matteLayer,
+                        mattedLayer,
+                        contextSize,
+                        Sn.Vector2.Zero, invert ? CanvasComposite.DestinationOut : CanvasComposite.DestinationIn);
```
simeoncran: >invert ? CanvasComposite.DestinationOut : CanvasComposite.DestinationI [](start = 41, length = 70)

The final argument should be on a separate line, to be consistent with the other arguments.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+        }
+
+        SpriteVisual CompositeVisuals(
```
simeoncran: Needs a comment explaining the intention.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -268,121 +334,229 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+            Visual source,
+            Visual destination,
+            Sn.Vector2 size,
+            Sn.Vector2 offset,
+            CanvasComposite compositeMode)
+        {
+            // There is a bug where if the root visual is set as a
```
simeoncran: >bug [](start = 26, length = 3)

Bug in Composition? Make that clear. Is there any chance the bug will be fixed one day? If so you should mention which versions are known to have the problem.

source/Lottie/Instantiator.cs
eliezerpMS: This switch statement requires effectFactory but effectFactory requires the first switch statement so I need to break them up.

---
In reply to: [283529598](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283529598) [](ancestors = 283529598)

plaiMS: The two switches both switch on "effect.Type" which I don't see how it's related to effectFactory. Am I missing something?

---
In reply to: [285746240](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r285746240) [](ancestors = 285746240,283529598)

source/WinCompData/Compositor.cs
eliezerpMS: Not sure why I added this. I will remove it and use CreateEffectFactory

---
In reply to: [283918629](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283918629) [](ancestors = 283918629)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: We need to also update the build system to use the latest sdk as well.

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: I will add a comment.

---
In reply to: [283925509](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283925509) [](ancestors = 283925509)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: This color is not used later on. I just need a color that is not transparent.

---
In reply to: [283925078](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283925078) [](ancestors = 283925078)

eliezerpMS:
> # Lottie-Windows Warning LT0030

I have not pulled master since submitting this PR. After this iteration I will look to do that.

---
In reply to: [491966919](https://github.com/windows-toolkit/Lottie-Windows/pull/106#issuecomment-491966919) [](ancestors = 491966919)

---
Refers to: source/Issues/LT0030.md:9 in ab660f0. [](commit_id = ab660f07a81b4e686728977ce90a206047e4e629, deletion_comment = False)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: I can't make any promises on when this bug will be fixed but I can call out when it was first introduced.

---
In reply to: [283929978](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283929978) [](ancestors = 283929978)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: As discussed offline. I will just spit this function and have another function that does the matte work

---
In reply to: [283922427](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283922427) [](ancestors = 283922427)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: As discussed offline I just added a github issue. We should look into switching the build system and default SDK over to the newest release. However, we also probably need to be able to surface issues when a user tries to run on an older Windows build.

---
In reply to: [283919302](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r283919302) [](ancestors = 283919302)

source/LottieToWinComp/TranslationContext.cs
```diff
@@ -42,6 +43,8 @@ abstract class TranslationContext
 
         internal double Height { get; private set; }
 
+        internal System.Numerics.Vector2 Dimensions { get; private set; }
```
simeoncran: Do a "using System.Numerics" rather than fully qualifying.

source/LottieToWinComp/TranslationContext.cs
```diff
@@ -29,6 +29,7 @@ abstract class TranslationContext
             DurationInFrames = lottieComposition.OutPoint - lottieComposition.InPoint;
             Width = lottieComposition.Width;
             Height = lottieComposition.Height;
+            Dimensions = new System.Numerics.Vector2((float)Width, (float)Height);
```
simeoncran: How is "Dimensions" different from Width and Height? Needs comments on all 3 so that it's obvious.
Also, in general I've been avoiding storing floating point values in 32 bit fields - we might as well keep the highest precision during translation, so maybe this should be some other data structure, or just 2 doubles.

source/LottieData/Layer.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
 #endif
     abstract class Layer : LottieObject
     {
-        readonly Mask[] _masks;
+        readonly Mask[] _masks = { };
```
simeoncran: This will allocate a new array on every instantiation, and then it may get overwritten with yet another array if args.Masks != null. 
Instead you should have: 
    static readonly Mask[] s_emptyMasks = { };
    readonly Mask[] _masks = s_emptyMasks;

That way we allocate the empty one only onc.e

source/Lottie/Instantiator.cs
```diff
@@ -813,9 +813,9 @@ Wc.CompositionVisualSurface GetCompositionVisualSurface(Wd.CompositionVisualSurf
                 result.SourceVisual = GetVisual(obj.SourceVisual);
             }
 
-            result.SourceSize = obj.SourceSize;
+            result.SourceSize = obj.SourceSize.Value;
```
simeoncran: >SourceSize [](start = 36, length = 10)

You need to check that obj.SourceSize and obj.SourceOffset have values, otherwise the .Value calls will throw.

source/Lottie/Instantiator.cs
```diff
@@ -884,19 +884,27 @@ Wc.CompositionEffectBrush GetCompositionEffectBrush(Wd.CompositionEffectBrush ob
...
                     {
-                        wcCompositeEffect.Sources.Add(new Wc.CompositionEffectSourceParameter(source.Name));
+                        compositeEffect.Sources.Add(new Wc.CompositionEffectSourceParameter(source.Name));
+                    }
+
+                    // Create the EffectFactory.
+                    // The IGraphicsEffect must be fully initialized and populated before calling CreateEffectFactory.
+                    var effectFactory = _c.CreateEffectFactory(compositeEffect);
+
+                    // Initialize the brush
```
simeoncran: Periods on the ends of all sentences.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -16,6 +16,14 @@
 // property without going through a controller. Enable this to prevent flashes.
 #define ControllersSynchronizationWorkaround
 
+// define POST_RS5_SDK if using an SDK that is for a release
+// after RS5
+#if POST_RS5_SDK
+// For allowing of Windows.UI.Composition.VisualSurface and the
+// Lottie features that rely on it
```
simeoncran: Period on ends of sentences.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+            Layer.MatteType matteType = Layer.MatteType.None;
 
-                if (mask.Inverted)
+            // NOTE: we assume here that the items appear in reverse order from how they appear in the original Lottie file.
+            foreach (var item in items)
+            {
+                var layerIsMattedLayer = false;
+#if AllowVisualSurface
+                layerIsMattedLayer = item.layer.LayerMatteType == Layer.MatteType.Add ||
+                                    item.layer.LayerMatteType == Layer.MatteType.Invert;
```
simeoncran: The check should be for !Layer.MatterType.None

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -207,9 +217,14 @@ void Translate()
...
 
+            // Go through the layers and compose matte layer and layer to be matted into
+            // the resulting visuals. Any layer that is not a matte or matted layer is
+            // simply returned unmodified.
+            var shapeOrVisuals = ComposeMattedLayers(context, translatedLayers);
+
             // Layers are translated into either a Visual tree or a Shape tree. Convert the list of Visual and
             // Shape roots to a list of Visual roots by wrapping the Shape trees in ShapeVisuals.
-            var translatedAsVisuals = VisualsAndShapesToVisuals(context, translatedLayers.Select(a => a.translatedLayer));
+            var translatedAsVisuals = VisualsAndShapesToVisuals(context, shapeOrVisuals.ToArray());
```
simeoncran: Move the ".ToArray() up to the call to ComposeMattedLayers. That prevents someone from accidentally evaluating ComposeMattedLayers more than once.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
-            if (masks.Any())
-            {
-                var mask = masks[0];
+            // Save off the visual for the layer to be matted when we encounter it. The very next
+            // layer is the matte layer
+            Visual mattedVisual = null;
+            Layer.MatteType matteType = Layer.MatteType.None;
 
-                if (mask.Inverted)
+            // NOTE: we assume here that the items appear in reverse order from how they appear in the original Lottie file.
```
simeoncran: >we assume here that the items a [](start = 21, length = 31)

Seeing as the code in this file determines that, I'd make a stronger statement - they are in that order. And explain what that order means, e.g. the matte layer is always before the matted layer (or whatever it is).

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                    case CompositionObjectType.CompositionContainerShape:
+                    case CompositionObjectType.CompositionSpriteShape:
+                        if (layerIsMattedLayer ||
+                            mattedVisual != null)
+                        {
+                            var shapeVisual = _c.CreateShapeVisual();
+
+                            // ShapeVisual clips to its size
+#if NoClipping
+                            shapeVisual.Size = Vector2(float.MaxValue);
```
simeoncran: Note - we have a helper that does this now (create the shape visual, adds the shape, and set the size)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                        break;
+                    default:
+                        throw new InvalidOperationException();
                 }
 
-                if (mask.Mode != Mask.MaskMode.Additive)
+                if (visual != null)
                 {
-                    _issues.MaskWithUnsupportedMode(mask.Mode.ToString());
+                    // The layer to be matted is expected to come first. It is also expected
```
simeoncran: >expected [](start = 49, length = 8)

Be strong! It's not just expected, it DOES come first. We're not even checking for it, right?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                    }
                 }
-
-                // Translation currently does not support having multiple paths for masks.
-                // If possible users should combine masks when exporting to json.
-                if (masks.Length > 1)
+                else
                 {
-                    _issues.MultipleShapeMasksIsNotSupported();
+                    yield return item.translatedLayer;
```
simeoncran: Add a comment on each of the yield returns explaining what they're returning. That would help me understand the difference between returning "visual" and "item.translatedLayer". I think I understand what the compositedMatteVisual is.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                Visual visual = null;
+
+                switch (item.translatedLayer.Type)
                 {
-                    _issues.MaskWithAlphaIsNotSupported();
+                    case CompositionObjectType.CompositionContainerShape:
+                    case CompositionObjectType.CompositionSpriteShape:
+                        if (layerIsMattedLayer ||
+                            mattedVisual != null)
+                        {
```
simeoncran: Add a comment explaining (I think) that this layer is involved in matteing, but it is a Shape so it must be converted to a Visual.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+        // which is the final composed masked result.
+        Visual TranslateAndApplyMasksOnShapeTree(
+            TranslationContext context,
+            CompositionContainerShape containerShapeToMask)
+        {
+            var contextSize = context.Dimensions;
+
+            var contentShapeVisual = CreateShapeVisual();
+            contentShapeVisual.Size = contextSize;
+            contentShapeVisual.Shapes.Add(containerShapeToMask);
```
simeoncran: Use the new helper for this.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+            Mask.MaskMode maskMode = Mask.MaskMode.None;
+
+            if (layer.Masks.Any())
+            {
+                // Translate the mask paths
+                var compositionPathGeometryMasks = new List<(Mask mask, CompositionPathGeometry geometry)>();
+                foreach (var mask in layer.Masks)
+                {
+                    var currentMaskMode = mask.Mode;
+                    bool maskCanBeTranslated = true;
```
simeoncran: >bool [](start = 20, length = 4)

var

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,125 +281,335 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+                    if (mask.Inverted)
+                    {
+                        _issues.MaskWithInvertIsNotSupported();
+                        maskCanBeTranslated = false;
```
simeoncran: >skCanBeTranslated = false; [](start = 26, length = 26)

I'm not sure I follow the logic... is the idea that you will ignore this mask but translate all the others? Why not just replace the "maskCanBeTranslated = false" with a "continue" (and a comment).

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3602,6 +3816,8 @@ void Describe(IDescribable obj, string longDescription, string shortDescription
 
         static Sn.Vector3 Vector3(double x, double y, double z) => new Sn.Vector3((float)x, (float)y, (float)z);
 
+        static Sn.Vector3 Vector3(float x) => new Sn.Vector3((float)x, (float)x, (float)x);
```
simeoncran: >float [](start = 34, length = 5)

I couldn't see where this is called. Assuming it's useful, can the parameter be a double? If it needs to be a float then the body of the method doesn't need the casts.

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: Yep that is fine too.

---
In reply to: [287539729](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r287539729) [](ancestors = 287539729)

source/LottieToWinComp/TranslationContext.cs
eliezerpMS: As discussed offline, I will rename this to Size and get rid of Width and Height

---
In reply to: [287534668](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r287534668) [](ancestors = 287534668)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: As discussed offline it looks like the helper is not there. 

---
In reply to: [287539420](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r287539420) [](ancestors = 287539420)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: As discussed offline it looks like the helper is not there. 

---
In reply to: [287538866](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r287538866) [](ancestors = 287538866)

source/LottieToWinComp/TranslationContext.cs
eliezerpMS: I will use "using Sn = System.Numerics" I can't just use System.Numerics because then the Vector3 used in this file becomes ambiguous

---
In reply to: [287533201](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r287533201) [](ancestors = 287533201)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -311,79 +436,186 @@ void TranslateAndApplyMask(TranslationContext context, Visual visualForMask)
...
+            CanvasComposite compositeMode)
+        {
+            // There is a bug, introudced in the first Windows Release that supported
+            // CompositionVisualSurface where if the root visual is set as a source
+            // for a visual surface then its animations do not work. To work around
+            // this bug a top level parent is added for the source and destination.
+            var sourceIntermediateParent = CreateContainerVisual();
+            sourceIntermediateParent.Children.Add(source);
+
+            var destinationIntermediateParent = CreateContainerVisual();
```
simeoncran: Would it help future readers to have some ASCII art here that explains how the visuals and the effects are constructed?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -311,79 +436,186 @@ void TranslateAndApplyMask(TranslationContext context, Visual visualForMask)
...
+
+        // This function combines two visual trees using a Composite Effect. The way that the trees are combined
+        // is determined by the composite mode. The composition works as follows:
+        // 1) The visual trees are set as the source for CompositionVisualSurfaces. CompositionVisualSurface allows
+        // a user to use the contents of a visual tree as a surface.
+        // 2) Each CompositionVisualSurface is then hooked up to a CompositionVisualSurfaceBrush so that it can be rendered.
+        // 3) A CompositeEffect is created which will be used to do the combining of the contents of the CompositionVisualSurfaceBrushes.
+        // 4) A CompositionEffectBrush linked to the CompositeEffect is created so that the CompositeEffect can be rendered.
+        // 5) A SpriteVisual is created and the CompositionEffectBrush is assigned to it so that the CompositionEffectBrush contents are
+        // actually rendered as part of the visual tree.
```
simeoncran: If you do ASCII art I suspect these explanations aren't needed. I think ASCII art would be quicker to grok.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,44 +281,153 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+            TranslationContext context,
+            Layer layer,
+            CompositionContainerShape maskContainerShape)
+        {
+            Mask.MaskMode maskMode = Mask.MaskMode.None;
+
+            if (layer.Masks.Any())
+            {
+                // Translate the mask paths
+                var compositionPathGeometryMasks = new List<(Mask mask, CompositionPathGeometry geometry)>();
```
simeoncran: Is this used? I'm not seeing it.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -311,79 +436,186 @@ void TranslateAndApplyMask(TranslationContext context, Visual visualForMask)
...
+                    return null;
+                }
+
+                var contextSize = context.Size;
+
+                var maskShapeVisual = CreateShapeVisual();
+                maskShapeVisual.Size = contextSize;
+                maskShapeVisual.Shapes.Add(containerShapeMaskRootNode);
+
+                var maskMode = TranslateAndAddMaskPaths(context, layer, containerShapeMaskContentNode);
```
simeoncran: >maskMode [](start = 20, length = 8)

It feels awkward to be smuggling the success/failure flag into the maskMode. Why not make this bool TryTranslateAndAddMaskPaths(context, layer, containerShapeMaskContentMode, out maskMode) so that it's explicit.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -265,44 +281,153 @@ IEnumerable<Visual> VisualsAndShapesToVisuals(TranslationContext context, IEnume
...
+
+        // Helper for shape trees that need a mask applied.
+        // This function creates a ShapeVisual to house the
+        // shapes that we need to mask. It returns a visual
+        // which is the final composed masked result.
+        Visual TranslateAndApplyMasksOnShapeTree(
+            TranslationContext context,
+            CompositionContainerShape containerShapeToMask)
+        {
+            var contextSize = context.Size;
```
simeoncran: > var contextSize = context.Size; [](start = 11, length = 32)

Why not just set the .Size = context.Size; rather than storing it then setting it?

LottieGen/Program.cs
```diff
@@ -114,7 +114,7 @@ RunResult Run()
 #if DO_NOT_PROCESS_IN_PARALLEL
         foreach (var (file, relativePath) in matchingInputFiles)
         {
-            if (!LottieFileProcessor.ProcessFile(_options, _reporter, file, Path.Combine(outputFolder, relativePath)))
+            if (!LottieFileProcessor.ProcessFile(_options, _reporter, file, System.IO.Path.Combine(outputFolder, relativePath)))
```
simeoncran: >System.IO. [](start = 76, length = 10)

I don't see any other change that made this necessary? If anything it looks like we can remove the qualification from the other System.IO.Path usages.

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: I thought about it, but I was thinking that what it is doing does not lend itself super well to an actual tree structure.

---
In reply to: [288812803](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r288812803) [](ancestors = 288812803)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: hmm I guess I just forgot to delete it. I will delete this.

---
In reply to: [288813296](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r288813296) [](ancestors = 288813296)

LottieGen/Program.cs
eliezerpMS: gives me: error CS0104: 'Path' is an ambiguous reference between 'Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Path' and 'System.IO.Path'

---
In reply to: [288814420](https://github.com/windows-toolkit/Lottie-Windows/pull/106#discussion_r288814420) [](ancestors = 288814420)

PR #108, opened 2019-05-04
Move the YAML object model so UIData can use it.

simeoncran:
The YAML stuff was exclusive to LottieData before. But now I want to use Yaml for dumping WinCompData trees, thus this refactoring.


source/YamlData/YamlMap.cs
```diff
@@ -0,0 +1,53 @@
...
+#if PUBLIC_YamlData
+    public
+#endif
+    sealed class YamlMap : YamlObject, IEnumerable<(string key, YamlObject value)>
+    {
+        readonly List<(string key, YamlObject value)> _keysAndValues = new List<(string key, YamlObject value)>();
+        readonly HashSet<string> _keys = new HashSet<string>();
+
+        /// <summary>
+        /// Adds the give value to the map.
```
plaiMS: >give [](start = 21, length = 4)

given

source/YamlData/YamlMap.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (!_keys.Add(key))
+            {
+                throw new InvalidOperationException();
+            }
+
+            _keysAndValues.Add((key, value));
+        }
+
+        /// <summary>
+        /// Adds the give scalar to the map.
```
plaiMS: >give [](start = 21, length = 4)

given

PR #111, opened 2019-05-10
Enable image layer with external and embedded image assets.

plaiMS:


LottieGen/LottieFileProcessor.cs
```diff
@@ -339,25 +339,28 @@ bool TryGenerateCSharpCode(string outputFilePath)
...
         }
 
-        var code = CSharpInstantiatorGenerator.CreateFactoryCode(
-                    _className,
-                    _rootVisual,
-                    (float)_lottieComposition.Width,
-                    (float)_lottieComposition.Height,
-                    _lottieComposition.Duration,
-                    _options.DisableCodeGenOptimizer);
+        CSharpInstantiatorGenerator.CreateFactoryCode(
```
simeoncran: Now that we are not using the return value for the code, please have it return a bool and rename the method to TryCreateFactoryCode

LottieGen/LottieFileProcessor.cs
```diff
@@ -339,25 +339,28 @@ bool TryGenerateCSharpCode(string outputFilePath)
...
+                _className,
+                _rootVisual,
+                (float)_lottieComposition.Width,
+                (float)_lottieComposition.Height,
+                _lottieComposition.Duration,
+                out var csText,
+                out var infoText,
```
simeoncran: Put the outs last in the parameter list.

LottieGen/LottieFileProcessor.cs
```diff
@@ -339,25 +339,28 @@ bool TryGenerateCSharpCode(string outputFilePath)
...
             return false;
         }
 
-        var result = TryWriteTextFile(outputFilePath, code);
+        var result = TryWriteTextFile(outputFilePath, csText);
 
         if (result)
         {
             _reporter.WriteInfo($"C# code for class {_className} written to {outputFilePath}");
+            _reporter.WriteInfo(infoText);
```
simeoncran: >infoText [](start = 32, length = 8)

The info text should be created by LottieFileProcessor from information given to it by the CSharpInstantiatorGenerator. The CSharpInstantiatorGenerator should not be creating messages for the user.

source/Lottie/LottieVisualDiagnostics.cs
```diff
@@ -90,17 +90,22 @@ public string GenerateWinCompXml()
             return WinCompData.Tools.CompositionObjectXmlSerializer.ToXml(RootVisual).ToString();
         }
 
-        public string GenerateCSharpCode()
+        public void GenerateCSharpCode(out string csText)
```
simeoncran: > public void GenerateCSharpCode(out string csText) [](start = 7, length = 50)

Why is the signature being changed?

source/Lottie/Instantiator.cs
```diff
@@ -1228,8 +1228,17 @@ Wm.LoadedImageSurface GetLoadedImageSurface(Wmd.LoadedImageSurface obj)
...
+            switch (obj.LoadType)
+            {
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceLoadType.FromStream:
+                    var bytes = obj.Bytes;
+                    result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+                    break;
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceLoadType.FromUri:
+                    result = null;
+                    break;
+            }
```
simeoncran: > } [](start = 11, length = 2)

Add a default case the throws.

source/Lottie/Instantiator.cs
```diff
@@ -1228,8 +1228,17 @@ Wm.LoadedImageSurface GetLoadedImageSurface(Wmd.LoadedImageSurface obj)
...
-            var bytes = obj.Bytes;
-            result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+            switch (obj.LoadType)
+            {
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceLoadType.FromStream:
+                    var bytes = obj.Bytes;
+                    result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+                    break;
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceLoadType.FromUri:
+                    result = null;
```
simeoncran: >result = null; [](start = 20, length = 14)

Missing implementation?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 // Enable use of Image Layer
-//#define EnableImageLayer
```
simeoncran: I think this #define can die now, right?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -670,21 +670,40 @@ Visual TranslateImageLayer(TranslationContext context, ImageLayer layer)
...
+                    imageAssetHeight = embeddedImageAsset.Height;
+                    break;
+                case ImageAsset.ImageAssetType.External:
+                    var externalImageAsset = (ExternalImageAsset)imageAsset;
+                    surface = LoadedImageSurface.StartLoadFromUri($"{externalImageAsset.Path}{externalImageAsset.FileName}");
+                    imageAssetWidth = externalImageAsset.Width;
+                    imageAssetHeight = externalImageAsset.Height;
+                    _issues.ExternalImageFilesExpected($"{externalImageAsset.Path}{externalImageAsset.FileName}");
+                    break;
+            }
```
simeoncran: Add a default case that throws. Then get rid of the unnecessary initialization of variables that are initialized in the non-throwing cases.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -670,21 +670,40 @@ Visual TranslateImageLayer(TranslationContext context, ImageLayer layer)
...
-
             var content = CreateSpriteVisual();
             containerVisualContentNode.Children.Add(content);
             content.Size = new Sn.Vector2((float)imageAsset.Width, (float)imageAsset.Height);
 
-            var embeddedImageAsset = (EmbeddedImageAsset)imageAsset;
-            var surface = LoadedImageSurface.StartLoadFromStream(embeddedImageAsset.Bytes);
-            var imageBrush = CreateSurfaceBrush(surface);
+            LoadedImageSurface surface = null;
+            CompositionSurfaceBrush imageBrush = null;
```
simeoncran: >imageBrush [](start = 36, length = 10)

Declare at first use rather than up here.

source/UIData/CodeGen/Canonicalizer.cs
```diff
@@ -284,6 +321,29 @@ void CanonicalizeColorBrushes()
...
+                    from item in nodes
+                    let obj = item.Object
+                    where (_ignoreCommentProperties || obj.Comment == null)
+                       && obj.Properties.IsEmpty
+                    select (item.Node, obj);
+
+                var grouping =
+                    from item in items
+                    let obj = item.obj
+                    let animators = obj.Animators.ToArray()
```
simeoncran: >ToArray [](start = 50, length = 7)

Is there any case where we would could canonicalize with an animation? If we'll never want to canonicalize with an animation, don't covert to an array, just use "where !animators.Any()"

source/UIData/CodeGen/Canonicalizer.cs
```diff
@@ -468,6 +528,33 @@ into grouped
...
+                        break;
+                    case LoadedImageSurface.LoadedImageSurfaceLoadType.FromUri:
+                        var groupingExternal =
+                            from item in items
+                            let obj = item.Object
+                            group item.Node by
+                            new
+                            {
+                                obj.FilePath,
+                            }
```
simeoncran: Do you need to create this object, or can you just use obj.FilePath directly?

source/UIData/CodeGen/CodeBuilder.cs
```diff
@@ -234,6 +235,49 @@ static string GetLine(string text, int maxLineLength, out string remainder)
             return text.Substring(0, breakAt);
         }
 
+        internal void BytesToLiteral(byte[] bytes, int maximumColumns = 100)
```
simeoncran: > = 100 [](start = 69, length = 6)

Not clear to me why we should have a default value, or why it would be 100.

source/UIData/CodeGen/CodeBuilder.cs
```diff
@@ -234,6 +235,49 @@ static string GetLine(string text, int maxLineLength, out string remainder)
...
+        {
+            var bytesLines = BytesToBytesList(bytes, maximumColumns - 1 - IndentSize).ToArray();
+
+            int i;
```
simeoncran: > int i [](start = 11, length = 6)

Please add a comment explaining the intention of this code... the pattern is a bit unusual.

source/UIData/CodeGen/CodeBuilder.cs
```diff
@@ -234,6 +235,49 @@ static string GetLine(string text, int maxLineLength, out string remainder)
...
+            for (i = 0; i < bytesLines.Length - 1; i++)
+            {
+                WriteLine($"{bytesLines[i]},");
+            }
+
+            WriteLine(bytesLines[i]);
+        }
+
+        internal IEnumerable<string> BytesToBytesList(byte[] bytes, int maximumWidth)
```
simeoncran: >internal [](start = 8, length = 8)

Can this be private? Also can it be static?

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -239,6 +255,64 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
+                case LoadedImageSurface.LoadedImageSurfaceLoadType.FromStream:
+                    builder.WriteLine($"var result = Windows.UI.Xaml.Media.LoadedImageSurface.StartLoadFromStream({fieldName}.AsBuffer().AsStream().AsRandomAccessStream());");
+                    break;
+                case LoadedImageSurface.LoadedImageSurfaceLoadType.FromUri:
+                    builder.WriteLine($"var result = Windows.UI.Xaml.Media.LoadedImageSurface.StartLoadFromUri(new Uri(\"{_stringifier.ExternalImageFileUri(info.ClassName, obj.FilePath)}\"));");
+                    break;
+            }
+        }
+
+        static void BytesToCSharpLiteral(CodeBuilder builder, byte[] bytes, int maximumColumns = 100)
```
simeoncran: > s [](start = 7, length = 2)

This code is also in CodeBuilder. Can it be shared?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -130,17 +140,31 @@ namespace AnimatedVisuals
             builder.WriteLine("using namespace Windows::UI::Composition;");
             builder.WriteLine("using namespace Windows::Graphics;");
             builder.WriteLine("using namespace Microsoft::WRL;");
+            if (info.UsesLoadedImageSurfaceFromStream || info.UsesLoadedImageSurfaceFromUri)
```
simeoncran: Can/should we break these flags up so that they refer to the thing that the code needs rather than the thing that is in the code? E.g. UsesWindowsUiXamlMedia
Think about what happens when there's some other feature that we add then the decoding of the flags will get awkward.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -376,6 +412,31 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
             builder.WriteLine($"result = {FieldAssignment(fieldName)}new GeoSource(transformed);");
         }
 
+        /// <inheritdoc/>
+        protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, CodeGenInfo info, LoadedImageSurface obj, string fieldName)
+        {
+            switch (obj.LoadType)
+            {
+                case LoadedImageSurface.LoadedImageSurfaceLoadType.FromStream:
+                    builder.WriteLine("InMemoryRandomAccessStream ^ stream = ref new InMemoryRandomAccessStream();");
```
simeoncran: >  [](start = 65, length = 1)

No space before the hat.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1883,6 +1977,10 @@ protected internal abstract class StringifierBase : IStringifier
             public abstract string Vector3(Vector3 value);
 
             public string Hex(int value) => $"0x{value.ToString("X2")}";
+
+            public virtual string Static => "static";
+
+            public virtual string ExternalImageFileUri(string className, string filePath) => $"ms-appx:///Assets/{className}/{filePath}";
```
simeoncran: The name doesn't seem right - it's an asset, not necessarily an image file. And it's not external.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -376,6 +412,31 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
+                    builder.WriteLine("stream->Seek(0);");
+                    builder.WriteLine($"{_stringifier.Var} result = Windows::UI::Xaml::Media::LoadedImageSurface::StartLoadFromStream(stream);");
+                    break;
+                case LoadedImageSurface.LoadedImageSurfaceLoadType.FromUri:
+                    builder.WriteLine($"{_stringifier.Var} result = Windows::UI::Xaml::Media::LoadedImageSurface::StartLoadFromUri(ref new Uri(\"{_stringifier.ExternalImageFileUri(info.ClassName, obj.FilePath)}\"));");
+                    break;
+            }
+        }
+
+        protected override void WriteBytesField(CodeBuilder builder, string fieldName)
```
simeoncran: Is it necessary to create a virtual method for this? I'm hoping the base class can express this concept.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +283,48 @@ static string SanitizeTypeName(string name)
...
+                }
+            }
+
+            if (bytesWritten)
+            {
+                builder.WriteLine();
+            }
+        }
+
+        protected virtual void WriteBytesField(CodeBuilder builder, string fieldName)
```
simeoncran: Needs comments, especially so because it is now part of the protected interface for the class.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +283,48 @@ static string SanitizeTypeName(string name)
...
+            CodeBuilder builder,
+            CodeGenInfo info,
+            Wmd.LoadedImageSurface obj,
+            string fieldName);
+
+        void WriteBytesFields(CodeBuilder builder, ObjectData[] nodes)
+        {
+            bool bytesWritten = false;
+
+            foreach (var node in nodes)
```
simeoncran: >foreach [](start = 12, length = 7)

Rather than filtering inside a foreach, just do the foreach over a filtered list of nodes.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +283,48 @@ static string SanitizeTypeName(string name)
...
+        /// <param name="info">A <see cref="CodeGenInfo"/> used to get information about the code.</param>
+        /// <param name="obj">Describes the object that should be instantiated by the factory code.</param>
+        /// <param name="fieldName">The name of the Bytes field that should be used for StartLoadFromStream().</param>
+        protected abstract void WriteLoadedImageSurfaceFactory(
+            CodeBuilder builder,
+            CodeGenInfo info,
+            Wmd.LoadedImageSurface obj,
+            string fieldName);
+
+        void WriteBytesFields(CodeBuilder builder, ObjectData[] nodes)
```
simeoncran: This seems weird to pass in all of the nodes, or are they not all the nodes?

simeoncran: Please move this private method down below so it isn't interrupting the list of protected methods.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -239,6 +255,64 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
+        /// <inheritdoc/>
+        protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, CodeGenInfo info, LoadedImageSurface obj, string fieldName)
+        {
+            switch (obj.LoadType)
+            {
+                case LoadedImageSurface.LoadedImageSurfaceLoadType.FromStream:
+                    builder.WriteLine($"var result = Windows.UI.Xaml.Media.LoadedImageSurface.StartLoadFromStream({fieldName}.AsBuffer().AsStream().AsRandomAccessStream());");
+                    break;
+                case LoadedImageSurface.LoadedImageSurfaceLoadType.FromUri:
+                    builder.WriteLine($"var result = Windows.UI.Xaml.Media.LoadedImageSurface.StartLoadFromUri(new Uri(\"{_stringifier.ExternalImageFileUri(info.ClassName, obj.FilePath)}\"));");
```
simeoncran: >info.ClassName [](start = 156, length = 14)

This method shouldn't need to know the class name. Storing files in assets should be something that the base class should know about.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2015,5 +2124,25 @@ internal string TypeName
             // Sets the first character to lower case.
             static string CamelCase(string value) => $"_{char.ToLowerInvariant(value[0])}{value.Substring(1)}";
         }
+
+        /// <summary>
+        /// Call this to generate information text to output to the user.
+        /// </summary>
+        /// <param name="className">Class name of the code.</param>
+        /// <returns>Information about the code.</returns>
+        protected string GenerateInfoText(string className)
```
simeoncran: Put this up with the other protected methods for the class.
In general, the stuff at the bottom of the file should be nested classes, not methods.
But see other comments about how we should not be returning strings for this info.

source/UIData/Tools/ObjectGraph.cs
```diff
@@ -47,7 +47,13 @@ public static new ObjectGraph<T> FromCompositionObject(CompositionObject root, b
             return result;
         }
 
-        public IEnumerable<T> Nodes => _compositionObjectReferences.Values.Concat(_compositionPathReferences.Values).Concat(_canvasGeometryReferences.Values);
+        public IEnumerable<T> Nodes
+        {
+            get
```
simeoncran: > ge [](start = 11, length = 3)

Why switch to the get syntax? It's inconsistent with the other properties around here.

source/WinUIXamlMediaData/LoadedImageSurface.cs
```diff
@@ -9,20 +9,50 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinUIXamlMediaData
...
+
+        public LoadedImageSurfaceLoadType LoadType => _loadType;
+
+        /// <inheritdoc/>
+        public string LongDescription { get; set; }
+
+        /// <inheritdoc/>
+        public string ShortDescription { get; set; }
+
+        public enum LoadedImageSurfaceLoadType
```
simeoncran: >LoadedImageSurfaceLoadType [](start = 20, length = 26)

This is making me feel uncomfortable. Why not just have the caller check with Bytes or FilePath is null?
If you did have this, it would make sense to make LoadedImageSurface abstract and have a subclass for each load type.

source/WinUIXamlMediaData/LoadedImageSurface.cs
```diff
@@ -9,20 +9,50 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinUIXamlMediaData
...
+            _filePath = filePath;
+            _loadType = LoadedImageSurfaceLoadType.FromUri;
         }
 
         public static LoadedImageSurface StartLoadFromStream(byte[] bytes)
         {
             return new LoadedImageSurface(bytes);
         }
 
+        public static LoadedImageSurface StartLoadFromUri(string filePath)
```
simeoncran: >filePath [](start = 65, length = 8)

Why is it a file path if the method is called "FromUri"?

simeoncran: Should/can we support the version of StartLoadFromUri that takes a size?

simeoncran:
> [comment]: # (name:LayerEffectsIsNotSupported)

While you're editing the .md files, can you try saving them as UTF8 with no BOM so that they look nice in CodeFlow.

---
Refers to: source/Issues/LP0008.md:3 in 964dbbd. [](commit_id = 964dbbdee35b25297b12146993a6b22d64bba310, deletion_comment = False)

source/UIData/Tools/ObjectGraph.cs
plaiMS: VS suggested that change :P Switched back to original format.

---
In reply to: [283913002](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283913002) [](ancestors = 283913002)

LottieGen/LottieFileProcessor.cs
plaiMS: The disableFieldOptimization parameter is optional and must be the last parameter.

---
In reply to: [283587899](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283587899) [](ancestors = 283587899)

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
plaiMS: This code shouldn't be here. Removed.

---
In reply to: [283590296](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283590296) [](ancestors = 283590296)

LottieGen/LottieFileProcessor.cs
simeoncran: Should it be optional? 

---
In reply to: [283956130](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283956130) [](ancestors = 283956130,283587899)

LottieGen/LottieFileProcessor.cs
simeoncran: Actually... let's add a class for the result and shove everything into that class
* CSharpGeneratedCode
* CxGeneratedCode
Return null on failure.
In the CSharp one have properties:  CsText, ReferencedAssetPaths
In the Cx one have properties:  CxText, HeaderText, ReferencedAssetPaths

---
In reply to: [283587851](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283587851) [](ancestors = 283587851)

LottieGen/LottieFileProcessor.cs
simeoncran: And it should be in the result object from the Cs/Cx-InstantiatorGenerator. And inside LottieGen (i.e. here!) you should turn it into a nice message.

---
In reply to: [283588132](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283588132) [](ancestors = 283588132)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
simeoncran: UsesNamespaceWindowsUIXamlMedia
UsesNamespaceWindowsStorageStreams
UsesNamespacePlatform

---
In reply to: [283905663](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283905663) [](ancestors = 283905663)

LottieGen/LottieFileProcessor.cs
plaiMS: I realized that I can get the asset file list from _lottieComposition, which means I don't need to modify CreateFactoryCode at all. Should I revert the function back to its original form and not create the new class? 

---
In reply to: [283991046](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283991046) [](ancestors = 283991046,283587851)

plaiMS: I realized that I can get the asset file list from _lottieComposition, which means I don't need to modify CreateFactoryCode at all. Should I revert the function back to its original form and not create the new class? 

---
In reply to: [283991046](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283991046) [](ancestors = 283991046,283587851)

plaiMS: As discussed, changed to return Tuple instead of adding a new class.

---
In reply to: [284484652](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r284484652) [](ancestors = 284484652,283991046,283587851)

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -39,23 +44,28 @@ sealed class CSharpInstantiatorGenerator : InstantiatorGeneratorBase
         /// Returns the C# code for a factory that will instantiate the given <see cref="Visual"/> as a
         /// Windows.UI.Composition Visual.
         /// </summary>
-        /// <returns>The C# code.</returns>
-        public static string CreateFactoryCode(
+        /// <returns>A tuple containing the C# code and list of referenced asset files.</returns>
+        public static Tuple<string, List<string>> CreateFactoryCode(
```
simeoncran: I meant the new ValueTuple, not a Tuple. ValueTuple has a nicer syntax, and allows you to name the members.  e.g.    public static (string csText, List<string> infoText) CreateFactoryCode(....)

simeoncran: Don't return list. IReadOnlyList, or IEnumerable, or [] are ok.
List exists for creating lists, not really for passing their contents around.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +297,47 @@ static string SanitizeTypeName(string name)
...
+        /// Call this to get a list of the asset files referenced by the generated code.
+        /// </summary>
+        /// <param name="className">Class name of the code.</param>
+        /// <returns>List of asset files.</returns>
+        protected List<string> GetAssetFileList(string className)
+        {
+            var assetList = new List<string>();
+            foreach (var node in _nodes.Where(n => n.UsesAssetFile))
+            {
+                    assetList.Add($"{node.LoadedImageSurfaceImageUriString}");
```
simeoncran: >                     [](start = 0, length = 20)

Indenting.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +297,47 @@ static string SanitizeTypeName(string name)
...
+        {
+            builder.WriteLine($"{_stringifier.Static} {_stringifier.Readonly(_stringifier.ReferenceTypeName(_stringifier.ByteArray))} {fieldName} = {_stringifier.New} {_stringifier.ByteArray}");
+        }
+
+        /// <summary>
+        /// Call this to get a list of the asset files referenced by the generated code.
+        /// </summary>
+        /// <param name="className">Class name of the code.</param>
+        /// <returns>List of asset files.</returns>
+        protected List<string> GetAssetFileList(string className)
```
simeoncran: >className [](start = 55, length = 9)

className isn't used.

simeoncran: >List [](start = 18, length = 4)

Don't return List. An array would be fine here (assetList.ToArray()).

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: >$"{ [](start = 34, length = 3)

$" isn't doing anything useful here - it's already a string.

source/Lottie/Instantiator.cs
```diff
@@ -1228,8 +1228,20 @@ Wm.LoadedImageSurface GetLoadedImageSurface(Wmd.LoadedImageSurface obj)
...
-            var bytes = obj.Bytes;
-            result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+            switch (obj.Type)
+            {
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceType.FromStream:
+                    var bytes = obj.Bytes;
+                    result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+                    break;
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceType.FromUri:
+                    // Loading external image asset is not supported yet. LottieViewer will show this as an issue.
```
simeoncran: >LottieViewer [](start = 74, length = 12)

Don't mention LottieViewer in comments here - this code knows nothing about LottieViewer.

source/Lottie/LottieVisualDiagnostics.cs
```diff
@@ -94,13 +94,15 @@ public string GenerateCSharpCode()
...
-                    (float)LottieComposition.Width,
-                    (float)LottieComposition.Height,
-                    LottieComposition.Duration);
+            var generatedCode = CSharpInstantiatorGenerator.CreateFactoryCode(
+                SuggestedClassName,
+                RootVisual,
+                (float)LottieComposition.Width,
+                (float)LottieComposition.Height,
+                LottieComposition.Duration,
+                false);
```
simeoncran: >false [](start = 16, length = 5)

Add a name to this parameter so the reader knows what this refers to.

source/UIData/CodeGen/CodeBuilder.cs
simeoncran: Don't mark it with "private" - members are private by default and the rest of the code never includes the accessibility modifier unless it is different from the default.

---
In reply to: [283589874](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283589874) [](ancestors = 283589874)

source/UIData/CodeGen/CodeBuilder.cs
simeoncran: My point was really that the CodeBuilder should define the default policy. So it shouldn't have a default value.
Also, default values for parameters are usually a bit suspicious

---
In reply to: [283589741](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r283589741) [](ancestors = 283589741)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -670,28 +666,42 @@ Visual TranslateImageLayer(TranslationContext context, ImageLayer layer)
...
-            }
-
             var content = CreateSpriteVisual();
             containerVisualContentNode.Children.Add(content);
             content.Size = new Sn.Vector2((float)imageAsset.Width, (float)imageAsset.Height);
 
-            var embeddedImageAsset = (EmbeddedImageAsset)imageAsset;
-            var surface = LoadedImageSurface.StartLoadFromStream(embeddedImageAsset.Bytes);
+            LoadedImageSurface surface;
+            double imageAssetWidth;
```
simeoncran: Can you initialize these here rather than doing it twice in the switch?

source/WinUIXamlMediaData/LoadedImageSurfaceFromStream.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinUIXamlMediaData
+{
+#if PUBLIC_WinUIXamlMediaData
+    public
+#endif
+    sealed class LoadedImageSurfaceFromStream : LoadedImageSurface
+    {
+        private LoadedImageSurfaceFromStream(byte[] bytes)
```
simeoncran: >private [](start = 8, length = 7)

Don't use default accessibility modifiers.

source/WinUIXamlMediaData/LoadedImageSurface.cs
```diff
@@ -2,27 +2,36 @@
...
-        LoadedImageSurface(byte[] bytes)
+        private protected LoadedImageSurface()
         {
-            _bytes = bytes;
         }
 
-        public static LoadedImageSurface StartLoadFromStream(byte[] bytes)
+        public byte[] Bytes { get; set; }
+
+        public Uri ImageUri { get; set; }
```
simeoncran: Bytes and ImageUri should not be on the base class - they only make sense on the implementation classes.

source/WinUIXamlMediaData/LoadedImageSurfaceFromStream.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            Bytes = bytes;
+        }
+
+        public static LoadedImageSurfaceFromStream StartLoadFromStream(byte[] bytes)
```
simeoncran: >static [](start = 15, length = 6)

The static factory methods belong on the base class.

source/WinUIXamlMediaData/LoadedImageSurfaceFromUri.cs
```diff
@@ -0,0 +1,26 @@
...
+    public
+#endif
+    sealed class LoadedImageSurfaceFromUri : LoadedImageSurface
+    {
+        private LoadedImageSurfaceFromUri(Uri imageUri)
+        {
+            ImageUri = imageUri;
+        }
+
+        public static LoadedImageSurfaceFromUri StartLoadFromUri(Uri imageUri)
```
simeoncran: >imageUri [](start = 69, length = 8)

Use the same name as used in the real class - "uri"

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -239,6 +260,20 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
             builder.WriteLine($"var result = {FieldAssignment(fieldName)}{CallFactoryFor(obj.SourceGeometry)}.Transform({_stringifier.Matrix3x2(obj.TransformMatrix)});");
         }
 
+        /// <inheritdoc/>
+        protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, CodeGenInfo info, LoadedImageSurface obj, string fieldName, string imageUriString)
```
simeoncran: Why is this passed in? It isn't used in both cases, and AFAICT it came from the LoadedImageSurface that you're passing in.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +297,47 @@ static string SanitizeTypeName(string name)
...
+            Wmd.LoadedImageSurface obj,
+            string fieldName,
+            string imageUriString);
+
+        /// <summary>
+        /// Write a Bytes field.
+        /// </summary>
+        /// <param name="builder">A <see cref="CodeBuilder"/> used to create the code.</param>
+        /// <param name="fieldName">The name of the Bytes field to be written.</param>
+        protected virtual void WriteBytesField(CodeBuilder builder, string fieldName)
```
simeoncran: >virtual [](start = 18, length = 7)

Why is this virtual? Do you expect someone to override it?

source/UIData/CodeGen/Canonicalizer.cs
```diff
@@ -468,6 +527,28 @@ into grouped
...
+                        break;
+                    case LoadedImageSurface.LoadedImageSurfaceType.FromUri:
+                        var groupingExternal =
+                            from item in items
+                            let obj = item.Object
+                            group item.Node by obj.ImageUri into grouped
+                            select grouped;
+                        CanonicalizeGrouping(groupingExternal);
+                        break;
+                }
```
simeoncran: >  } [](start = 14, length = 3)

Needs a default case.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
plaiMS: the fieldName and imageUriString are extra information added to the Node, not in Object Graph.

---
In reply to: [284908454](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r284908454) [](ancestors = 284908454)

simeoncran: Isn't imageUriString derived from the LoadedImagSurface? Anyhow, let's discuss in person on the next iteration.

---
In reply to: [284914485](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r284914485) [](ancestors = 284914485,284908454)

source/UIData/CodeGen/CodeBuilder.cs
simeoncran: Comment is wrong now... talks about the default.

---
In reply to: [284905129](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r284905129) [](ancestors = 284905129,283589741)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -282,6 +296,40 @@ static string SanitizeTypeName(string name)
...
+        /// <param name="fieldName">The name of the Bytes field to be written.</param>
+        protected void WriteBytesField(CodeBuilder builder, string fieldName)
+        {
+            builder.WriteLine($"{_stringifier.Static} {_stringifier.Readonly(_stringifier.ReferenceTypeName(_stringifier.ByteArray))} {fieldName} = {_stringifier.New} {_stringifier.ByteArray}");
+        }
+
+        /// <summary>
+        /// Call this to get a list of the asset files referenced by the generated code.
+        /// </summary>
+        /// <returns>List of asset files.</returns>
```
simeoncran: Needs a little more info - are these paths, or just file names? Is the generated code going to load from that path, or is it relative to the assets directory in a UWP, etc.

source/Lottie/Instantiator.cs
```diff
@@ -1228,8 +1228,20 @@ Wm.LoadedImageSurface GetLoadedImageSurface(Wmd.LoadedImageSurface obj)
...
-            var bytes = obj.Bytes;
-            result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+            switch (obj.Type)
+            {
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceType.FromStream:
+                    var bytes = ((Wmd.LoadedImageSurfaceFromStream)obj).Bytes;
+                    result = Wm.LoadedImageSurface.StartLoadFromStream(bytes.AsBuffer().AsStream().AsRandomAccessStream());
+                    break;
+                case Wmd.LoadedImageSurface.LoadedImageSurfaceType.FromUri:
+                    // Loading external image asset is not supported yet and it is captured on the issue list.
```
simeoncran: >captured on the issue list [](start = 83, length = 26)

I don't understand this comment. Is it referring to the list of issues in GitHub? If it's that we output an issue, it might need a bit of explaining about where that is.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -670,28 +666,38 @@ Visual TranslateImageLayer(TranslationContext context, ImageLayer layer)
...
-            }
-
             var content = CreateSpriteVisual();
             containerVisualContentNode.Children.Add(content);
             content.Size = new Sn.Vector2((float)imageAsset.Width, (float)imageAsset.Height);
 
-            var embeddedImageAsset = (EmbeddedImageAsset)imageAsset;
-            var surface = LoadedImageSurface.StartLoadFromStream(embeddedImageAsset.Bytes);
+            LoadedImageSurface surface;
+            double imageAssetWidth = imageAsset.Width;
```
simeoncran: >double [](start = 12, length = 6)

Use var.

source/WinUIXamlMediaData/LoadedImageSurface.cs
```diff
@@ -2,27 +2,42 @@
...
 namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinUIXamlMediaData
 {
 #if PUBLIC_WinUIXamlMediaData
     public
 #endif
-    sealed class LoadedImageSurface : ICompositionSurface
+    abstract class LoadedImageSurface : ICompositionSurface, IDescribable
     {
-        readonly byte[] _bytes;
+        protected LoadedImageSurface()
```
simeoncran: Can this be private protected? Otherwise other assemblies can call it.

simeoncran:
> 

LT0004.md and LT0030.md - can you update them to UTF8 seeing as you're changing them?

---
Refers to: source/Issues/LT0004.md:12 in b205280. [](commit_id = b2052802ccf45b71cc1d4c717d075d33fcf038dc, deletion_comment = False)

simeoncran:
> [comment]: # (text:External image file expected at {filePath})

In what regard are these "External"? They're just files aren't they?

---
Refers to: source/Issues/LT0030.md:5 in b205280. [](commit_id = b2052802ccf45b71cc1d4c717d075d33fcf038dc, deletion_comment = False)

source/UIData/CodeGen/Canonicalizer.cs
```diff
@@ -468,6 +527,30 @@ into grouped
                 CanonicalizeGrouping(grouping);
             }
 
+            void CanonicalizeLoadedImageSurface(LoadedImageSurface.LoadedImageSurfaceType type)
+            {
+                // Canonicalize LoadeImageSurfaces.
```
simeoncran: Typo

source/UIData/CodeGen/CppStringifier.cs
```diff
@@ -102,8 +102,12 @@ public override string Matrix4x4(Matrix4x4 value)
 
         public override string FactoryCall(string value) => $"CanvasGeometryToIGeometrySource2D({value})";
 
+        public override string ByteArray => "Array<byte>";
+
         public string FailFastWrapper(string value) => $"FFHR({value})";
 
+        public string Const(string value) => $"const {value}";
```
simeoncran: There is a Const(...) method that is exactly the same in InstantiatorGeneratorBase. Seeing as const is the same in C# and CX, can this be on the StringifierBase as a non-virtual?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1883,6 +2004,10 @@ protected internal abstract class StringifierBase : IStringifier
             public abstract string Vector3(Vector3 value);
 
             public string Hex(int value) => $"0x{value.ToString("X2")}";
+
+            public virtual string Static => "static";
```
simeoncran: Does this need to be virtual? I would expect that it is the same in all languages.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: They're uris. Please return them as an IEnumerable<Uri> and call it GetAssetsList()

---
In reply to: [285242356](https://github.com/windows-toolkit/Lottie-Windows/pull/111#discussion_r285242356) [](ancestors = 285242356)

simeoncran:
> [comment]: # (text:Image file expected at {filePath})

I think this should be "required". ImageFileRequired. Saying "expected" implies that we're reporting the issue because the expectation was not met. It might be worth explaining what this means in the Remarks section because people are going to see this reported as an issue in LottieViewer, be confused, and go to this page to find out what's going on.

---
Refers to: source/Issues/LT0030.md:3 in c131919. [](commit_id = c131919e9422f13a480f78c79345b3a165146b49, deletion_comment = False)

PR #112, opened 2019-05-16
Add data representation of IRandomAccessStream for use by LoadedImage…

simeoncran:
…Surface.

LoadedImageSurface doesn't actually do anything with IRandomAccessStream yet.
Also tweaked the samples csproj to get rid of a warning.


source/WinUIXamlMediaData/LoadedImageSurface.cs
```diff
@@ -23,6 +24,12 @@ public static LoadedImageSurface StartLoadFromStream(byte[] bytes)
             return new LoadedImageSurface(bytes);
         }
 
+        public static LoadedImageSurface StartLoadFromStream(IRandomAccessStream bytes)
```
plaiMS: >bytes [](start = 81, length = 5)

name it "stream" to be same as real class?

PR #121, opened 2019-05-20
Setting Opacity transforms on layer.

plaiMS:


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -759,6 +789,8 @@ sealed class ShapeContentContext
         {
             readonly LottieToWinCompTranslator _owner;
 
+            static readonly Animatable<double> s_defaultOpacityPercent = new Animatable<double>(100, null);
```
simeoncran: Reorder - statics fields before non-static.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -880,12 +913,12 @@ internal void SetTransform(Transform transform)
 
             Animatable<double> ComposeOpacityPercents(Animatable<double> a, Animatable<double> b)
             {
-                if (a == null)
+                if (a == null || a == s_defaultOpacityPercent)
```
simeoncran: object.ReferenceEquals(a, s_defaultOpacityPercent). The goal is to short-circuit the check if someone passed in our special default value object, not to actually compare the value (which might be expensive).

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2335,7 +2375,12 @@ CompositionColorBrush TranslateShapeFill(TranslationContext context, SolidColorF
 
             containerShapeContentNode.Shapes.Add(rectangle);
 
-            rectangle.FillBrush = CreateAnimatedColorBrush(context, layer.Color, context.TrimAnimatable(layer.Transform.OpacityPercent));
+            // If the layer has masks then the opacity is set on a Visual in the chain returned
+            // for the layer from TryCreateContainerVisualTransformChain.
+            // If there is no masks then the opacity needs to be animated on the brush.
```
simeoncran: "If that layer has no masks, opacity is implemented via the alpha channel on the brush."

Issue #115, opened 2019-05-18
Linear and Radial Gradient Fills


FrayxRulez:
Do you have any plan to support gradients in the near future?

sohchatt:
@FrayxRulez yes -- it's been on our wish-list but had no concrete customer requests until yours. Thanks for the feedback; we'll be sure to add it to the feature backlog.

FrayxRulez:
I’ve actually worked on this over the weekend and I’ve fully implemented linear ones, would a PR be appreciated for them?

sohchatt:
Absolutely!

sohchatt:
Addressed by #123

PR #123, opened 2019-05-21
Implemented linear gradient fill

FrayxRulez:


source/LottieData/Color.cs
```diff
@@ -28,13 +28,13 @@ public static Color FromArgb(double a, double r, double g, double b)
 
         public static Color TransparentBlack { get; } = new Color(0, 0, 0, 0);
 
-        public double A { get; }
+        public double A { get; set; }
```
simeoncran: As much as possible the LottieData objects are meant to be immutable. Immutability helps reasoning about state, so we want to preserve it. Please remove the setters.

source/LottieData/GradientStop.cs
```diff
@@ -73,5 +66,10 @@ public override string ToString()
         }
 
         static string ToHex(double value) => ((byte)(value * 255)).ToString("X2");
+
+        public bool Equals(GradientStop other)
+        {
+            return other.Color == Color && other.Offset == Offset;
```
simeoncran: == Color is doing reference equality because we haven't overridden == on the Color class. You should add an == operator to Color, and add one for GradientStop too

source/LottieData/ShapeLayerFill.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
+{
+#if PUBLIC_LottieData
+    public
+#endif
+    abstract class ShapeLayerFill : ShapeLayerContent
```
simeoncran: To follow the existing pattern, this should get a type enum so users of a ShapeLayerFill can switch on its type.

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -1154,8 +1154,8 @@ LinearGradientFill ReadLinearGradientFill(JObject obj, in ShapeLayerContent.Shap
             IgnoreFieldThatIsNotYetSupported(obj, "hd");
 
             var opacityPercent = ReadOpacityPercent(obj);
-            var startPoint = ReadAnimatableVector2(obj.GetNamedObject("s"));
-            var endPoint = ReadAnimatableVector2(obj.GetNamedObject("e"));
+            var startPoint = ReadAnimatableVector3(obj.GetNamedObject("s"));
```
simeoncran: Why the change to vector3?

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -1825,7 +1825,8 @@ protected override Sequence<GradientStop> ReadValue(JToken obj)
                     throw new LottieCompositionReaderException("Fewer gradient stop values than expected");
                 }
 
-                var gradientStops = new List<GradientStop>(colorStopsDataLength / 4);
+                var colorStops = new List<GradientStop>(colorStopsDataLength / 4);
```
simeoncran: What is the purpose of changing to these 2 lists rather than using the existing implementation?

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -1881,25 +1882,158 @@ protected override Sequence<GradientStop> ReadValue(JToken obj)
...
 
-                // Merge the stops that have the same offset, and order by offset.
-                var merged =
-                    from stop in gradientStops
-                    group stop by stop.Offset into grouped
+                // Interpolate color and opacity stops into a single sequence
+                return new Sequence<GradientStop>(StopsBuilder.Build(colorStops, opacityStops));
+            }
+
+            class StopsBuilder
```
simeoncran: Can this be moved to another file? And please comment it, follow the order of members used elsewhere (e.g. constructor should be the first non-field member), make all state readonly if it can be (i.e. _result)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -819,7 +807,7 @@ internal void UpdateFromStack(Stack<ShapeLayerContent> stack)
                             break;
 
                         case ShapeContentType.SolidColorFill:
-                            Fill = ComposeSolidColorFill(Fill, (SolidColorFill)popped);
+                            Fill = ComposeSolidColorFill(Fill as SolidColorFill, (SolidColorFill)popped);
```
simeoncran: Don't use "as" when the type is known - just do a cast. Otherwise the reader has to assume that the result of the "as" could be null, which forces them to read more code than they need to.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3279,6 +3275,124 @@ CompositionColorBrush CreateNonAnimatedColorBrush(Color color)
...
+            {
+                result.EndPoint = Vector2(endPoint.InitialValue);
+            }
+
+            ApplyColorGradientStops(context, gradient.GradientStops, result, opacityPercent);
+
+            return result;
+        }
+
+        void ApplyColorGradientStops(TranslationContext context, in Animatable<Sequence<GradientStop>> stops, in CompositionGradientBrush result, in TrimmedAnimatable<double> opacityPercent)
```
simeoncran: Please move this method so that it's with the other "Apply*" methods and not in the middle of the "Create*" methods.

source/WinCompData/CompositionColorGradientStop.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Numerics;
```
simeoncran: Do we need System.Numerics? Remove all unused usings.

source/LottieReader/Serialization/LottieCompositionReader.cs
FrayxRulez: Looks like most `Vector2` animatables are actually `IAnimatableVector3` in the code, also `ApplyVector2KeyFrameAnimation` in the Translator asks for a `IAnimatableVector3` rather than `Animatable<Vector2>`

source/LottieReader/Serialization/LottieCompositionReader.cs
FrayxRulez: Any preference about class name & location?

source/LottieData/ShapeLayerFill.cs
FrayxRulez: In this case `SolidColorFill` would get a duplicate `FillType`.
It would make sense to move the already existing one to `ShapeLayerFill` and rename it to `PathFillType` or `FillMode` perhaps?

simeoncran: I'm not 100% sure that I'm understanding what you're thinking, but it sounds right. Let's see it in code.

source/LottieReader/Serialization/LottieCompositionReader.cs
simeoncran: Not yet... I'm actually still unclear about why it's needed. I thought we were already parsing everything needed.

FrayxRulez: Lottie uses a quite weird format for gradient stops:
There are two independant stop arrays within a single json array: one representing RGB values, the other representing opacity.
Stops positions don't match, obviously, so the idea is to merge those two values by interpolating the opacities with colors, to generate ARGB stops.

This should be done somewhere anyway: could be the Translator or the Parser, but this step is actually needed to render gradients properly.
I decided to have this in the parser as it is more convenient than having to process all the stops animatable again later in the Translator.

sohchatt: I believe @simeoncran  is referring to the AnimatableGradientStopsParser: https://github.com/windows-toolkit/Lottie-Windows/blob/f8b21d4ef5374a2fcaf3c0322425bc15ca37ce0c/source/LottieReader/Serialization/LottieCompositionReader.cs#L1802

sohchatt:
Closing because this was already addressed by #115

PR #133, opened 2019-06-07
Add support for Effects without Win2d for Cpp/Cx

eliezerpMS:
This change adds code to support allowing a developer to user Effects (which is necessary for Mattes and Masks) without the developer needing to take a dependency on Win2d.


source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -334,6 +334,24 @@ protected IEnumerable<Uri> GetAssetsList()
             return _nodes.Where(n => n.UsesAssetFile).Select(n => n.LoadedImageSurfaceImageUri).Distinct();
         }
 
+        /// <summary>
+        /// Write the CompositeEffect factory code.
+        /// </summary>
+        /// <param name="builder">A <see cref="CodeBuilder"/> used to create the code.</param>
+        /// <param name="compositeEffect">Composte effect object.</param>
```
plaiMS: >Composte [](start = 42, length = 8)

typo

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1849,6 +1874,11 @@ protected internal sealed class CodeGenInfo
             /// </summary>
             public bool UsesStreams { get; }
 
+            /// <summary>
+            /// Gets a value indicating whether the composition depends the composite effect used as a Windows.Graphics.Effects.IGraphicsEffect.
```
plaiMS: >whether the composition depends the composite effect used as a Windows.Graphics.Effects.IGraphicsEffect. [](start = 40, length = 104)

I don't understand this sentence. Please rephrase.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2123,6 +2153,9 @@ internal string LongComment
             // Set to indicate that the node uses asset file(s).
             internal bool UsesAssetFile => Object is Wmd.LoadedImageSurface lis && lis.Type == Wmd.LoadedImageSurface.LoadedImageSurfaceType.FromUri;
 
+            // Value indicating whether the composition depends the composite effect used as a Windows.Graphics.Effects.IGraphicsEffect.
```
plaiMS: >the composition depends the composite effect used as a Windows.Graphics.Effects.IGraphicsEffect [](start = 40, length = 95)

same sentence that needs rephrasing.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2123,6 +2153,9 @@ internal string LongComment
...
+            internal bool UsesCompositeEffect => Object is CompositionEffectBrush && ((CompositionEffectBrush)Object).GetEffect().Type == Mgce.GraphicsEffectType.CompositeEffect;
```
plaiMS: >Object is CompositionEffectBrush && ((CompositionEffectBrush)Object).GetEffect().Type == Mgce.GraphicsEffectType.CompositeEffect; [](start = 49, length = 129)

Can be written as :
Object is CompositionEffectBrush compositeEffectBrush && compositeEffectBrush.GetEffect().Type == Mgce.GraphicsEffectType.CompositeEffect;

which is more modern and consistent with the code above it.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -334,6 +334,24 @@ protected IEnumerable<Uri> GetAssetsList()
...
+        protected virtual string GenerateCompositeEffectFactory(CodeBuilder builder, Mgce.CompositeEffect compositeEffect)
+        {
+            builder.WriteLine($"{Var} compositeEffect = {New} CompositeEffect();");
+            builder.WriteLine($"compositeEffect{Deref}Mode = {CanvasCompositeMode(compositeEffect.Mode)};");
+            foreach (var source in compositeEffect.Sources)
+            {
+                builder.WriteLine($"compositeEffect{Deref}Sources{Deref}{IListAdd}({New} CompositionEffectSourceParameter({String(source.Name)}));");
+            }
+
+            return $"compositeEffect";
```
plaiMS: >$ [](start = 19, length = 1)

static string does not need the '$'.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -169,6 +170,13 @@ namespace AnimatedVisuals
                 // Typedef to simplify generation
                 builder.WriteLine("typedef ComPtr<GeoSource> CanvasGeometry;");
             }
+
+            if (info.UsesCanvasEffects)
+            {
+                // Write the composite effect class that will allow the use
+                // of this effect without win2d
```
plaiMS: >win2d [](start = 42, length = 5)

Missing a period at the end of sentence.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -169,6 +170,13 @@ namespace AnimatedVisuals
                 // Typedef to simplify generation
                 builder.WriteLine("typedef ComPtr<GeoSource> CanvasGeometry;");
             }
+
+            if (info.UsesCanvasEffects)
+            {
+                // Write the composite effect class that will allow the use
```
plaiMS: >//  [](start = 16, length = 3)

As we discussed offline last night, this PR will make the CX and C# code diverge. You should add a documentation on github about this.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -436,6 +444,22 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
             }
         }
 
+        /// <inheritdoc/>
+        protected override string GenerateCompositeEffectFactory(CodeBuilder builder, Mgce.CompositeEffect compositeEffect)
+        {
+            builder.WriteLine($"ComPtr<CompositeEffect> compositeEffect(new CompositeEffect());");
```
plaiMS: >$ [](start = 30, length = 1)

$ is not needed.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -436,6 +444,22 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
...
+            builder.WriteLine($"compositeEffect->SetMode({CanvasCompositeMode(compositeEffect.Mode)});");
+            foreach (var source in compositeEffect.Sources)
+            {
+                builder.OpenScope();
+                builder.WriteLine($"auto sourceParameter = ref new CompositionEffectSourceParameter({String(source.Name)});");
+                builder.WriteLine($"compositeEffect->AddSource(reinterpret_cast<ABI::Windows::Graphics::Effects::IGraphicsEffectSource*>(sourceParameter));");
```
plaiMS: >$ [](start = 34, length = 1)

$ is not needed.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -436,6 +444,22 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
...
+                builder.CloseScope();
+            }
+
+            return $"reinterpret_cast<Windows::Graphics::Effects::IGraphicsEffect^>(compositeEffect.Get())";
```
plaiMS: >$ [](start = 19, length = 1)

$ is not needed.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -446,6 +470,202 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
...
+    IFACEMETHODIMP GetProperty(
+        UINT index, 
+        ABI::Windows::Foundation::IPropertyValue ** value) override
+    {
+        return UsePropertyFactory([=](ABI::Windows::Foundation::IPropertyValueStatics* statics)
+        {
+            switch (index)
+            {
+            case D2D1_COMPOSITE_PROP_MODE: return statics->CreateUInt32(m_mode, (IInspectable**)value);
+            default: return E_INVALIDARG;
```
plaiMS: indent.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -446,6 +470,202 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
...
+
+    IFACEMETHODIMP GetTrustLevel(
+        TrustLevel* trustLvl) override
+    {
+        *trustLvl = BaseTrust;
+        return S_OK;
+    }
+
+private:
+    // Invokes a functor with the pointer to the property factory
```
plaiMS: >factory [](start = 58, length = 7)

ends with '.'

source/WinCompData/Compositor.cs
```diff
@@ -66,5 +66,7 @@ sealed class Compositor
         public CompositionSurfaceBrush CreateSurfaceBrush(ICompositionSurface surface) => new CompositionSurfaceBrush(surface);
 
         public CompositionEffectFactory CreateEffectFactory(Mgce.GraphicsEffectBase effect) => new CompositionEffectFactory(effect);
+
+        public CompositionEffectBrush CreateEffectBrush(Mgce.GraphicsEffectBase effect) => new CompositionEffectBrush(effect);
```
simeoncran: This isn't a method on Compositor AFAICT

eliezerpMS: Hmm I think I must not have done a merge yet. I branched this off of my previous change.

---
In reply to: [293542003](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r293542003) [](ancestors = 293542003)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
eliezerpMS: Ok. I will sync up with you offline about where to put documentation about this.

---
In reply to: [293504363](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r293504363) [](ancestors = 293504363)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
plaiMS: I don't see any changes for this sentence. 

---
In reply to: [293498016](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r293498016) [](ancestors = 293498016)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
plaiMS: Did you rephrase?

---
In reply to: [293497631](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r293497631) [](ancestors = 293497631)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
plaiMS: That'll be a question for Sohum :)

---
In reply to: [297456313](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r297456313) [](ancestors = 297456313,293504363)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -344,6 +344,24 @@ where n.IsLoadedImageSurface
                                 ((Wmd.LoadedImageSurface)n.Object).Type);
         }
 
+        /// <summary>
+        /// Write the CompositeEffect factory code.
+        /// </summary>
+        /// <param name="builder">A <see cref="CodeBuilder"/> used to create the code.</param>
+        /// <param name="compositeEffect">Composite effect object.</param>
+        /// <returns>string representation of the composite effect.</returns>
+        protected virtual string GenerateCompositeEffectFactory(CodeBuilder builder, Mgce.CompositeEffect compositeEffect)
```
simeoncran: >GenerateCompositeEffectFactory [](start = 33, length = 30)

Because the code is very different between C# and CX then this method should be abstract and you should add an implementation into both. There should not be default implementations of things in this class that only apply to C#.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1427,23 +1446,27 @@ bool GenerateCompositionEffectBrushFactory(CodeBuilder builder, CompositionEffec
...
             switch (effect.Type)
             {
                 case Mgce.GraphicsEffectType.CompositeEffect:
-                    var compositeEffect = (Mgce.CompositeEffect)effect;
-
-                    builder.WriteLine($"{Var} compositeEffect = {New} CompositeEffect();");
-                    builder.WriteLine($"compositeEffect{Deref}Mode = {CanvasCompositeMode(compositeEffect.Mode)};");
+                    effectCreationString = GenerateCompositeEffectFactory(builder, (Mgce.CompositeEffect)effect);
+                    break;
+                default:
```
simeoncran: >default [](start = 16, length = 7)

Do we ever generate code that produces other GraphicsEffectType? Looks like you're asserting that we don't... maybe needs a comment explaining that.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1737,7 +1760,7 @@ string StrokeLineJoin(CompositionStrokeLineJoin value)
             }
         }
 
-        string CanvasCompositeMode(CanvasComposite value)
+        protected string CanvasCompositeMode(CanvasComposite value)
```
simeoncran: >protected [](start = 8, length = 9)

Call me crazy, but... I'd prefer that we don't expose these helpers - they are only meant to be available inside this class. Clever conversion from a type to syntax should either be done in this class, or should be on the stringifier, or should be in the subclass. I want to keep the interface into this class as small as possible. For CX you should hold your nose, copy-paste, and replace the {ScopeResolve} with ::.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1803,7 +1826,8 @@ protected internal sealed class CodeGenInfo
                 bool usesNamespaceWindowsUIXamlMedia,
                 bool usesStreams,
                 bool hasLoadedImageSurface,
-                IEnumerable<LoadedImageSurfaceNode> loadedImageSurfaceNodes)
+                IEnumerable<LoadedImageSurfaceNode> loadedImageSurfaceNodes,
+                bool usesCompositeEffect)
```
simeoncran: For reasons of obsessive consistency, move this up with the other bools.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+    SourceAtop = 6,
+    DestinationAtop = 7,
+    Xor = 8,
+    Add = 9,
+    Copy = 10,
+    BoundedCopy = 11,
+    MaskInvert = 12,
+} CanvasComposite;
+
+class CompositeEffect WrlFinal :
```
simeoncran: >WrlFinal [](start = 22, length = 8)

WrlFinal is cruft from a past era. Replace with "final"

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
 }};
 }}";
         }
+
+        string CompositionEffectClass =>
+@"
+#include <vector>
+
+typedef enum CanvasComposite : int
```
simeoncran: >typedef [](start = 0, length = 7)

Shouldn't need to typedef.  I realize this is all copy-paste, but if it's not too onerous we should avoid inheriting ugly.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+        return S_OK;
+    }
+
+    IFACEMETHODIMP GetPropertyCount(UINT * count) override { *count = 1; return S_OK; }
+
+    IFACEMETHODIMP GetProperty(
+        UINT index, 
+        ABI::Windows::Foundation::IPropertyValue ** value) override
+    {
+        return UsePropertyFactory([=](ABI::Windows::Foundation::IPropertyValueStatics* statics)
```
simeoncran: >UsePropertyFactory [](start = 15, length = 18)

This seems way to clever for our use. Any reason to keep the template and lambda rather than just writing the code out directly?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+    MaskInvert = 12,
+} CanvasComposite;
+
+class CompositeEffect WrlFinal :
+    public ABI::Windows::Graphics::Effects::IGraphicsEffect,
+    public ABI::Windows::Graphics::Effects::IGraphicsEffectSource,
+    public ABI::Windows::Graphics::Effects::IGraphicsEffectD2D1Interop
+{
+public:
+    CompositeEffect() = default;
```
simeoncran: >   CompositeEffect() = default; [](start = 1, length = 31)

Is this necessary?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
simeoncran: >CompositeEffect [](start = 6, length = 15)

It would be good to have a small comment explaining the purpose of this code (which I think is that it's a partial implementation of various IGraphicsEffect* interfaces, sufficient for use by Composition.... right?)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+        return S_OK; 
+    }
+
+    IFACEMETHODIMP GetSourceCount(UINT* count) override
+    {
+        if (count == nullptr)
+        {
+            return E_INVALIDARG;
+        }
+
```
simeoncran: Consider tightening up some of this code - we really don't care if someone calls this with no UINT*, just don't deref the nullptr and return S_OK.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+    {
+        Microsoft::WRL::ComPtr<ABI::Windows::Foundation::IPropertyValueStatics> propertyValueFactory;
+        Microsoft::WRL::Wrappers::HStringReference activatableClassId{ RuntimeClass_Windows_Foundation_PropertyValue };
+        HRESULT hr = ABI::Windows::Foundation::GetActivationFactory(activatableClassId.Get(), &propertyValueFactory);
+        return FAILED(hr) ? hr : func(propertyValueFactory.Get());
+    }
+
+    ULONG m_cRef = 0;
+
+    CanvasComposite m_mode = CanvasComposite::SourceOver;
```
simeoncran: >SourceOver [](start = 46, length = 10)

Why do we initialize to this explicit value rather than just the default value (which happens to be the same thing). I.e. why not just {} initialize all of the fields?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+    // IGraphicsEffect
+    IFACEMETHODIMP get_Name(HSTRING* name) override { return m_name.CopyTo(name); }
+
+    IFACEMETHODIMP put_Name(HSTRING name) override { return m_name.Set(name); }
+
+    // IGraphicsEffectD2D1Interop
+    IFACEMETHODIMP GetEffectId(GUID* id) override 
+    { 
+        // set CLSID_D2D1Composite value
+        *id = { 0x48fc9f51, 0xf6ac, 0x48f1, { 0x8b, 0x58,  0x3b,  0x28,  0xac,  0x46,  0xf7,  0x6d } };
```
simeoncran: Not checking for id == nullptr here.  You should do what I suggest for GetSourceCount (i.e. just don't crash but return S_OK)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +735,212 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+
+        return cRef;
+    }
+
+    // IInspectable
+    IFACEMETHODIMP GetIids(
+        ULONG * iidCount,
+        IID ** iids) override
+    {
+        *iidCount = 0;
```
simeoncran: null checks?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
eliezerpMS: I will put in in stringifierBase and on the Istringifier.

---
In reply to: [297912117](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r297912117) [](ancestors = 297912117)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
eliezerpMS: We could support more in the future.

---
In reply to: [297911226](https://github.com/windows-toolkit/Lottie-Windows/pull/133#discussion_r297911226) [](ancestors = 297911226)

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -501,12 +502,27 @@ void WriteHandleLoadCompleted(CodeBuilder builder)
...
+        protected override string WriteCompositeEffectFactory(CodeBuilder builder, Mgce.CompositeEffect compositeEffect)
+        {
+            builder.WriteLine("var compositeEffect = new CompositeEffect();");
+            builder.WriteLine($"compositeEffect.Mode = {_stringifier.CanvasCompositeMode(compositeEffect.Mode)};");
+            foreach (var source in compositeEffect.Sources)
+            {
+                builder.WriteLine($"compositeEffect.Sources.Add(new CompositionEffectSourceParameter({String(source.Name)}));");
+            }
+
+            return "compositeEffect";
```
simeoncran: >compositeEffect [](start = 20, length = 15)

This string exists in 4 separate places in this method. It should be defined once and reused rather than copying the string around.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +741,223 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+    SourceAtop = 6,
+    DestinationAtop = 7,
+    Xor = 8,
+    Add = 9,
+    Copy = 10,
+    BoundedCopy = 11,
+    MaskInvert = 12,
+};
+
+// This class is a substitue for the Microsoft::Graphics::Canvas::Effects::CompositeEffect
```
simeoncran: >substitue [](start = 19, length = 9)

typo

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -709,5 +741,223 @@ void RaiseAnimatedVisualInvalidatedEvent(IDynamicAnimatedVisualSource^ sender, P
...
+        }
+
+        if (iid == __uuidof(IGraphicsEffectD2D1Interop))
+        {
+            AddRef();
+            *ppvObject = (IGraphicsEffectD2D1Interop*)this;
+            return S_OK;
+        }
+
+        return E_NOINTERFACE;
```
simeoncran: >return [](start = 8, length = 6)

*ppvObject is not assigned in this case. I think the contract is specified in various places as _out_... which means it should get nullptr.
Anyhow, consider refactoring this to make it shorter and more efficient:
a. check whether ppvObject is nullptr, and if it isn't...
b. assign *ppvObject with nullptr
c. do the __uuidof checks, and when one succeeds assign the result of the cast this pointer. Use static_cast<...> syntax to be modern. Use "else if" to avoid the multiple GUID comparisons.
d. After all the if/else-ifs, if *ppvObject != null, AddRef() and return S_OK. Otherwise return E_NONTERFACE.

PR #134, opened 2019-06-12
CodeGen IDynamicAnimatedVisualSource for composition with LoadedImageSurface.

plaiMS:


source/UIData/CodeGen/CodeBuilder.cs
```diff
@@ -41,6 +41,20 @@ internal void WriteComment(string comment)
...
         }
 
+        internal void WriteSummaryComment(string comment)
+        {
+            if (!string.IsNullOrWhiteSpace(comment))
+            {
+                WriteLine($"/// <summary>");
+                foreach (var line in BreakUpLine(comment))
+                {
+                    WriteLine($"/// {line}");
```
simeoncran: I noticed before that these lines seem to have an extra space at the start. An easy fix is {line.Trim()} but maybe the BreakUpLine code needs a tweak so the trim isn't necessary here.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -90,12 +90,14 @@ protected override void WriteFileStart(CodeBuilder builder, CodeGenInfo info)
             if (info.UsesNamespaceWindowsUIXamlMedia)
             {
                 builder.WriteLine("using Windows.UI.Xaml.Media;");
+                builder.WriteLine("using System.Runtime.InteropServices.WindowsRuntime;");
```
simeoncran: These should get ordered. The cod here needs to get a bit more clever... rather than just adding the usings to the builder, add them to a list, then sort them, then add them all.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -105,6 +107,24 @@ protected override void WriteFileStart(CodeBuilder builder, CodeGenInfo info)
             builder.WriteLine("using Windows.UI.Composition;");
 
             builder.WriteLine();
+
+            // If the composition has LoadedImageSurface, write a class that implements the IDynamicAnimatedVisualSource interface.
+            // Otherwise, implement the IAnimatedVisualSource interface.
+            if (info.UsesNamespaceWindowsUIXamlMedia)
```
simeoncran: The comment says this depends on the use of LoadedImageSurface, but the if says something different.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -260,18 +421,91 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
+            builder.CloseScope();
+            builder.CloseScope();
+            builder.WriteLine();
+        }
+
+        void WriteAnimatedVisualCall(CodeBuilder builder, IEnumerable<ObjectData> nodes)
+        {
+            builder.WriteLine("new AnimatedVisual(compositor,");
+            builder.Indent();
+            for (var i = 0; i < nodes.Count(); i++)
```
simeoncran: Never use Count() in this way... it will enumerate all the items. If you need the size from an IEnumerable then call ToArray() on it and access the array instead.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -260,18 +421,91 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
-                case LoadedImageSurface.LoadedImageSurfaceType.FromStream:
-                    builder.WriteLine($"var result = Windows.UI.Xaml.Media.LoadedImageSurface.StartLoadFromStream({fieldName}.AsBuffer().AsStream().AsRandomAccessStream());");
-                    break;
-                case LoadedImageSurface.LoadedImageSurfaceType.FromUri:
-                    builder.WriteLine($"var result = Windows.UI.Xaml.Media.LoadedImageSurface.StartLoadFromUri(new Uri(\"{imageUri}\"));");
-                    break;
+                if (i < nodes.Count() - 1)
+                {
+                    // Append "," to each parameter except the last one.
+                    builder.WriteLine($"{_stringifier.MakeVariableName(nodes.ElementAt(i).Name)},");
```
simeoncran: Most of these 2 cases are identical. You should be able to do a build.Write(...) for the common bit.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -63,16 +65,21 @@ sealed class CxInstantiatorGenerator : InstantiatorGeneratorBase
 
             var cppText = generator.GenerateCode(className, width, height);
 
-            var hText = GenerateHeaderText(className);
+            var hText = GenerateHeaderText(className, generator.GetLoadedImageSurfacesNodes());
```
simeoncran: >GetLoadedImageSurfacesNodes [](start = 64, length = 27)

Rather than calling the instance to get the nodes then pass it to a static method, wouldn't it make more sense to make the GenerateHeaderText method an instance method?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -634,7 +634,8 @@ void WriteCodeForNode(CodeBuilder builder, ObjectData node, CodeGenInfo info)
                         GenerateCanvasGeometryFactory(builder, (CanvasGeometry)node.Object, node);
                         return;
                     case Graph.NodeType.LoadedImageSurface:
-                        GenerateLoadedImageSurfaceFactory(builder, info, (Wmd.LoadedImageSurface)node.Object, node);
+                        //GenerateLoadedImageSurfaceFactory(builder, info, (Wmd.LoadedImageSurface)node.Object, node);
+                        // LoadedImageSurface is written out in the IDynamicAnimatedVisualSource class, so does not need to do anything here.
```
simeoncran: Is this code meant to be commented out?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -676,7 +677,7 @@ bool GenerateCanvasGeometryFactory(CodeBuilder builder, CanvasGeometry obj, Obje
             return true;
         }
 
-        bool GenerateObjectFactory(CodeBuilder builder, CompositionObject obj, ObjectData node)
+        bool GenerateObjectFactory(CodeBuilder builder, CompositionObject obj, ObjectData node, CodeGenInfo info)
```
simeoncran: >CodeGenInfo [](start = 96, length = 11)

Please make the CodeGenInfo the second parameter? The goal is to order the parameters by most-stable to least-stable.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2016,10 +2010,18 @@ protected internal abstract class StringifierBase : IStringifier
...
+
+            // Prepends "_" to camel case of the passed in value.
+            public string MakeVariableName(string value) => $"_{CamelCase(value)}";
+
+            public string Int32TypeName => "int";
         }
 
         // A node in the object graph, annotated with extra stuff to assist in code generation.
-        sealed class ObjectData : Graph.Node<ObjectData>
+        protected sealed class ObjectData : Graph.Node<ObjectData>
```
simeoncran: >otecte [](start = 10, length = 6)

Is there a way to avoid this? The ObjectData is supposed to be a secret.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2016,10 +2010,18 @@ protected internal abstract class StringifierBase : IStringifier
...
 
             public string Const(string value) => $"const {value}";
+
+            // Sets the first character to lower case.
+            public string CamelCase(string value) => $"{char.ToLowerInvariant(value[0])}{value.Substring(1)}";
+
+            // Prepends "_" to camel case of the passed in value.
+            public string MakeVariableName(string value) => $"_{CamelCase(value)}";
+
+            public string Int32TypeName => "int";
```
simeoncran: Move this up next to the int64 one.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2016,10 +2010,18 @@ protected internal abstract class StringifierBase : IStringifier
             public abstract string ByteArray { get; }
 
             public string Const(string value) => $"const {value}";
+
+            // Sets the first character to lower case.
+            public string CamelCase(string value) => $"{char.ToLowerInvariant(value[0])}{value.Substring(1)}";
+
+            // Prepends "_" to camel case of the passed in value.
+            public string MakeVariableName(string value) => $"_{CamelCase(value)}";
```
simeoncran: These shouldn't be part of the stringifier because they are not being asked to do language-specific things. CamelCasing and prepending an underscore are always going to work the same.

source/UIData/CodeGen/NodeNamer.cs
```diff
@@ -47,7 +48,19 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
                         baseName = "Geometry";
                         break;
                     case Graph.NodeType.LoadedImageSurface:
```
simeoncran: >oadedImageSurface [](start = 41, length = 17)

Factor this the same way CompositionObject is done... call out to a DescribeLoadedImageSurface(...) method. that creates a descriptive base name.

source/UIData/CodeGen/NodeNamer.cs
```diff
@@ -66,23 +79,47 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
             foreach (var entry in nodesByTypeName)
             {
                 var baseName = entry.Key;
-                var n = entry.Value;
-                if (n.Count == 1)
+                var nodeList = entry.Value;
+
+                if (baseName == "Image")
```
simeoncran: >if (baseName == "Image") [](start = 16, length = 24)

Don't special case Image here... do all the image-specific stuff when you generate the base name.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
plaiMS: Put $"{_stringifier.MakeVariableName(nodes.ElementAt(i).Name)}" in a var

---
In reply to: [293532232](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r293532232) [](ancestors = 293532232)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
plaiMS: As discussed, not changing this.

---
In reply to: [293536665](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r293536665) [](ancestors = 293536665)

source/UIData/CodeGen/CodeBuilder.cs
```diff
@@ -232,7 +246,7 @@ static string GetLine(string text, int maxLineLength, out string remainder)
             }
 
             remainder = text.Substring(breakAt);
-            return text.Substring(0, breakAt);
+            return text.Substring(0, breakAt).TrimStart();
```
simeoncran: What would be the difference if you just converted the TrimEnd() at the start of the method to a Trim()? It is triggering me a bit that we start by trimming one end of the string, then at the end we trim the other end of the string.
Either way, the comment on the method should probably indicate how the lines are trimmed so we'll know in future what the contract is.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -72,39 +72,69 @@ sealed class CSharpInstantiatorGenerator : InstantiatorGeneratorBase
         // Called by the base class to write the start of the file (i.e. everything up to the body of the Instantiator class).
         protected override void WriteFileStart(CodeBuilder builder, CodeGenInfo info)
         {
+            // A list to hold the namespaces that the generated code will use. The list will be ordered and written out after all namespaces are added.
+            List<string> namepaceList = new List<string>();
```
simeoncran: >List<string [](start = 12, length = 11)

var. And consider:
a) Call it "namespaces". I don't think the List-ness of it is interesting.
b) Make it a set. That way we're making it clear that you can add things as often as you like and we won't do the wrong thing.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -72,39 +72,69 @@ sealed class CSharpInstantiatorGenerator : InstantiatorGeneratorBase
...
+            namepaceList.Add("System");
+            namepaceList.Add("System.Numerics");
+            namepaceList.Add("Windows.UI");
+            namepaceList.Add("Windows.UI.Composition");
+
+            // Order the list alphabetically and write each namespace using out.
+            var orderList = namepaceList.OrderBy(n => n).ToArray();
+            foreach (var n in orderList)
+            {
+                builder.WriteLine($"using {n.ToString()};");
```
simeoncran: ToString is redundant. The { } will call ToString for you, and n is a string anyway.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -72,39 +72,69 @@ sealed class CSharpInstantiatorGenerator : InstantiatorGeneratorBase
...
-            builder.WriteLine("using Windows.UI.Composition;");
+            namepaceList.Add("Microsoft.UI.Xaml.Controls");
+            namepaceList.Add("System");
+            namepaceList.Add("System.Numerics");
+            namepaceList.Add("Windows.UI");
+            namepaceList.Add("Windows.UI.Composition");
+
+            // Order the list alphabetically and write each namespace using out.
+            var orderList = namepaceList.OrderBy(n => n).ToArray();
+            foreach (var n in orderList)
```
simeoncran: The ToArray() isn't necessary because you're only using it once. If you put this inside the foreach initializer then that will be even clearer because then the ordered list won't have a name that anyone can refer to it by.

source/UIData/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -260,18 +433,102 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
+            builder.WriteLine();
+        }
+
+        void WriteAnimatedVisualCall(CodeBuilder builder, CodeGenInfo info)
+        {
+            builder.WriteLine("new AnimatedVisual(compositor,");
+            builder.Indent();
+
+            var n = info.LoadedImageSurfaceNodes.ToArray();
+            for (var i = 0; i < n.Length; i++)
```
simeoncran: "nodes" instead of "n" would help my reading.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -63,16 +65,22 @@ sealed class CxInstantiatorGenerator : InstantiatorGeneratorBase
...
+            var hText = generator.GenerateHeaderText(className);
 
             var assetList = generator.GetAssetsList();
 
             return (cppText, hText, assetList);
         }
 
         // Generates the .h file contents.
-        static string GenerateHeaderText(string className)
+        string GenerateHeaderText(string className)
```
simeoncran: >GenerateHeaderText [](start = 15, length = 18)

Refactor this method so it doesn't contain the non-dynamic version. Instead have a IAnimatedVisualSourceHeaderText(...) and a IDynamicAnimatedVisualSourceHeaderText(...) and this method just returns one or the other. 
And put both of the header text methods next to each other. That will make it easier for readers to see the difference.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -63,16 +65,22 @@ sealed class CxInstantiatorGenerator : InstantiatorGeneratorBase
...
         {
+            var loadedImageSurfacesNodes = GetLoadedImageSurfacesNodes();
+            if (loadedImageSurfacesNodes?.Any() == true)
```
simeoncran: >== true [](start = 48, length = 7)

Is the comparison with true necessary?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -128,6 +136,11 @@ namespace AnimatedVisuals
                 builder.WriteLine("#include <iostream>");
             }
 
+            builder.WriteLine();
+
+            // A list to hold the namespaces that the generated code will use. The list will be ordered and written out after all namespaces are added.
+            List<string> namespaceList = new List<string>();
```
simeoncran: >namespaceList [](start = 25, length = 13)

See comments in the C# version about this pattern.

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -218,15 +237,56 @@ protected override void WriteInstantiatorStart(CodeBuilder builder, CodeGenInfo
...
-            builder.Indent();
-            builder.WriteLine(": _c(compositor)");
+            if (info.HasLoadedImageSurface)
+            {
+                builder.WriteLine("AnimatedVisual(Compositor^ compositor,");
+                builder.Indent();
+
+                // Pass in the image surfaces to AnimatedVisual() constructor.
+                var nodes = GetLoadedImageSurfacesNodes();
+                for (var i = 0; i < nodes.Count(); i++)
```
simeoncran: >Count [](start = 42, length = 5)

BAD!

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -416,24 +497,128 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
-                    builder.WriteLine($"dataWriter->WriteBytes({fieldName});");
-                    builder.WriteLine("dataWriter->StoreAsync();");
-                    builder.WriteLine("dataWriter->FlushAsync();");
-                    builder.WriteLine("stream->Seek(0);");
-                    builder.WriteLine($"{_stringifier.Var} result = Windows::UI::Xaml::Media::LoadedImageSurface::StartLoadFromStream(stream);");
-                    break;
-                case LoadedImageSurface.LoadedImageSurfaceType.FromUri:
-                    builder.WriteLine($"{_stringifier.Var} result = Windows::UI::Xaml::Media::LoadedImageSurface::StartLoadFromUri(ref new Uri(\"{imageUri}\"));");
-                    break;
+                var parameterString = $"m{_stringifier.MakeVariableName(nodes[i].Name)}";
```
simeoncran: This pattern keeps showing up. Can we abstract it?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -447,5 +632,103 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
...
 
         string Vector2(Vector2 value) => _stringifier.Vector2(value);
+
+        static string IDynamicAnimatedVisualSourceHeaderText(string className, IEnumerable<LoadedImageSurfaceNodes> loadedImageSurfacesNodes)
+        {
+            var nodes = loadedImageSurfacesNodes.ToArray();
+            var imageVariableString = new StringBuilder();
+            foreach (var n in nodes)
+            {
+                imageVariableString.AppendLine($"    {n.TypeName}^ m_{char.ToLowerInvariant(n.Name[0])}{n.Name.Substring(1)};");
```
simeoncran: This looks like the MakeVariableName stuff you added... can it be reused?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -447,5 +632,103 @@ protected override void WriteLoadedImageSurfaceFactory(CodeBuilder builder, Code
...
+    }}
+
+private:
+    const int c_loadedImageSurfaceCount = {loadedImageSurfacesNodes.Distinct().Count()};
+    event Windows::Foundation::TypedEventHandler<IDynamicAnimatedVisualSource^, Platform::Object^>^ m_InternalHandler;
+    int m_loadCompleteEventCount;
+    bool m_isAnimatedVisualSourceDynamic = true;
+    bool m_imageLoadingCompleted;
+    double m_imageSuccessfulLoadingProgress;
+    bool m_tryCreateAnimatedVisualCalled;
```
simeoncran: Any reason that some of these are initialized, but not all of them?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -334,6 +319,29 @@ protected IEnumerable<Uri> GetAssetsList()
...
         }
 
+        /// <summary>
+        /// Gets a list of the <see cref="LoadedImageSurfaceNodes"/> representing the LoadedImageSurface of the composition and its properties.
+        /// </summary>
+        /// <returns>List of the <see cref="LoadedImageSurfaceNodes"/> representing the LoadedImageSurface and its properties.</returns>
+        protected IEnumerable<LoadedImageSurfaceNodes> GetLoadedImageSurfacesNodes()
+        {
+            var loadedImageSurfaces = _nodes.Where(n => n.IsLoadedImageSurface).ToArray();
+            LoadedImageSurfaceNodes[] result = new LoadedImageSurfaceNodes[loadedImageSurfaces.Length];
```
simeoncran: >LoadedImageSurfaceNodes [](start = 12, length = 23)

var

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -334,6 +319,29 @@ protected IEnumerable<Uri> GetAssetsList()
...
+            var i = 0;
+            foreach (var l in loadedImageSurfaces)
+            {
+                result[i].TypeName = l.TypeName;
+                result[i].Name = l.Name;
+                result[i].BytesFieldName = l.LoadedImageSurfaceBytesFieldName;
+                result[i].ImageUri = l.LoadedImageSurfaceImageUri;
+                result[i].LoadedImageSurfaceType = ((Wmd.LoadedImageSurface)l.Object).Type;
+                i++;
+            }
```
simeoncran: This method can all be simplified to a linq query or a simple .Select on the loadedImageSurfaces.
return _nodes.Where(n=> n.IsLoadedImageSurface).Select(n => new LoadedImageSurfaceNodes{TypeName = n.TypeName, Name = n.Name,  ...etc
or

return
   from n in _nodes
   where n.IsLoadedImageSurface
   select new LoadedImageSurfaceNodes{ TypeName = n.TypeName, Name = n.Name, ... etc

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2164,5 +2203,15 @@ internal string TypeName
             // Sets the first character to lower case.
             static string CamelCase(string value) => $"_{char.ToLowerInvariant(value[0])}{value.Substring(1)}";
         }
+
+        // A struct for representing the LoadedImageSurface objects in the composition.
+        protected internal struct LoadedImageSurfaceNodes
```
simeoncran: >LoadedImageSurfaceNodes [](start = 34, length = 23)

Why plural?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2164,5 +2203,15 @@ internal string TypeName
             // Sets the first character to lower case.
             static string CamelCase(string value) => $"_{char.ToLowerInvariant(value[0])}{value.Substring(1)}";
         }
+
+        // A struct for representing the LoadedImageSurface objects in the composition.
```
simeoncran: Comment should be a doc comment so it shows up in intellisense.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: >LoadedImageSurfaceNodes [](start = 34, length = 23)

Convert to properties (i.e. add { get; set; } to each field.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1514,7 +1534,7 @@ bool GenerateSpriteShapeFactory(CodeBuilder builder, CompositionSpriteShape obj,
             return true;
         }
 
-        bool GenerateCompositionSurfaceBrushFactory(CodeBuilder builder, CompositionSurfaceBrush obj, ObjectData node)
+        bool GenerateCompositionSurfaceBrushFactory(CodeBuilder builder, CompositionSurfaceBrush obj, ObjectData node, CodeGenInfo info)
```
simeoncran: >odeGenInfo info [](start = 120, length = 15)

Not used.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -676,7 +696,7 @@ bool GenerateCanvasGeometryFactory(CodeBuilder builder, CanvasGeometry obj, Obje
             return true;
         }
 
-        bool GenerateObjectFactory(CodeBuilder builder, CompositionObject obj, ObjectData node)
+        bool GenerateObjectFactory(CodeBuilder builder, CodeGenInfo info, CompositionObject obj, ObjectData node)
```
simeoncran: >odeGenInfo info [](start = 57, length = 15)

Not needed

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -625,7 +645,7 @@ void WriteCodeForNode(CodeBuilder builder, ObjectData node, CodeGenInfo info)
                 switch (node.Type)
                 {
                     case Graph.NodeType.CompositionObject:
-                        GenerateObjectFactory(builder, (CompositionObject)node.Object, node);
+                        GenerateObjectFactory(builder, info, (CompositionObject)node.Object, node);
                         break;
```
simeoncran: >break [](start = 24, length = 5)

While you're here, the fact that some of these are breaks and some are returns is bad... please make them all breaks.

source/UIData/CodeGen/NodeNamer.cs
```diff
@@ -35,6 +36,7 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
             {
                 string baseName;
 
+                // Generate descriptive name for each object. The name is as specific as possible.
```
simeoncran: >The name is as specific as possible. [](start = 62, length = 36)

I'm not convinced that this comment would have solved the original misreading. Would it have helped you?

source/UIData/CodeGen/CxInstantiatorGenerator.cs
plaiMS: Added a new function for this pattern because simply MakeVariableName() without the 'm' prefix is called in this file as well.

---
In reply to: [294942308](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r294942308) [](ancestors = 294942308)

source/UIData/CodeGen/CodeBuilder.cs
plaiMS: It has the same effect.

---
In reply to: [294934976](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r294934976) [](ancestors = 294934976)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -2040,7 +2065,7 @@ sealed class ObjectData : Graph.Node<ObjectData>
 
             public string Name { get; set; }
 
-            public string FieldName => RequiresStorage ? CamelCase(Name) : null;
+            public string FieldName => RequiresStorage || IsLoadedImageSurface ? CamelCase(Name) : null;
```
simeoncran: Seems like this test should only be for RequiresStorage. If it doesn't require storage then there's no need for a field. Or is this a different type of storage?

plaiMS: LoadedImageSurface also needs a field. So I'm reusing an existing node property to create/hold the field name.

---
In reply to: [295447871](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r295447871) [](ancestors = 295447871)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -416,24 +476,129 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
+            builder.OpenScope();
+            builder.WriteLine("return nullptr;");
+            builder.CloseScope();
+            builder.WriteLine("return ref new AnimatedVisual(compositor,");
+            builder.Indent();
+
+            var nodes = info.LoadedImageSurfaceNodes.ToArray();
+            for (var i = 0; i < nodes.Length; i++)
+            {
+                var parameterString = $"{MakeClassMember(nodes[i].Name)}";
```
simeoncran: The $"" is redundant here.
The name isn't right. A "member" could mean many things. I'm not sure what the intention is.
Also why "parameterString" rather than paramterName or fieldName, or whatever it represents. Knowing that it's a string doesn't help the reader know the intention.

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: Why not use linq syntax. The way it's written it looks like linq syntax, it just isn't.

---
In reply to: [294945297](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r294945297) [](ancestors = 294945297)

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
plaiMS: Have another property RequireReadonlyStorage and set LoadedImageSurface to both RequireReadonlyStorage  and RequiresStorage,

---
In reply to: [295450615](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r295450615) [](ancestors = 295450615,295447871)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
simeoncran: What would be the difference between MakeMemberVariableName and MakeFieldName? I'm thinking it's about creating a name that is suitable for a (non-static) field. 

---
In reply to: [295451245](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r295451245) [](ancestors = 295451245)

source/UIData/CodeGen/CxInstantiatorGenerator.cs
simeoncran: The pattern I was referring to was creating a comma separated list of parameters that are on separate lines. The code is awkward and is used in at least 2 places.
On CodeBuilder, have a method: WriteCommaSeparatedLines(IEnumerable<string> items). Don't worry about the closing parenthesis - it probably looks fine on the next line. That would also allow the semicolon to be on the line with the closing parens.

---
In reply to: [295030942](https://github.com/windows-toolkit/Lottie-Windows/pull/134#discussion_r295030942) [](ancestors = 295030942,294942308)

PR #143, opened 2019-08-28
Prepare for gradients

simeoncran:
Changes needed to implement gradient support.
Includes some refactoring and cleanup that I want to get out of the way now rather than including in the more complicated stuff needed to translate gradients.


eliezerpMS:
> #if POST_RS5_SDK

You can probably also delete this and all its uses in this file

---
Refers to: source/LottieToWinComp/LottieToWinCompTranslator.cs:21 in 1230b0a. [](commit_id = 1230b0a2e285a13ee000a6b62a15c6156932a0b9, deletion_comment = False)

PR #145, opened 2019-08-30
Enabling simple LinearGradientFill cases.

simeoncran:
This is a work in progress - fill in more pieces to enable gradients. Some simple cases work.


source/UIData/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1127,10 +1127,10 @@ void InitializeCompositionGradientBrush(CodeBuilder builder, CompositionGradient
                 builder.WriteLine($"result{Deref}InterpolationSpace = TODO");
             }
 
-            // TODO - mappingmode
-            if (obj.MappingMode.HasValue)
+            // Default MappingMode is Relative
+            if (obj.MappingMode.HasValue && obj.MappingMode.Value != WinCompData.CompositionMappingMode.Relative)
             {
-                builder.WriteLine($"result{Deref}MapingMode = TODO");
+                builder.WriteLine($"result{Deref}MapingMode = {CompositionMappingMode(obj.MappingMode.Value)};");
```
eliezerpMS: >MapingMode [](start = 49, length = 10)

MappingMode

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2553,6 +2569,42 @@ CompositionColorBrush TranslateShapeFill(TranslationContext context, SolidColorF
...
+            // BodyMovin specifies start and end points in absolute values.
+            result.MappingMode = CompositionMappingMode.Absolute;
+
+            result.StartPoint = Vector2(context.TrimAnimatable(shapeFill.StartPoint).InitialValue);
+            result.EndPoint = Vector2(context.TrimAnimatable(shapeFill.EndPoint).InitialValue);
+
+            var colorStops = result.ColorStops;
+            foreach (var stop in stops.InitialValue.Items)
+            {
+                var offset = stop.Offset;
```
eliezerpMS: >var offset = stop.Offset; [](start = 16, length = 25)

nit: why not just pass this to the line below directly?

source/UIData/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: Didn't see your comment. Will be fixed when the rest of the codegen is fixed.

---
In reply to: [319324578](https://github.com/windows-toolkit/Lottie-Windows/pull/145#discussion_r319324578) [](ancestors = 319324578)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Didn't see your comment. The idea was to be consistent with the other parameter (color), although I could be persuaded either way.

---
In reply to: [319325326](https://github.com/windows-toolkit/Lottie-Windows/pull/145#discussion_r319325326) [](ancestors = 319325326)

PR #146, opened 2019-08-31
Enable more linear gradient fill cases, including animated linear gradients.

simeoncran:
Linear gradient fills should work fine for common cases now.


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2579,32 +2579,114 @@ CompositionLinearGradientBrush TranslateLinearGradientFill(TranslationContext co
...
+            if (stops.IsAnimated)
+            {
+                // Lottie represents animation of color stops as a sequence of lists of stops.
+                // WinComp uses a single list of stops where each stop is animated.
+                var stopsCount = stops.InitialValue.Items.Length;
+                var keyframesCount = stops.KeyFrames.Length;
+
+                var gradientStopKeyFrames = stops.KeyFrames.ToArray();
+
+                // Create the Composition stops and animated them.
```
eliezerpMS: >animated [](start = 52, length = 8)

nit typo: animate

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2579,32 +2579,114 @@ CompositionLinearGradientBrush TranslateLinearGradientFill(TranslationContext co
...
+                    }
+
+                    colorStops.Add(_c.CreateColorGradientStop(Float(offset), color));
+                }
             }
 
             return result;
         }
 
+        static IEnumerable<KeyFrame<T>> ExtractKeyFramesFromGradientStopKeyFrames<T>(KeyFrame<Sequence<GradientStop>>[] stops, int stopIndex, Func<GradientStop, T> selector)
```
eliezerpMS: > static IEnumerable<KeyFrame<T>> ExtractKeyFramesFromGradientStopKeyFrames<T>(KeyFrame<Sequence<GradientStop>>[] stops, int stopIndex, Func<GradientStop, T> selector) [](start = 7, length = 166)

Probably make this line shorter

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2579,32 +2579,114 @@ CompositionLinearGradientBrush TranslateLinearGradientFill(TranslationContext co
...
+
+                var gradientStopKeyFrames = stops.KeyFrames.ToArray();
+
+                // Create the Composition stops and animated them.
+                for (var i = 0; i < stopsCount; i++)
+                {
+                    var gradientStop = _c.CreateColorGradientStop();
+                    colorStops.Add(gradientStop);
+
+                    var colorKeyFrames = ExtractKeyFramesFromGradientStopKeyFrames(gradientStopKeyFrames, i, gs => MultiplyColorByOpacityPercent(MultiplyColorByOpacityPercent(gs.Color, gs.OpacityPercent), opacityPercentValue)).ToArray();
```
eliezerpMS: >var colorKeyFrames = ExtractKeyFramesFromGradientStopKeyFrames(gradientStopKeyFrames, i, gs => MultiplyColorByOpacityPercent(MultiplyColorByOpacityPercent(gs.Color, gs.OpacityPercent), opacityPercentValue)).ToArray(); [](start = 20, length = 217)

should make this line not as long

eliezerpMS: >MultiplyColorByOpacityPercent(MultiplyColorByOpacityPercent(gs.Color, gs.OpacityPercent), opacityPercentValue) [](start = 115, length = 110)

Make a function for the MultiplyColorByOpacityPercent passed to MultiplyColorByOpacityPercent perhaps along with a comment since the idea of propagating opacity in addition to the opacity on the gradient stops is something we should call out that we are supporting.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2579,32 +2579,114 @@ CompositionLinearGradientBrush TranslateLinearGradientFill(TranslationContext co
...
-                var offset = stop.Offset;
-                var color = stop.Opacity.HasValue ? MultiplyColorByOpacityPercent(stop.Color, stop.Opacity.Value) : stop.Color;
-                colorStops.Add(_c.CreateColorGradientStop(Float(offset), color));
+                foreach (var stop in stops.InitialValue.Items)
+                {
+                    var offset = stop.Offset;
+                    var color = MultiplyColorByOpacityPercent(MultiplyColorByOpacityPercent(stop.Color, stop.OpacityPercent), opacityPercentValue);
+                    if (color == null)
+                    {
+                        // We currently do not support gradients that contain separate opacity stops.
```
eliezerpMS: >// We currently do not support gradients that contain separate opacity stops. [](start = 24, length = 77)

Suggestion: Maybe this comment can read "We currently do not support gradient stops that are opacity or color only"

From reviewing what MultiplyColorByOpacityPercent does and I assume that Lotties could have stops that are opacity but not color as well?

simeoncran: Updated. We do support color only. It's the opacity-only ones that we can't support (because we don't know what color to use). Eventually we will handle this by creating a ShapeVisual per Shape/Group and putting the opacity on the ShapeVisual

---
In reply to: [319705548](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r319705548) [](ancestors = 319705548)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Not sure what you are asking for with the comment.

---
In reply to: [319705117](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r319705117) [](ancestors = 319705117)

source/LottieData/ColorGradientStop.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                throw new ArgumentException("Color must have an alpha of 1");
+            }
+
+            Offset = offset;
+            Color = color;
+        }
+
+        public readonly Color Color;
+        public readonly double Offset;
```
eliezerpMS: >public readonly double Offset; [](start = 8, length = 30)

I know we don't gain much but would it make sense to have a base class that is GradientStop and knows how to print the offset and then you just re-use that when doing the ToString

simeoncran: The only thing that's the same in the ToString is the @{Offset} part. The ColorGradientStop and OpacityGradientStop have no relationship with each other, i.e. we never want to treat them as just of type GradientStop.

---
In reply to: [321464452](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r321464452) [](ancestors = 321464452)

source/LottieData/LinearGradientFill.cs
```diff
@@ -15,19 +15,23 @@ sealed class LinearGradientFill : ShapeFill
...
             IAnimatableVector3 endPoint,
-            Animatable<Sequence<GradientStop>> gradientStops)
+            Animatable<Sequence<ColorGradientStop>> colorStops,
+            Animatable<Sequence<OpacityGradientStop>> opacityPercentStops)
             : base(in args, fillType, opacityPercent)
         {
             StartPoint = startPoint;
             EndPoint = endPoint;
-            GradientStops = gradientStops;
+            ColorStops = colorStops;
```
eliezerpMS: >ColorStops = colorStops; [](start = 12, length = 24)

It is legal to not have any Color stops? (this should be totally legal in most art programs, but might be weird for our scenarios unless we use the gradient in masks)

eliezerpMS: Although I could also see that we want to just store the data and do WinComp related support checks in the translator in which case you can ignore this comment.

---
In reply to: [321465656](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r321465656) [](ancestors = 321465656)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2571,40 +2571,121 @@ CompositionColorBrush TranslateSolidColorFill(TranslationContext context, SolidC
...
+                // If there are no color stops then we can't create a brush.
                 return null;
             }
 
-            // TODO - support animated start and end points, and color stops.
-            if (shapeFill.StartPoint.IsAnimated || shapeFill.EndPoint.IsAnimated || stops.IsAnimated)
+            if (opacityPercent.IsAnimated)
             {
+                // We don't yet support animated opacity with LinearGradientFill.
+                _issues.GradientFillIsNotSupported();
```
eliezerpMS: >GradientFillIsNotSupported [](start = 24, length = 26)

should this error message maybe be a bit more specific?

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: Comment is no longer relevant with the new checks at the top of this function.

---
In reply to: [320396945](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r320396945) [](ancestors = 320396945,319705117)

source/LottieData/LinearGradientFill.cs
simeoncran: If someone gives us a silly Lottie file, they'll get silly output. AFAICT there's nothing illegal about opacity only.

---
In reply to: [321468369](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r321468369) [](ancestors = 321468369,321465656)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: There are quite a few cases that we can't get quite right at the moment with gradients, so rather than trying to enumerate them all, I figure we can at least point out that something isn't right with gradients.
When it becomes clear what we definitely can or cannot support, then it might be worth making these more specific, but for now, to make progress, I'll keep it general.

---
In reply to: [321468727](https://github.com/windows-toolkit/Lottie-Windows/pull/146#discussion_r321468727) [](ancestors = 321468727)

PR #154, opened 2019-09-13
Improve many masks cases, especially combinations of Add and Subtract…

simeoncran:
… masks.

### DIFF
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -482,7 +482,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         }
 
         // Enumerates the segments of Masks with the same MaskMode.
-        IEnumerable<(int index, int count)> EnumerateMaskListSegments(Mask[] masks)
+        static IEnumerable<(int index, int count)> EnumerateMaskListSegments(Mask[] masks)
         {
             int i;
 
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -339,92 +339,77 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             CompositionContainerShape containerShapeToMask)
         {
             var contentShapeVisual = _c.CreateShapeVisualWithChild(containerShapeToMask, context.Size);
-            return TranslateAndApplyMasks(context, context.Layer, contentShapeVisual);
+            return TranslateAndApplyMasksForLayer(context, context.Layer, contentShapeVisual);
         }
 
-        // This function is used to take the paths for the masks defined on a layer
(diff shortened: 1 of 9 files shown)


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -339,135 +339,221 @@ IEnumerable<ShapeOrVisual> ComposeMattedLayers(TranslationContext context, IEnum
...
+                    }
+                }
+            }
+
+            return result;
+        }
 
-                var maskShapeVisual = _c.CreateShapeVisualWithChild(containerShapeMaskRootNode, context.Size);
+        // Enumerates the segments of Masks with the same MaskMode.
+        IEnumerable<(int index, int count)> EnumerateMaskListSegments(Mask[] masks)
```
simeoncran: >IEnumerable [](start = 8, length = 11)

Can be static

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -339,135 +339,221 @@ IEnumerable<ShapeOrVisual> ComposeMattedLayers(TranslationContext context, IEnum
...
-                    return layerRootContainerVisual;
+                    currentMode = mode;
+                    segmentIndex = i;
                 }
             }
 
-            // Currently no callers do anything useful if this function fails to apply the mask.
-            // In the event of failure, return the visual that was supposed to be masked.
-            return visualToBeMasked;
+            // Output the last segment it it's not empty.
```
eliezerpMS: >it it's [](start = 39, length = 7)

Typo?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -339,135 +339,221 @@ IEnumerable<ShapeOrVisual> ComposeMattedLayers(TranslationContext context, IEnum
...
+
+                        default:
+                            _issues.MaskWithUnsupportedMode(masks[0].Mode.ToString());
+                            break;
+                    }
                 }
+                else
+                {
+                    // Uncommon case: multiple masks. Get the contiguous segments of masks that have the same mode.
+                    foreach (var maskSegmentWithSameMode in EnumerateMaskListSegments(layer.Masks.ToArray()))
```
eliezerpMS: >foreach (var maskSegmentWithSameMode in EnumerateMaskListSegments(layer.Masks.ToArray())) [](start = 20, length = 89)

I would not mind another comment here about what is being done. I understand it now because you explained it to me in person before. You could have a comment about how you are walking the list of masks and grouping contiguous masks that have the same mode. You then apply all the groups one by one and get the end composite result.

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Stutter

---
In reply to: [325382328](https://github.com/windows-toolkit/Lottie-Windows/pull/154#discussion_r325382328) [](ancestors = 325382328)

PR #156, opened 2019-10-15
Prerequisites for versioning.

simeoncran:
This adds a number of pieces required to generate version-adaptive code, but doesn't
yet hook those pieces up. This allows testing against a corpus to ensure nothing is
broken. A future commit will add version-adaptivity.

### DIFF
--- a/LottieGen/CommandLineOptions.cs
+++ b/LottieGen/CommandLineOptions.cs
@@ -45,6 +45,10 @@ sealed class CommandLineOptions
 
     internal bool DisableCodeGenOptimizer { get; private set; }
 
+    internal uint? MinimumUapVersion { get; private set; }
+
+    internal uint? TargetUapVersion { get; private set; }
+
     enum Keyword
     {
         None,
@@ -56,6 +60,8 @@ sealed class CommandLineOptions
         Strict,
         DisableTranslationOptimizer,
         DisableCodeGenOptimizer,
+        MinimumUapVersion,
+        TargetUapVersion,
     }
 
     // Returns the parsed command line. If ErrorDescription is non-null, then the parse failed.
@@ -110,7 +116,9 @@ sealed class CommandLineOptions
             .AddPrefixedKeyword("outputfolder", Keyword.OutputFolder)
             .AddPrefixedKeyword("strict", Keyword.Strict)
             .AddPrefixedKeyword("disablecodegenoptimizer", Keyword.DisableCodeGenOptimizer)
-            .AddPrefixedKeyword("disabletranslationoptimizer", Keyword.DisableTranslationOptimizer);
+            .AddPrefixedKeyword("disabletranslationoptimizer", Keyword.DisableTranslationOptimizer)
+            .AddPrefixedKeyword("minimumuapversion", Keyword.MinimumUapVersion)
+            .AddPrefixedKeyword("targetuapversion", Keyword.TargetUapVersion);
 
         // The last keyword recognized. This defines what the following parameter value is for,
         // or None if not expecting a parameter value.
@@ -149,6 +157,8 @@ sealed class CommandLineOptions
                         case Keyword.InputFile:
                         case Keyword.Language:
                         case Keyword.OutputFolder:
+                        case Keyword.MinimumUapVersion:
+                        case Keyword.TargetUapVersion:
                             previousKeyword = keyword;
                             break;
                         default:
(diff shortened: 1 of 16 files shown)


LottieGen/CommandLineOptions.cs
```diff
@@ -173,11 +183,47 @@ void ParseCommandLineStrings(string[] args)
...
+                    if (MinimumUapVersion != null)
+                    {
+                        ErrorDescription = ArgumentSpecifiedMoreThanOnce("minimum UAP version");
+                        return;
+                    }
+
+                    {
+                        if (!uint.TryParse(arg, out var version))
+                        {
+                            ErrorDescription = "minimum UAP version must be an positive integer";
```
eliezerpMS: >minimum UAP version must be an positive integer [](start = 48, length = 47)

Nit: Is this the established format for these? (First word not capitalized and no ending period) If you had the capitalization and period then you could more easily have them on the same line in the future if needed?

LottieGen/Program.cs
```diff
@@ -67,6 +67,28 @@ RunResult Run()
             _reporter.ErrorStream.WriteLine(_options.ErrorDescription);
             return RunResult.InvalidUsage;
         }
+        else if (_options.MinimumUapVersion.HasValue && _options.MinimumUapVersion < 7)
```
eliezerpMS: >7 [](start = 85, length = 1)

In your next pass at versioning I assume you will define this 7 somewhere global?

LottieGen/Program.cs
```diff
@@ -67,6 +67,28 @@ RunResult Run()
...
+        {
+            // Unacceptable version.
+            _reporter.WriteError("Minimum UAP version must be 7 or above.");
```
eliezerpMS: >_reporter.WriteError("Minimum UAP version must be 7 or above."); [](start = 12, length = 64)

Nit: maybe these error messages could say the version that was passed in. This would help for the scenarios where one person wrote the LottieGen automation script and another person needs to figure out later why things are broken.

LottieGen/Program.cs
```diff
@@ -217,6 +239,18 @@ static void ShowUsage(TextWriter writer)
...
+                       to 7. Must be 7 or higher. Code will be generated that will
+                       run down to this version. If less than TargetUapVersion,
+                       extra code will be generated if necessary to support the
+                       lower versions.
+         -TargetUapVersion
+                       The target UAP version on which the result will run. Must be 7
+                       or higher and >= MinimumUapVersion. Code will be generated
+                       that may take advantage of features in this version in order
+                       to produce a better result. If not specified, defaults to 
+                       the latest UAP version.
```
eliezerpMS: >the latest UAP version. [](start = 23, length = 23)

Do you mean the newest UAP version the OS supports or the newest version on the newest released version of the OS?

simeoncran: Discussed in person. The default acts as if you had specified the latest published version of the UAP contract. That doesn't mean that LottieGen knows about that latest version... it currently only knows up to version 8. But when 9 gets released, specifying 9 here would produce code that uses features in version 8, which is also what happens if you rely on the default.
I don't think we can make this clearer while keeping it concise, and I'm guessing it is ok as is.

---
In reply to: [335223414](https://github.com/windows-toolkit/Lottie-Windows/pull/156#discussion_r335223414) [](ancestors = 335223414)

LottieGen/CommandLineOptions.cs
simeoncran: I tried to make it more consistent, and got it half right. Will do what you suggest (capitalized and ending period). Note that we only output the first error during parsing of the command line anyway.

---
In reply to: [335220830](https://github.com/windows-toolkit/Lottie-Windows/pull/156#discussion_r335220830) [](ancestors = 335220830)

source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -150,7 +150,10 @@ internal ColorKeyFrameAnimation CreateColorKeyFrameAnimation()
 
         internal InsetClip CreateInsetClip() => _compositor.CreateInsetClip();
 
-        internal CompositionGeometricClip CreateGeometricClip() => _compositor.CreateGeometricClip();
+        internal CompositionGeometricClip CreateGeometricClip()
+        {
+            return _compositor.CreateGeometricClip();
+        }
```
eliezerpMS: Should this one be switched back as well?

simeoncran: Yes.

---
In reply to: [335240886](https://github.com/windows-toolkit/Lottie-Windows/pull/156#discussion_r335240886) [](ancestors = 335240886)

PR #163, opened 2019-10-24
Define a more sensible sorting order for elements in codegen

simeoncran:
Order codegen elements in a way that humans would expect.

This deals better than the old ordering with names that have numbers embedded in them,
for example img_001, img_11.

The old ordering was particularly bad for images provided by designers, resulting in code that looked like it had things in the wrong order.


source/UIData/CodeGen/AlphanumericStringComparer.cs
```diff
@@ -0,0 +1,135 @@
...
+            for (int xi = 0, yi = 0; xi < x.Length && yi < y.Length; xi++, yi++)
+            {
+                var chX = x[xi];
+                var chY = y[yi];
+
+                // Can both sides be interpreted as hex or decimal integers?
+                if (IsHexDigit(chX) && IsHexDigit(chY))
+                {
+                    // Try to match as upper-case hex.
+                    var matchX = s_upperHexRecognizer.Match(x, xi);
```
eliezerpMS: >s_upperHexRecognizer.Match(x, xi); [](start = 33, length = 34)

Maybe I am misunderstanding this but is this line just checking to see if this char matches the regex? If so maybe update the comment?

eliezerpMS: Also should you have it only try to match that one char if this is the case?

---
In reply to: [338685834](https://github.com/windows-toolkit/Lottie-Windows/pull/163#discussion_r338685834) [](ancestors = 338685834)

source/UIData/CodeGen/AlphanumericStringComparer.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            if (y is null)
+            {
+                return 1;
+            }
+
+            var xLength = x.Length;
+            var yLength = y.Length;
+            for (int xi = 0, yi = 0; xi < x.Length && yi < y.Length; xi++, yi++)
```
eliezerpMS: > for (int xi = 0, yi = 0; xi < x.Length && yi < y.Length; xi++, yi++) [](start = 11, length = 69)

I would not mind a comment summarizing the general idea of this for loop.

PR #164, opened 2019-10-24
Update nuget dependencies to latest non-prerelease versions.

simeoncran:


samples/LottieSamples/LottieSamples.csproj
```diff
@@ -235,13 +235,13 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>6.1.9</Version>
+      <Version>6.1.7</Version>
```
simeoncran: Hmmm.. that was unexpected. Looks like this should really be 6.2.9. Will fix.

PR #176, opened 2019-10-30
Anti-alias the edges of masks and mattes for a smoother result.

simeoncran:
The default BorderMode for Visuals is Inherit, and if not specified the value Hard (i.e. not anti-aliased) is passed into the root.
The default worked for us previously because the desktop tree sets the BorderMode to Soft and we inherit it so we got Soft behavior everywhere.
However the trees created for masks and mattes are not rooted in the desktop, so they were getting set to Hard and some edges looked aliased.

This change adds the ability to set the BorderMode and sets it to Soft for the trees created for masks and mattes.

### DIFF
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -670,9 +670,21 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             // the visual we want captured by the visual surface has a parent to use as the
             // origin of its offsets.
             var sourceIntermediateParent = _c.CreateContainerVisual();
+
+            // Because this the root of a tree, the BorderMode is not inherited and
+            // will default to Hard. We want it to be Soft in order to enable anti-aliasing.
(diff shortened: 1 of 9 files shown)


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -670,9 +670,21 @@ Visual TranslateAndApplyMasks(TranslationContext context, ReadOnlySpan<Mask> mas
             // the visual we want captured by the visual surface has a parent to use as the
             // origin of its offsets.
             var sourceIntermediateParent = _c.CreateContainerVisual();
+
+            // Because this the root of a tree, the BorderMode is not inherited and
```
simeoncran: typo, here and below.

source/WinCompData/Visual.cs
```diff
@@ -16,6 +16,10 @@ private protected Visual()
         {
         }
 
+        // Defaults to Inherit. For trees rooted are set to Soft.
```
simeoncran: typo

PR #181, opened 2019-11-02
Add support for gradient strokes.

simeoncran:


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2653,30 +2768,26 @@ CompositionBrush TranslateShapeFill(TranslationContext context, ShapeFill shapeF
...
-                    context.TrimAnimatable(shapeFill.Color),
-                    context.TrimAnimatable(shapeFill.OpacityPercent)),
-                opacityPercent);
+                    context.TrimAnimatable(color),
+                    context.TrimAnimatable(opacityPercentA)),
+                opacityPercentB);
         }
 
+        CompositionColorBrush TranslateSolidColorFill(TranslationContext context, SolidColorFill shapeFill, TrimmedAnimatable<double> opacityPercent)
+            => TranslateSolidColor(context, shapeFill.Color, shapeFill.OpacityPercent, opacityPercent);
```
scott-moore-ms: I'm not up on the latest C# features - is there a functional difference between this syntax and

CompositionColorBrush TranslateSolidColorFill(TranslationContext context, SolidColorFill shapeFill, TrimmedAnimatable<double> opacityPercent)
{
    return TranslateSolidColor(context, shapeFill.Color, shapeFill.OpacityPercent, opacityPercent);
}

?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2776,73 +2970,23 @@ CompositionColorBrush TranslateSolidColorFill(TranslationContext context, SolidC
...
         {
-            var colorStops = context.TrimAnimatable(shapeFill.ColorStops);
-            var opacityPercentStops = context.TrimAnimatable(shapeFill.OpacityPercentStops);
-
-            if (colorStops.InitialValue.Items.IsEmpty)
+            if (!IsUapApiAvailable(nameof(CompositionRadialGradientBrush), versionDependentFeatureDescription: "Radial gradient fill"))
             {
-                // If there are no color stops then we can't create a brush.
-                return null;
+                return TranslateLinearGradient(context, gradient, opacityPercentValue);
```
eliezerpMS: >return TranslateLinearGradient(context, gradient, opacityPercentValue); [](start = 16, length = 71)

I would not mind a comment here just spelling out that the code will treat the radial gradient as a linear gradient as a fallback (maybe so that a reader of the code many years from now trying to reason about what they are seeing can more quickly trace through what is happening)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: None, they're equivalent. I like it for space saving, but also because it aligns with my desire to reason about code in a declarative way rather than as a set of steps.
In this case it's a teensy bit of code that is shared by a couple callers and the less attention that occupies while I'm reading, the better.

---
In reply to: [342799450](https://github.com/windows-toolkit/Lottie-Windows/pull/181#discussion_r342799450) [](ancestors = 342799450)

PR #184, opened 2019-11-13
Fix bug where combining of strokes could fail during translation.

simeoncran:
We were correctly recognizing that we couldn't compose strokes of different kinds, but then we tried to compose them anyway which would result in a failed cast.


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -1400,6 +1400,7 @@ ShapeStroke ComposeStrokes(ShapeStroke a, ShapeStroke b)
                 if (a.StrokeKind != b.StrokeKind)
                 {
                     _owner._issues.MultipleStrokesIsNotSupported();
+                    return b;
```
scott-moore-ms: >return b; [](start = 20, length = 9)

whoopsies =)

PR #190, opened 2019-11-20
Make parsing resilient to issues seen in our corpus.

simeoncran:
Handle unexpected matte types by parsing as "Unknown" and treating them as Add.
Allow parsing errors in layers to result in the layer being ignored rather than giving up. Note that this can result in an issue later on if the layer is searched for but not found.
Handle case of a layer with inPoint == outPoint by ignoring that layer.
Fix comparsion of CubicBezierEasings - equivalent CubicBezierEasings were being reported as being non-equal due to missing == operator overloads.
Handle the case of multiple color gradient stops at the same offset - always use the last one. Designers might do this because in one keyframe the stops are all bunched up against offset 0 or 1, then later they get spread out.
In general, throw exceptions from the parser less often.
Parse and save any extra top-level data in the json file. I plan to expose this in generated code and LottieVisualSource in a later check-in.


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -33,6 +33,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Xml.Serialization;
```
simeoncran: Hmmm... that doesn't look right.

source/LottieData/Easing.cs
```diff
@@ -44,12 +44,22 @@ public bool Equals(Easing other)
...
                     var xCb = (CubicBezierEasing)this;
                     var yCb = (CubicBezierEasing)other;
-                    return xCb.ControlPoint1.Equals(yCb.ControlPoint1);
+                    return xCb.Equals(yCb);
                 default:
                     throw new InvalidOperationException();
             }
         }
 
+        public static bool operator ==(Easing a, Easing b) => a is Easing && a.Equals(b);
```
eliezerpMS: >a is Easing [](start = 62, length = 11)

For my own learning: why do the "is Easing" here?

source/LottieData/Optimization/GradientStopOptimizer.cs
```diff
@@ -76,7 +76,7 @@ public static IEnumerable<ColorGradientStop> Optimize(IEnumerable<GradientStop>
                     nextColorStop = null;
 
                     // If there's an opacity stop at the same offset, it will be the next stop in the list.
-                    if (i + 1 < orderedStops.Length && orderedStops[i + 1].Offset == currentStop.Offset)
+                    if (i + 1 < orderedStops.Length && orderedStops[i + 1].Offset == currentStop.Offset && orderedStops[i + 1].Kind == GradientStop.GradientStopKind.Opacity)
```
eliezerpMS: >if (i + 1 < orderedStops.Length && orderedStops[i + 1].Offset == currentStop.Offset && orderedStops[i + 1].Kind == GradientStop.GradientStopKind.Opacity) [](start = 20, length = 153)

Nit: wrap if statement

eliezerpMS: Also sorry for not catching this case in the previous review

---
In reply to: [348668490](https://github.com/windows-toolkit/Lottie-Windows/pull/190#discussion_r348668490) [](ancestors = 348668490)

source/LottieData/Easing.cs
simeoncran: This is the new hotness. Now that C# supports "is null" for null checking, "is Easing" is a way of expression "!= null". But the "is" pattern has the advantage of not being hijacked by operator overloading. Previously you'd need to do object.ReferenceEquals(a, null) to guarantee that you were really doing a test for null. You could also do "is object"

---
In reply to: [348667405](https://github.com/windows-toolkit/Lottie-Windows/pull/190#discussion_r348667405) [](ancestors = 348667405)

PR #191, opened 2019-11-21
Add a cast to eliminate a compiler warning.

simeoncran:
std::vector returns a size_t from size(). We assign it to a UINT which results in a warning. Adding a cast to make that warning go away.

### DIFF
--- a/source/UIData/CodeGen/CxInstantiatorGenerator.cs
+++ b/source/UIData/CodeGen/CxInstantiatorGenerator.cs
@@ -869,7 +869,7 @@ public:
     {
         if (count != nullptr)
         {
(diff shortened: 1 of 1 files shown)


source/UIData/CodeGen/CxInstantiatorGenerator.cs
```diff
@@ -869,7 +869,7 @@ void AddSource(IGraphicsEffectSource* source)
     {
         if (count != nullptr)
         {
-            *count = m_sources.size();
+            *count = (UINT)m_sources.size();
```
eliezerpMS: >(UINT) [](start = 21, length = 6)

nit: static_cast?

PR #193, opened 2019-11-22
Remove workaround for CompositionRectangleGeometry when running on UAP v8.

simeoncran:
Now that we have version-dependent translation, we can replace the workaround for CompositionRectangleGeometry (which misbehaves in RS5) with version-dependent behavior. If on v8 or beyond, we will use CompositionRectangleGeometry instead of CompositionRoundedRectangleGeometry with tiny corners.

### DIFF
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -2,9 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-// Enable workaround for RS5 where rotated rectangles were not drawn correctly.
-#define WorkAroundRectangleGeometryHalfDrawn
-
 // Use the simple algorithm for combining trim paths. We're not sure of the correct semantics
 // for multiple trim paths, so it's possible this is actually the most correct.
 #define SimpleTrimPathCombining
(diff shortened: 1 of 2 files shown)


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2199,26 +2199,46 @@ CompositionShape TranslateRectangleContent(TranslationContext context, ShapeCont
...
+                    // Convert size and position into offset. This is necessary because a geometry's offset is for
+                    // its top left corner, wherease a Lottie position is for its centerpoint.
+                    rectangleGeometry.Offset = Vector2(position.InitialValue - (size.InitialValue / 2));
+
+                    if (!size.IsAnimated)
+                    {
+                        rectangleGeometry.Size = Vector2(size.InitialValue);
+                    }
+
+                    compositionRectangle.Geometry = geometry;
```
eliezerpMS: >compositionRectangle.Geometry = geometry; [](start = 20, length = 41)

nit: does this need to be here if it is also on the line below?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2199,26 +2199,46 @@ CompositionShape TranslateRectangleContent(TranslationContext context, ShapeCont
...
-                var geometry = _c.CreateRectangleGeometry();
-#endif
-                compositionRectangle.Geometry = geometry;
+                if (_targetUapVersion <= 7)
+                {
+                    // V7 did not reliably draw non-rounded rectangles.
+                    // Work around the problem by using a rounded rectangle with a tiny corner radius.
+                    var roundedRectangleGeometry = _c.CreateRoundedRectangleGeometry();
+                    geometry = roundedRectangleGeometry;
+                    roundedRectangleGeometry.CornerRadius = new Sn.Vector2(0.000001F, 0.000001F);
```
eliezerpMS: >0.000001 [](start = 75, length = 8)

Just curious, how did you choose this number? Is it the case that as long as we specify something for CornerRadius then we ok?

PR #194, opened 2019-12-02
Adding a metadata facility to CompositionObject.

simeoncran:
This adds a facility for arbitrary metadata to be added to CompositionObject and uses it for the IDescribable interface.
The metadata facility is being added so that we can smuggle information from names in LottieData through to the optimizer and code generator.
I also made the IDescribable implementation on CompositionObject explicitly implemented so that it doesn't pollute the API surface of CompositionObject. The goal is to have only one extra concept in WinCompData, i.e. classes look just like the real WinComp classes, except with one addition: metadata. IDescribable + metadata would have been 2 concepts.
And the IDescribable storage is now implemented by the metadata store, which should result in space savings at runtime as we now reserve only one pointer (for the metadata) rather than one for the LongDescription and another for the ShortDescription.

### DIFF
--- a/source/WinCompData/CompositionObject.cs
+++ b/source/WinCompData/CompositionObject.cs
@@ -15,7 +15,14 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData
 #endif
     abstract class CompositionObject : IDisposable, IDescribable
     {
+        // Keys used to identify the short description and long description metadata.
+        static readonly Guid s_shortDescriptionMetadataKey = new Guid("AF01D303-5572-4540-A4AC-E48F1394E1D1");
+        static readonly Guid s_longDescriptionMetadataKey = new Guid("63514254-2B3E-4794-B01D-9F67D5946A7E");
+
         readonly ListOfNeverNull<Animator> _animators = new ListOfNeverNull<Animator>();
+
+        // Null until the first metatadata is set.
+        SortedDictionary<Guid, object> _metadata;
 
         private protected CompositionObject()
         {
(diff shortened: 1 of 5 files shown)


source/WinCompData/CompositionObject.cs
```diff
@@ -30,21 +37,63 @@ private protected CompositionObject()
...
+        /// <summary>
+        /// Associates metadata with the object. The type of metadata is indicated by the <paramref name="key"/>
+        /// <see cref="Guid"/>. If metadata for the given <paramref name="key"/> is already associated with the
+        /// object, the meatadata will be replaced with the new <paramref name="value"/>.
+        /// </summary>
+        /// <param name="key">A <see cref="Guid"/> that identifies that type of metadata.</param>
+        /// <param name="value">The value of the metadata.</param>
+        public void SetMetadata(in Guid key, object value)
+        {
+            if (_metadata == null)
```
simeoncran: >== [](start = 26, length = 2)

is instead of ==

source/WinCompData/CompositionObject.cs
```diff
@@ -30,21 +37,63 @@ private protected CompositionObject()
...
+        /// <summary>
+        /// Returns the metadata associated with this object that is identified by the given
+        /// <paramref name="key"/>, or null if no such metadata has been associated with this
+        /// object yet.
+        /// </summary>
+        /// <param name="key">A <see cref="Guid"/> that identifies that type of metadata.</param>
+        /// <returns>The metadata, or null.</returns>
+        public object TryGetMetadata(in Guid key)
+        {
+            if (_metadata != null && _metadata.TryGetValue(key, out var result))
```
simeoncran: >!= [](start = 26, length = 2)

is object instead of !=

PR #195, opened 2019-12-03
Getting Lottie metadata piped through.

simeoncran:
Adding Lottie metadata to the TranslationResult in GenericData format.
Some improvements of GenericData as a result of using it more.
Adding code to convert from GenericData to Windows.Data.Json. Not yet used but will be added to LottieVisualSource.
Replacing all of our custom empty arrays with Array.Empty<T>().


source/GenericData/GenericDataBool.cs
```diff
@@ -23,6 +23,8 @@ abstract class GenericDataBool : GenericDataObject
 
         public override GenericDataObjectType Type => GenericDataObjectType.Bool;
 
+        public static implicit operator GenericDataBool(bool value) => Create(value);
```
scott-moore-ms: With this existing, would/should anyone ever call the public Create method or should that become private?

source/GenericData/GenericDataList.cs
```diff
@@ -45,5 +46,7 @@ public override string ToString()
             => Count == 0
                 ? "[]"
                 : $"[{string.Join(", ", _items.Select(x => ToString(x)))}]";
+
+        public static implicit operator GenericDataList(GenericDataObject[] value) => Create(value);
```
scott-moore-ms: >GenericDataObject [](start = 56, length = 17)

Why specifically a GenericDataObject[] vs IEnumerable<GenericDataObject> as the Create() method takes?

source/GenericData/GenericDataObject.cs
```diff
@@ -11,6 +13,14 @@ abstract class GenericDataObject
     {
         public abstract GenericDataObjectType Type { get; }
 
+        public static implicit operator GenericDataObject(bool value) => GenericDataBool.Create(value);
```
scott-moore-ms: >GenericDataObject [](start = 40, length = 17)

Oh interesting - why not => (GenericDataBool)value ?

source/LottieToWinComp/TranslationResult.cs
```diff
@@ -19,18 +20,25 @@ sealed class TranslationResult
...
             Visual rootVisual,
             IEnumerable<TranslationIssue> translationIssues,
-            uint minimumRequiredUapVersion)
+            uint minimumRequiredUapVersion,
+            GenericDataMap sourceMetadata)
         {
             RootVisual = rootVisual;
             TranslationIssues = translationIssues.ToArray();
             MinimumRequiredUapVersion = minimumRequiredUapVersion;
+            SourceMetadata = sourceMetadata;
```
scott-moore-ms: >SourceMetadata [](start = 12, length = 14)

Should we assert that this is non-null?

source/GenericData/GenericDataBool.cs
simeoncran: The implicit conversions are handy but I want to allow callers to be explicit if it makes things clearer at the callsite.

---
In reply to: [353405137](https://github.com/windows-toolkit/Lottie-Windows/pull/195#discussion_r353405137) [](ancestors = 353405137)

source/GenericData/GenericDataList.cs
simeoncran: Implicit operators don't accept interfaces as arguments.

---
In reply to: [353405786](https://github.com/windows-toolkit/Lottie-Windows/pull/195#discussion_r353405786) [](ancestors = 353405786)

source/GenericData/GenericDataObject.cs
simeoncran: Given the choice between an explicit cast that converts into a call to Create, and just calling the method, they both take up the same space but the latter is clearer to me.


---
In reply to: [353407114](https://github.com/windows-toolkit/Lottie-Windows/pull/195#discussion_r353407114) [](ancestors = 353407114)

source/LottieToWinComp/TranslationResult.cs
simeoncran: Good philosophy question. :-) Once we can switch to C# 8 we can use the type system to make it clear whether this value is optional or not. Let's discuss in person.

---
In reply to: [353412670](https://github.com/windows-toolkit/Lottie-Windows/pull/195#discussion_r353412670) [](ancestors = 353412670)

PR #198, opened 2019-12-05
Adding a property bindings language parser.

simeoncran:
The parser is designed to allow designers to embed information about how colors should be hooked up to theme colors in the app or OS on which the Lottie runs.
Currently the parser is hooked up to parse property bindings from solid color fills, but it doesn't do anything with the result - this is just to get some coverage on it.
Eventually the property bindings will results in generated code that puts the given color in a CompositionPropertySet property and binds it appropriately so that clients can change the color at runtime.

### DIFF
--- a/source/LottieToWinComp/PropertyBindingsParser.cs
+++ b/source/LottieToWinComp/PropertyBindingsParser.cs
@@ -1,0 +1,83 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
+{
+    static class PropertyBindingsParser
+    {
+        // Property binding language definition
+        // ====================================
+        // Property bindings are lists of property-name+binding-name pairs.
+        // They are specified using using a regular language (i.e. parseable
+        // by a regular expression) that is designed to look similar to CSS var
+        // functions. Just like CSS, the property bindings are enclosed within
+        // curly braces and multiple property bindings are separated by semicolons.
+        //
+        // <PropertyBindings>    ::= "{" <OptWS> <PropertyBindingList> <OptWS> "}"
+        // <PropertyBindingList> ::= <PropertyBinding> | <PropertyBinding> <OptWS> ";" <OptWS> <PropertyBindingList>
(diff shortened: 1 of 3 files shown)


source/LottieToWinComp/PropertyBindingsParser.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
+{
+    static class PropertyBindingsParser
+    {
+        // Property binding language definition
+        // ====================================
+        // Property bindings are lists of property-name+binding-name pairs.
+        // They are specified using using a regular language (i.e. parseable
```
simeoncran: >using [](start = 36, length = 5)

typo

source/LottieToWinComp/PropertyBindingsParser.cs
```diff
@@ -0,0 +1,83 @@
...
+        // <PropertyBindingList> ::= <PropertyBinding> | <PropertyBinding> <OptWS> ";" <OptWS> <PropertyBindingList>
+        // <PropertyBinding>     ::= <PropertyName> <OptWS> ":" <OptWS> "var(" <OptWS> <BindingName> <OptWS> ")"
+        // <PropertyName>        ::= <Identifier>
+        // <BindingName>         ::= <Identifier>
+        // <OptWS>               ::= <Whitespace> <OptWS> | ""
+        // <Identifier>          ::= a word starting with an alpha character.
+        // ------------------------------------
+        // Example: "{ color :var( Foreground);color1:var(Foreground1) ; color3:var(L33t )   }"
+        // ------------------------------------
+        const string PropertyNameSelector = "PropertyName";
```
simeoncran: >PropertyName [](start = 45, length = 12)

These don't need to be human readable. Might as well make them single letters and save some space.

source/LottieToWinComp/PropertyBindingsParser.cs
```diff
@@ -0,0 +1,83 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
+{
+    static class PropertyBindingsParser
+    {
+        // Property binding language definition
```
scott-moore-ms: >Property binding language definition [](start = 11, length = 36)

Is this something that we're defining, or a format that someone else is producing and we're consuming?

simeoncran: We are defining it. But it's inspired by CSS because our brief survey suggested that motion designers usually know some CSS.

---
In reply to: [354647432](https://github.com/windows-toolkit/Lottie-Windows/pull/198#discussion_r354647432) [](ancestors = 354647432)

PR #200, opened 2019-12-07
More support for property bindings.

simeoncran:
Add support for Color values in CompositionPropertySet.
Add support for TryGet* methods in CompositionPropertySet.
Refactor the CompositionPropertySet implementation so that it's easier to add new types.
Adds a switch to the translator to determine whether or not it recognizes property bindings. Property bindings are only useful for codegen so this switch is off for LottieVisualSource and on for LottieGen.
Translator will now insert a property for a solid color fill that has a property binding for "Color".

### DIFF
--- a/source/WinCompData/CompositionPropertySet.cs
+++ b/source/WinCompData/CompositionPropertySet.cs
@@ -2,20 +2,26 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
+using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Wui;
 
 namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData
 {
+    // CompositionPropertySet was introduced in version 2. Boolean types
+    // were added in version 3.
     [MetaData.UapVersion(2)]
 #if PUBLIC_WinCompData
     public
 #endif
     sealed class CompositionPropertySet : CompositionObject
     {
-        readonly Dictionary<string, float> _scalarProperties = new Dictionary<string, float>();
-        readonly Dictionary<string, Vector2> _vector2Properties = new Dictionary<string, Vector2>();
+        HashSet<string> _names;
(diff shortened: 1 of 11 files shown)


source/WinCompData/CompositionPropertySet.cs
```diff
@@ -24,19 +30,105 @@ internal CompositionPropertySet(CompositionObject owner)
...
+            if (_names is null)
+            {
+                value = default(T);
+                return CompositionGetValueStatus.NotFound;
+            }
+
+            // The name is in the property set - does it refer to the right type?
+            return bag.TryGetValue(propertyName, out value)
+                ? CompositionGetValueStatus.Succeeded
+                : CompositionGetValueStatus.TypeMismatch;
```
eliezerpMS: >CompositionGetValueStatus.TypeMismatch [](start = 18, length = 38)

Question: What does it mean to the caller if they get this enum value? It might be good to have some comments in the enum declaration.

From the names I would have assumed that CompositionGetValueStatus.NotFound is used when _names does not contain propertyName and I guess i am not not sure about CompositionGetValueStatus.TypeMismatch.

simeoncran: This has the same semantics as the real CompositionPropertySet object. If a value has been stored with a particular key name it is forever associated with the type of the value. So if someone inserted a Vector2 called Barry, then you asked for a Color called Barry, you'd get this error that tells you that Barry exists, it's just not the Barry that's you're looking for.

---
In reply to: [355613036](https://github.com/windows-toolkit/Lottie-Windows/pull/200#discussion_r355613036) [](ancestors = 355613036)

simeoncran: I've added comments to the CompositoinGetValueStatus enum, copied from [https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Composition.CompositionGetValueStatus](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Composition.CompositionGetValueStatus). Hopefully that will make the next reader's job easier.

---
In reply to: [355616436](https://github.com/windows-toolkit/Lottie-Windows/pull/200#discussion_r355616436) [](ancestors = 355616436,355613036)

PR #206, opened 2019-12-13
Add metadata comments to generated code: markers, bound properties, et al

simeoncran:
This adds comments to generated code that is useful for consumers, such as:
```
    // Name:        QA_Wifi_Tagged
    // Frame rate:  60 fps
    // Frame count: 60
    // ===========
    // Segments (aka markers):
    // "1"             0mS  player.SetProgress(0)
    // "2"             0mS  player.SetProgress(0.5)
    // "Tap_On"      450mS  player->PlayAsync(0.55, 0.775, _)
    // "Tap_Off"     400mS  player->PlayAsync(0.775, 0.975, _)
    // "DeadFrames"   33mS  player->PlayAsync(0.975, 0.9915, _)
    // "Opaque"        0mS  player->SetProgress(0.991666666666667)
    // ===========
    // Property bindings:
    // "Foreground"
```


source/Lottie/LottieVisualSource.cs
```diff
@@ -383,7 +381,7 @@ internal async Task<ContentFactory> LoadAsync(LottieVisualOptions options)
                     // and dividing by OutPoint - InPoint
                     diagnostics.Markers = lottieComposition.Markers.Select(m =>
                     {
-                        return new KeyValuePair<string, double>(m.Name, (m.Progress * lottieComposition.FramesPerSecond) / lottieComposition.Duration.TotalSeconds);
+                        return new KeyValuePair<string, double>(m.Name, (m.Frame * lottieComposition.FramesPerSecond) / lottieComposition.Duration.TotalSeconds);
```
eliezerpMS: >return new KeyValuePair<string, double>(m.Name, (m.Frame * lottieComposition.FramesPerSecond) / lottieComposition.Duration.TotalSeconds); [](start = 24, length = 137)

nit: maybe try wrapping this line

source/LottieData/Marker.cs
```diff
@@ -10,22 +10,24 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
...
+            DurationMilliseconds = durationMilliseconds;
         }
 
         /// <summary>
-        /// Gets the time value of the marker. This value must be multipled by the composition
-        /// duration to get the actual time offset.
+        /// Gets the frame for the start of the marker.
         /// </summary>
-        public double Progress { get; }
+        public double Frame { get; }
```
eliezerpMS: >Frame [](start = 22, length = 5)

Perhaps a silly question but Frame is a substitute for time?

simeoncran: It's the frame "number". So if the lottie is 60FPS, and you want to refer to the point at 1 second in, you'd use Frame==60. But you can refer to somewhere between frames as well, so it's a floating point value. I didn't want to call it FrameNumber because that seemed to imply an integer.

I had the names wrong for the properties of a maker previously - they didn't reflect what the markers properties actually meant. The names now reflect the true meaning. Maybe you can come up with even better names.

---
In reply to: [357871916](https://github.com/windows-toolkit/Lottie-Windows/pull/206#discussion_r357871916) [](ancestors = 357871916)

source/Lottie/LottieVisualSource.cs
simeoncran: And comments are wrong there too.

---
In reply to: [357871559](https://github.com/windows-toolkit/Lottie-Windows/pull/206#discussion_r357871559) [](ancestors = 357871559)

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -289,6 +293,55 @@ protected void WriteBytesField(CodeBuilder builder, string fieldName)
             builder.WriteLine($"{_stringifier.Static} {_stringifier.Readonly(_stringifier.ReferenceTypeName(_stringifier.ByteArray))} {fieldName} = {_stringifier.New} {_stringifier.ByteArray}");
         }
 
+        /// <summary>
+        /// Returns text that describes the contents of the source metadata.
```
eliezerpMS: >Returns text that describes the contents of the source metadata. [](start = 12, length = 64)

Nit: Maybe describe the "rules" used here for formatting the output in the event that a person in the future needs to add to it but wants to keep it consistent.

source/UIDataCodeGen/CodeGen/LottieMetadataDecoder.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        internal string CompositionName => _metadata.CompositionName;
+
+        internal double FramesPerSecond => _metadata.FramesPerSecond;
+
+        internal TimeSpan Duration => TimeSpan.FromSeconds(DurationInFrames / FramesPerSecond);
+
+        internal double DurationInFrames => _metadata.OutPoint - _metadata.InPoint;
+
+        internal IEnumerable<(string name, (TimeSpan time, double progress) start, (TimeSpan time, double progress) end)> Markers
```
eliezerpMS: >Markers [](start = 122, length = 7)

It this going to get called often? Does it make sense for the result to cached the first time and return the cached value on future requests?

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: Not sure how to address that. I just fiddled with it until it looked somewhat pretty. Next person to touch it should do the same. The comments aren't meant to be machine readable.

---
In reply to: [357874395](https://github.com/windows-toolkit/Lottie-Windows/pull/206#discussion_r357874395) [](ancestors = 357874395)

source/UIDataCodeGen/CodeGen/LottieMetadataDecoder.cs
simeoncran: I think that is a premature optimization. It will typically get called once or twice. First time to see if there are any items, second to return the items. Each time it will do a couple of cheap allocations to get the enumerable and the enumerator, and the there will rarely be more than about 8 markers, and usually there will be none.

---
In reply to: [357875829](https://github.com/windows-toolkit/Lottie-Windows/pull/206#discussion_r357875829) [](ancestors = 357875829)

PR #208, opened 2020-01-07
Vector4 didn't support simplification. Now it does.

simeoncran:
This includes some cleanup to make Vector expression code more consistent, and takes advantage of C# 7 switching on types.

### DIFF
--- a/source/WinCompData/Expressions/ScalarSubchannel.cs
+++ b/source/WinCompData/Expressions/ScalarSubchannel.cs
@@ -20,11 +20,66 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Expressions
         public string ChannelName { get; }
 
         /// <inheritdoc/>
-        protected override Expression Simplify() => this;
+        protected override Expression Simplify()
+        {
+            switch (Value)
+            {
+                case Vector4 vector:
+                    switch (ChannelName)
+                    {
+                        case "X":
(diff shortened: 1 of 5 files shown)


source/WinCompData/Expressions/Vector3.cs
```diff
@@ -9,23 +9,37 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Expressions
...
+
+        public new Expression Z { get; }
+
         /// <inheritdoc/>
-        protected override Expression Simplify() => this;
+        protected override Expression Simplify()
+        {
+            var x = X.Simplified;
+            var y = Y.Simplified;
+            var z = Z.Simplified;
```
simeoncran: For consistency: whitespace.

source/WinCompData/Expressions/ScalarSubchannel.cs
```diff
@@ -20,11 +20,66 @@ internal ScalarSubchannel(Expression value, string channelName)
         public string ChannelName { get; }
 
         /// <inheritdoc/>
-        protected override Expression Simplify() => this;
+        protected override Expression Simplify()
+        {
+            switch (Value)
+            {
+                case Vector4 vector:
+                    switch (ChannelName)
```
simeoncran: lower-case the channel name to reduce the number of cases.

PR #211, opened 2020-01-09
Improve NodeNamer readability.

simeoncran:
This is a follow-up from a previous PR where there were some questions about the NodeNamer, so I've refactored it to take advantage of C# switch expressions and added better comments.
Also move LottieGen to .NET Core 3.1 (was 2.2). This means we can use newer features, and we get an actual exe rather than having to use the "dotnet" bootstrapper.

### DIFF
--- a/LottieGen/README.md
+++ b/LottieGen/README.md
@@ -25,7 +25,7 @@ Local builds can be installed from your bin\nupkg directory:
 
 Local builds can be run directly:
 
-    dotnet f:\GitHub\Lottie-Windows\LottieGen\bin\AnyCpu\Debug\netcoreapp3.1\lottiegen.exe
+    f:\GitHub\Lottie-Windows\LottieGen\bin\AnyCpu\Debug\netcoreapp3.1\lottiegen.exe
 
 ## Updating
     dotnet tool update -g LottieGen
--- a/LottieGen/README.md
+++ b/LottieGen/README.md
@@ -2,7 +2,7 @@
 
 LottieGen is a tool for generating C#, C++, and other outputs from Lottie / Bodymovin JSON files. LottieGen is built as a [.NET Core global tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools), which means it will run anywhere .NET Core is available, including Linux and Mac.
 
-LottieGen requires [.NET Core 2.2 SDK](https://dotnet.microsoft.com/download/dotnet-core/2.2) or later.
+LottieGen requires [.NET Core 3.1 SDK](https://dotnet.microsoft.com/download/dotnet-core/3.1) or later.
 
 ## Installing
 *The following commands are examples only; adjust the paths and versions as necessary.*
@@ -13,19 +13,19 @@ The latest release version can be [installed from NuGet](https://www.nuget.org/p
 
 A specific version can be installed from NuGet:
 
-    dotnet tool install -g LottieGen --version 6.0.0
+    dotnet tool install -g LottieGen --version 6.1.0
 
 CI builds can be [installed from MyGet](https://dotnet.myget.org/feed/uwpcommunitytoolkit/package/nuget/LottieGen):
 
-    dotnet tool install -g LottieGen --add-source https://dotnet.myget.org/F/uwpcommunitytoolkit/api/v3/index.json --version 6.1.0-build.3
+    dotnet tool install -g LottieGen --add-source https://dotnet.myget.org/F/uwpcommunitytoolkit/api/v3/index.json --version 6.1.0-build.19
 
 Local builds can be installed from your bin\nupkg directory:
 
-    dotnet tool install -g LottieGen --add-source f:\GitHub\Lottie-Windows\bin\nupkg --version 6.1.0-build.11.g31523b44e4
+    dotnet tool install -g LottieGen --add-source f:\GitHub\Lottie-Windows\bin\nupkg --version 6.1.0-build.18.g31523b44e4
 
 Local builds can be run directly:
 
-    dotnet f:\GitHub\Lottie-Windows\LottieGen\bin\AnyCpu\Debug\netcoreapp2.2\lottiegen.dll
+    dotnet f:\GitHub\Lottie-Windows\LottieGen\bin\AnyCpu\Debug\netcoreapp3.1\lottiegen.exe
 
 ## Updating
     dotnet tool update -g LottieGen
--- a/source/UIDataCodeGen/CodeGen/NodeNamer.cs
+++ b/source/UIDataCodeGen/CodeGen/NodeNamer.cs
@@ -189,26 +189,27 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen
             var expression = obj.Expression;
             var expressionType = expression.InferredType;
             return expressionType.IsValid && !expressionType.IsGeneric
-                ? $"{expressionType.Constraints.ToString()}ExpressionAnimation"
+                ? $"{expressionType.Constraints}ExpressionAnimation"
                 : "ExpressionAnimation";
         }
 
         static string DescribeStepEasingFunction(StepEasingFunction obj)
         {
             // Recognize 2 common patterns: HoldThenStep and StepThenHold
-            if (obj.StepCount == 1 && obj.IsFinalStepSingleFrame && !obj.IsInitialStepSingleFrame)
+            if (obj.StepCount == 1)
             {
-                return "HoldThenStepEasingFunction";
+                if (obj.IsFinalStepSingleFrame && !obj.IsInitialStepSingleFrame)
+                {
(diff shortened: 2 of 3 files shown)


LottieGen/README.md
```diff
@@ -13,19 +13,19 @@ The latest release version can be [installed from NuGet](https://www.nuget.org/p
...
 
 Local builds can be installed from your bin\nupkg directory:
 
-    dotnet tool install -g LottieGen --add-source f:\GitHub\Lottie-Windows\bin\nupkg --version 6.1.0-build.11.g31523b44e4
+    dotnet tool install -g LottieGen --add-source f:\GitHub\Lottie-Windows\bin\nupkg --version 6.1.0-build.18.g31523b44e4
 
 Local builds can be run directly:
 
-    dotnet f:\GitHub\Lottie-Windows\LottieGen\bin\AnyCpu\Debug\netcoreapp2.2\lottiegen.dll
+    dotnet f:\GitHub\Lottie-Windows\LottieGen\bin\AnyCpu\Debug\netcoreapp3.1\lottiegen.exe
```
derobinsMSFT: >exe [](start = 87, length = 3)

Why is this extension changed?

simeoncran: Because .NET Core 3.1 generates a LottieGen.exe for dotnet tools. Previously you'd have to use "dotnet Lottiegen.dll" or you'd install the tool, but now a LottieGen.exe is generated so we might as well point to that.
But now that you've drawn my attention to this - it shouldn't have "dotnet" at the front of this command. I will fix.

---
In reply to: [364923285](https://github.com/windows-toolkit/Lottie-Windows/pull/211#discussion_r364923285) [](ancestors = 364923285)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -91,6 +91,156 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
+                _ => obj.ToString(),
+            };
+
+            // Remove the "Composition" prefix so the name is easier to read.
+            // The prefix is redundant as far as the reader is concerned because most of the
+            // objects have it and it doesn't indicate anything useful to the reader.
+            return StripPrefix(result, "Composition");
+        }
+
+        static string DescribeCompositionColorBrush(TNode node, CompositionColorBrush obj)
```
derobinsMSFT: >TNode node [](start = 52, length = 10)

Why is this parameter included? It doesn't appear to be used. Same for the other new `Describe*` helpers.

derobinsMSFT:
>         static string DescribeAnimationRange(ColorKeyFrameAnimation animation) => DescribeAnimationRange(animation, v => v.Name);

Move these so that they're next to the other `Describe*` methods.

---
Refers to: source/UIDataCodeGen/CodeGen/NodeNamer.cs:265 in 3e156fc. [](commit_id = 3e156fc529ce0f8494349518b835d6589472353b, deletion_comment = False)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -91,6 +91,156 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
+        static string DescribeExpressionAnimation(TNode node, ExpressionAnimation obj)
+        {
+            var expression = obj.Expression;
+            var expressionType = expression.InferredType;
+            return expressionType.IsValid && !expressionType.IsGeneric
+                ? $"{expressionType.Constraints.ToString()}ExpressionAnimation"
+                : "ExpressionAnimation";
+        }
+
+        static string DescribeStepEasingFunction(TNode node, StepEasingFunction obj)
```
derobinsMSFT: >DescribeStepEasingFunction [](start = 22, length = 26)

Nit: order the `Describe*` methods alphabetically?

source/UIDataCodeGen/CodeGen/NodeNamer.cs
simeoncran: For consistency - so all the Describe* have a similar looking API. I'll see if anyone them apart from DescribeCompositionObject use the Node and if not I'll remove the parameter.

---
In reply to: [364985227](https://github.com/windows-toolkit/Lottie-Windows/pull/211#discussion_r364985227) [](ancestors = 364985227)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
simeoncran: They are, except for the top-level DescribeCompositionObject, which I like having first because it's the thing that the GenerateNodeNames calls.

---
In reply to: [364989193](https://github.com/windows-toolkit/Lottie-Windows/pull/211#discussion_r364989193) [](ancestors = 364989193)

simeoncran:
>         static string DescribeAnimationRange(ColorKeyFrameAnimation animation) => DescribeAnimationRange(animation, v => v.Name);

I'll move the DescribeAnimationRange methods up a bit, but I don't want to have them in with the describe methods that describe objects, but they do something different.

---
In reply to: [572783651](https://github.com/windows-toolkit/Lottie-Windows/pull/211#issuecomment-572783651) [](ancestors = 572783651)

---
Refers to: source/UIDataCodeGen/CodeGen/NodeNamer.cs:265 in 3e156fc. [](commit_id = 3e156fc529ce0f8494349518b835d6589472353b, deletion_comment = False)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
derobinsMSFT: `DescribeStepEasingFunction` appears before `DescribeLoadedImageSurface` as well.

---
In reply to: [364990957](https://github.com/windows-toolkit/Lottie-Windows/pull/211#discussion_r364990957) [](ancestors = 364990957,364989193)

simeoncran: That was intentional... LoadedImageSurface is a XAML type, the ones before it are all the CompositionObject types.

---
In reply to: [364991918](https://github.com/windows-toolkit/Lottie-Windows/pull/211#discussion_r364991918) [](ancestors = 364991918,364990957,364989193)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -91,205 +91,136 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
-            if (result.StartsWith(compositionPrefix))
+            else if (obj.StepCount == 1 && obj.IsInitialStepSingleFrame && !obj.IsFinalStepSingleFrame)
             {
-                result = result.Substring(compositionPrefix.Length);
+                return "StepThenHoldEasingFunction";
+            }
+            else
+            {
+                // Didn't recognize the pattern.
+                return obj.ToString();
```
simeoncran: This needs to be "EasingFunction"

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -91,205 +91,137 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
+                // Brush is not animated. Give it a name based on the color.
+                return AppendDescription("ColorBrush", obj.Color?.Name);
+            }
+        }
 
-        // Returns a string for use in an identifier that describes a KeyFrameAnimation, or null
-        // if the animation cannot be described.
-        static string DescribeAnimationRange<T>(KeyFrameAnimation<T> animation, Func<T, string> valueFormatter)
-            where T : struct
+        static string DescribeCompositionColorGradientStop(CompositionColorGradientStop obj)
```
janny10: almost seems a shame this and the above method aren't consolidated... but ultimately I agree with this

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -91,205 +91,137 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
+                    return "StepThenHoldEasingFunction";
+                }
             }
 
-            return result;
+            // Didn't recognize the pattern.
+            return "EasingFunction";
         }
 
         static string DescribeLoadedImageSurface(TNode node, LoadedImageSurface obj)
```
janny10: >DescribeLoadedIm [](start = 22, length = 16)

I like this conditional way better

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -91,205 +91,137 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
         {
             string result = null;
-            var loadedImageSurface = (LoadedImageSurface)node.Object;
```
janny10: good catch removing this

PR #212, opened 2020-01-09
Make CompositionPath names start at 0.

simeoncran:
CompositionPath factories get inlined during codegen unless they're shared. Because we were naming the factory nodes before determining whether the CompostionPath factories could be inlined, we ended up generating names for CompositionPaths that would never get a name. The result was that there would be gaps in the number of CompostionPath factories (e.g. CompostionPath_0, CompostionPath_4). With this change the CompostionPath numbers will be contiguous because we do the inlining before the naming.

### DIFF
--- a/source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
+++ b/source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
@@ -777,8 +777,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen
                                         n => n.Type == Graph.NodeType.CompositionPath &&
                                         IsEqualToOne(FilteredInRefs(n))))
                 {
-                    var pathSourceFactoryCall = CallFactoryFromFor(node, ((CompositionPath)node.Object).Source);
-                    node.ForceInline($"{New} CompositionPath({_stringifier.FactoryCall(pathSourceFactoryCall)})");
+                    node.ForceInline(() =>
+                    {
+                        var inlinedFactoryCode = CallFactoryFromFor(node, ((CompositionPath)node.Object).Source);
+                        return $"{New} CompositionPath({_stringifier.FactoryCall(inlinedFactoryCode)})";
+                    });
                 }
 
                 // Get the nodes that will produce factory methods.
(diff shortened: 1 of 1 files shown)


source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -450,9 +488,9 @@ static IEnumerable<ObjectData> FilteredInRefs(ObjectData node)
                     }
 
                     // ... is the animation animating a property on the current node or its property set.
```
derobinsMSFT: >. [](start = 104, length = 1)

Question mark.

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -734,6 +772,18 @@ sealed class AnimatedVisualGenerator : IAnimatedVisualInfo
                 // Build the object graph.
                 _objectGraph = ObjectGraph<ObjectData>.FromCompositionObject(graphRoot, includeVertices: true);
 
+                // Force inlining on CompositionPath nodes that are only referenced once, because they are always very simple.
```
derobinsMSFT: >they are always very simple [](start = 98, length = 27)

Not entirely sure what this means, or what makes a node referenced just once inherently "simple".

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -430,6 +430,44 @@ static void WriteField(CodeBuilder builder, string typeName, string fieldName)
...
+                count++;
+            }
+
+            return count == 1;
+        }
+
+        // Returns true iff the given sequence has more than one item in it.
+        // This is equivalent to Count() > 1, but doesn't require the whole
+        // sequence to be enumerated.
+        static bool IsGreaterThanOne<T>(IEnumerable<T> items)
```
janny10: >IsGreaterThanOne [](start = 20, length = 16)

Is this necessary? Or just desired? Do node filtered InRefs lists get quite large?

janny10: >IsGreaterThanOne [](start = 20, length = 16)

I am personally on board for these, but are these strictly necessary or just desired? Do the node InRef lists get quite large?

simeoncran: Things that are shared a lot can have a large number. For example, a White brush might get used all over the place. All we care about here is whether it is shared by multiple things or not, so it's wasteful to walk the whole list.
But now that I look at the implementation again, I can make it a bit clearer by using a bool

---
In reply to: [365379483](https://github.com/windows-toolkit/Lottie-Windows/pull/212#discussion_r365379483) [](ancestors = 365379483)

PR #213, opened 2020-01-13
Make Expression objects have stronger static types.

simeoncran:
This change allows more static checking for animation expressions.
The Expressions objects previously supported generic types and type inferencing because when this was first written it seemed like they would be used for a more general expression parser, but it meant that the type of an expression wasn't statically known. With this change, every expression has a static type, which allows stronger typing wherever an expression is used, and allows the C# type system to find some bugs in our use of expressions.

I found 2 bugs as a result of this - one that affected 2 Lottie files in our corpus (it produced code that would fail at runtime), and one that affected a codepath we never hit (it made it obvious that this codepath is never hit).

I've run this over all of our corpus and the results are exactly the same as before, apart from the 2 files that were hitting the bug mentioned above. So there are definitely no regressions for all known Lottie files.


raymond-wh-leung:
> #if ReuseExpressionAnimation

Curious why it needs this flag and in what cases the different path would be used.

---
Refers to: source/Lottie/Instantiator.cs:557 in d61268e. [](commit_id = d61268eb19a099fc8e92e4fc9845b863a206db8d, deletion_comment = False)

source/LottieToWinComp/ExpressionFactory.cs
```diff
@@ -5,98 +5,91 @@
...
     {
-        // The name used to bind to the property set that contains the Progress property.
-        const string RootName = "_";
-        static readonly Expression s_myTStart = Scalar("my.TStart");
-        static readonly Expression s_myTEnd = Scalar("my.TEnd");
+        internal static readonly Vector2 MyAnchor2 = MyVector2("Anchor");
+        internal static readonly Vector4 MyColor = MyVector4("Color");
+        internal static readonly Scalar MyInheritedOpacity = MyScalar("InheritedOpacity");
+        internal static readonly Scalar MyOpacity = MyScalar("Opacity");
+        internal static readonly Vector2 MyPosition2 = MyVector2("Position");
```
raymond-wh-leung: >2  [](start = 51, length = 2)

wondering why some properties are suffixed with the number (that seem to match the type) vector2 but some are not?

source/WinCompData/Expressions/Color.cs
```diff
@@ -0,0 +1,74 @@
...
+            public Asserted(string text)
+            {
+                _text = text;
+            }
+
+            /// <inheritdoc/>
+            protected override string CreateExpressionText() => _text;
+
+            protected override bool IsAtomic => true;
+        }
```
raymond-wh-leung: Trying to see what this class is used for..

source/LottieToWinComp/ExpressionFactory.cs
simeoncran: You're right... they're probably not necessary now. They mattered when the expressions didn't have statically known types.  There may also be cases where there were Vector3 (for Visual) and Vector2 (for Shape) versions, in which case the number matters.
I'll add this to my todo list and improve it in a separate PR. I'd fix it here, but there is another change coming that depends on this and it will be painful if these names change before that's in.

---
In reply to: [365941677](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r365941677) [](ancestors = 365941677)

source/WinCompData/Expressions/ExpressionType.cs
```diff
@@ -3,151 +3,30 @@
...
-using System.Collections.Generic;
 
 namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Expressions
 {
     /// <summary>
     /// Describes the type of an <see cref="Expression"/> or variable.
     /// </summary>
+    /// <summary>
+    /// Types to which an expression or variable is constrained.
+    /// </summary>
```
raymond-wh-leung: nit: there are two summary blocks for this.

source/LottieToWinComp/ExpressionFactory.cs
raymond-wh-leung: Gotcha

---
In reply to: [365951989](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r365951989) [](ancestors = 365951989,365941677)

source/WinCompData/Expressions/Color.cs
simeoncran: All of the .Asserted classes are used to give types to expressions that are expressed as text. For example: we use "_.Progress" to indicate "the property called Progress on the object that is bound to '_'". Have a look at ExpressionFactory.RootProgress to see an example. Expressions expressed as text need to have their types asserted because we have no way to figure out what the user intended.  The Color.Asserted class is called from Expression.cs. Today there are no upstream callers, but the ability to assert that some text evaluates to a color is included because I wrote it before I realized that I don't currently use it, so I'm leaving it in for future use.

---
In reply to: [365949845](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r365949845) [](ancestors = 365949845)

source/Lottie/Instantiator.cs
```diff
@@ -567,7 +568,7 @@ Wc.ExpressionAnimation GetExpressionAnimation(Wd.ExpressionAnimation obj)
                 result.Target = obj.Target;
             }
 
-            result.Expression = obj.Expression.Simplified.ToString();
+            result.Expression = obj.Expression.ToString();
```
simeoncran: Oops... this is should be "ToText()"

simeoncran:
> #if ReuseExpressionAnimation

Current this flag is always set. It means that our generated code creates a single ExpressionAnimation, stores it in a field, and clears it and re-parameterizes it for each expression animation that is started. The alternative is to create a new ExpressionAnimation for each use. I figure that it's better to reuse a single one, but the option is here to not reuse if it turns out to be better.

---
In reply to: [573779897](https://github.com/windows-toolkit/Lottie-Windows/pull/213#issuecomment-573779897) [](ancestors = 573779897)

---
Refers to: source/Lottie/Instantiator.cs:557 in d61268e. [](commit_id = d61268eb19a099fc8e92e4fc9845b863a206db8d, deletion_comment = False)

source/WinCompData/Expressions/ExpressionType.cs
simeoncran: Thanks!.. looks like a bad squash rebase.

---
In reply to: [365952758](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r365952758) [](ancestors = 365952758)

source/WinCompData/Expressions/Vector2.cs
```diff
@@ -2,48 +2,290 @@
...
+
+                return left != Left || right != Right
+                    ? new Multiply(left, right)
+                    : this;
+            }
+
+            /// <inheritdoc/>
+            protected override string CreateExpressionText()
+            {
+                var left = Left is Multiply ? Left.ToText() : Parenthesize(Left);
```
raymond-wh-leung: >Left is Multiply [](start = 27, length = 16)

wondering if it needs to check for Divide too.

source/WinCompData/Expressions/Matrix3x2.cs
```diff
@@ -2,36 +2,114 @@
...
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Expressions
 {
 #if PUBLIC_WinCompData
     public
 #endif
-    sealed class Matrix3x2 : Expression
+    abstract class Matrix3x2 : Expression_<Matrix3x2>
```
raymond-wh-leung: >Matrix3x2 [](start = 19, length = 9)

There are changes to the files for Color/Vector2,3,4, and Matrix3x2. Can't seem to find the changes for Matrix4x4 and Quaternion, not sure if that is expected?

simeoncran: They're gone. We're not currently using them. They'll get recreated if/when they're used.

---
In reply to: [365965739](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r365965739) [](ancestors = 365965739)

raymond-wh-leung:
> #if ReuseExpressionAnimation

Ok I see, thanks.

---
In reply to: [573802083](https://github.com/windows-toolkit/Lottie-Windows/pull/213#issuecomment-573802083) [](ancestors = 573802083,573779897)

---
Refers to: source/Lottie/Instantiator.cs:557 in d61268e. [](commit_id = d61268eb19a099fc8e92e4fc9845b863a206db8d, deletion_comment = False)

raymond-wh-leung:
naïve signoff

source/Lottie/Instantiator.cs
```diff
@@ -602,12 +603,12 @@ Wc.ColorKeyFrameAnimation GetColorKeyFrameAnimation(Wd.ColorKeyFrameAnimation ob
...
-                        var expressionKeyFrame = (Wd.KeyFrameAnimation<Wd.Wui.Color>.ExpressionKeyFrame)kf;
-                        result.InsertExpressionKeyFrame(kf.Progress, expressionKeyFrame.Expression, GetCompositionEasingFunction(kf.Easing));
+                    case Wd.KeyFrameType.Expression:
+                        var expressionKeyFrame = (Wd.KeyFrameAnimation<Wd.Wui.Color, Expr.Color>.ExpressionKeyFrame)kf;
+                        result.InsertExpressionKeyFrame(kf.Progress, expressionKeyFrame.Expression.ToText(), GetCompositionEasingFunction(kf.Easing));
                         break;
-                    case Wd.KeyFrameAnimation<Wd.Wui.Color>.KeyFrameType.Value:
-                        var valueKeyFrame = (Wd.KeyFrameAnimation<Wd.Wui.Color>.ValueKeyFrame)kf;
+                    case Wd.KeyFrameType.Value:
+                        var valueKeyFrame = (Wd.KeyFrameAnimation<Wd.Wui.Color, Expr.Color>.ValueKeyFrame)kf;
```
janny10: Hmm.. seems a little odd that Color (and other types) is in the Expressions namespace when they are also used outside of Expressions.. would it be better to move them into like a CompositionDataTypes namespace or something like that? To be used by Expressions and normal animations..?

janny10: Oh nevermind, I misunderstood. The 2nd parameter to KeyFrameAnimation describes Expression Keyframe types. Disregard!

---
In reply to: [366550855](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366550855) [](ancestors = 366550855)

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+
+                if (left.IsZero)
+                {
+                    return left;
+                }
+
+                if (right.IsOne)
+                {
+                    return left;
+                }
```
janny10: Do we maybe want a case for Left == Right or left == right? Or is that captured by one of the cases below and I'm just missing it?

source/WinCompData/Expressions/Vector3.cs
```diff
@@ -7,44 +7,78 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Expressions
...
     public
 #endif
-    sealed class Vector3 : Expression
+    abstract class Vector3 : Expression_<Vector3>
     {
-        internal Vector3(Expression x, Expression y, Expression z)
+        protected Vector3()
         {
-            X = x;
-            Y = y;
```
janny10: >; [](start = 17, length = 1)

Do expressions not support dividing vector3s or 4s?

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+
+                if (left.IsOne)
+                {
+                    return right;
+                }
+
+                if (right.IsOne)
+                {
+                    return left;
+                }
```
janny10: Is it worth including a case to return Pow if left == right?

source/WinCompData/Expressions/LessThan.cs
```diff
@@ -1,47 +0,0 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
janny10: I understand these binary operations have been spread out across the types they apply to. I'm wondering why there isn't/never was a GreaterThan.cs file? Or an EqualTo?

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+            /// <inheritdoc/>
+            protected override string CreateExpressionText()
+            {
+                var left = Left is Sum ? Left.ToText() : Parenthesize(Left);
+                var right = Right is Sum ? Right.ToText() : Parenthesize(Right);
+                return $"{left} + {right}";
+            }
+        }
+
+        internal sealed class Subtract : BinaryExpression
```
janny10: >Subtract [](start = 30, length = 8)

Similar to other cases: worth having a left==right return 0?

janny10: >Subtract [](start = 30, length = 8)

Also nit: Rename classes to Subtract and Add or Sum and Difference?

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+                    ? new Multiply(left, right)
+                    : this;
+            }
+
+            /// <inheritdoc/>
+            protected override string CreateExpressionText()
+            {
+                var left = Left is Multiply ? Left.ToText() : Parenthesize(Left);
+                var right = Right is Multiply ? Right.ToText() : Parenthesize(Right);
+                return $"{left} * {right}";
```
janny10: Is there a reason this isn't just the same as Divide's CreateExpressionText. Multiply isn't atomic and doesn't have its own Parenthesize implementation so ToText() would be called implicitly by Parenthesize right?

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+                    return new Literal(literalLeft.Value + literalRight.Value);
+                }
+
+                return left != Left || right != Right
+                    ? new Sum(left, right)
+                    : this;
+            }
+
+            /// <inheritdoc/>
+            protected override string CreateExpressionText()
```
janny10: same comment here as in Multiply.CreateExpressionText

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+
+            /// <inheritdoc/>
+            protected override Scalar Simplify()
+            {
+                var left = Left.Simplified;
+                var right = Right.Simplified;
+
+                var literalLeft = left as Literal;
+                var literalRight = right as Literal;
+                if (literalLeft != null && literalRight != null)
```
janny10: Believe this could conditional could be written as its analagous one is in Max

if (left is Literal literalLeft && right is Literal literalRight)

source/WinCompData/Expressions/Color.cs
```diff
@@ -0,0 +1,74 @@
...
+                var b = B.Simplified;
+                var a = A.Simplified;
+
+                return r != R || g != G || b != B || a != A
+                            ? new Constructed(r, g, b, a)
+                            : this;
+            }
+
+            /// <inheritdoc/>
+            protected override string CreateExpressionText() => $"ColorRGB({A},{R},{G},{B})";
```
janny10: Is there a reason this text doesn't Simplify A, R, G, or B?

source/WinCompData/Expressions/Scalar.cs
simeoncran: Common cases are captured by "is Literal" cases below. If they're not literals we don't have a general way to compare them. We could compare for identity, but it would be somewhat weird for someone to divide an expression by itself.
My feeling is that we should wait until we find specific use cases and add them then.

---
In reply to: [366599124](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366599124) [](ancestors = 366599124)

source/WinCompData/Expressions/Vector3.cs
simeoncran: Whatever Composition supports, we can support here - this is supposed to mimic the semantics of Comp. However I'm avoiding making it complete because if we're not using it it will go untested and will be an unnecessary maintenance burden.

---
In reply to: [366610667](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366610667) [](ancestors = 366610667)

source/WinCompData/Expressions/Scalar.cs
simeoncran: Sure... that's a good one to put in (actually it will be Square(left)) because it will result it less evaluation.

---
In reply to: [366631681](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366631681) [](ancestors = 366631681)

source/WinCompData/Expressions/Color.cs
simeoncran: CreateExpressionText should never have .Simplified calls in it. 
CreateExpressionText is called in 2 cases - one is from ToString() in which case it is expected to show you what you built. The other case is ToText, which will call Simplify before it calls CreateExpressionText().

---
In reply to: [366643883](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366643883) [](ancestors = 366643883)

source/WinCompData/Expressions/Scalar.cs
janny10: Yeah just realized square still existed

---
In reply to: [366644988](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366644988) [](ancestors = 366644988,366631681)

source/WinCompData/Expressions/Scalar.cs
simeoncran: This is attempting to handle commutivity. It's common to want to right "x * y * z". I think it's less likely to write "x / y / x"

---
In reply to: [366638710](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366638710) [](ancestors = 366638710)

source/WinCompData/Expressions/Scalar.cs
simeoncran: And the same explanation.

---
In reply to: [366639772](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366639772) [](ancestors = 366639772)

source/WinCompData/Expressions/Scalar.cs
simeoncran: Same comment as other cases - common cases are already handled. A general left == right is going to be very rare, so until we see it I'm leaving it out.

---
In reply to: [366634912](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366634912) [](ancestors = 366634912)

source/WinCompData/Expressions/LessThan.cs
simeoncran: I created what was needed for Lottie, plus sometimes some extra stuff. The implementation is only as complete as it needs to be. Adding more is goodness for the future, but it adds some maintenance burden.

---
In reply to: [366633790](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366633790) [](ancestors = 366633790)

source/WinCompData/Expressions/Scalar.cs
```diff
@@ -0,0 +1,485 @@
...
+            }
+
+            /// <inheritdoc/>
+            protected override string CreateExpressionText() => $"{Parenthesize(_value)}.{_channelName}";
+
+            /// <inheritdoc/>
+            protected override bool IsAtomic => true;
+        }
+
+        internal sealed new class Ternary : Scalar
```
janny10: >Ternary [](start = 34, length = 7)

Will there be support added for Ternaries involving nonscalars? Or does the Lottie translator not create ternaries involving nonscalars?

simeoncran: It currently doesn't need them, so not adding yet. This is one of the downsides of making the nodes have a static type - you have to create classes for each type for these operations.

---
In reply to: [366647826](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366647826) [](ancestors = 366647826)

source/WinCompData/Expressions/Scalar.cs
janny10: I meant either rename Subtract to Difference or Sum to Add - not rename both Subtract and Sum to Difference and Add respectively. This is in the interest of grammatical parallelism -- Subtract and Add are the operators and Difference and Sum are the result. 

---
In reply to: [366635889](https://github.com/windows-toolkit/Lottie-Windows/pull/213#discussion_r366635889) [](ancestors = 366635889)

PR #219, opened 2020-01-17
Strong type of Lottie Opacity.

simeoncran:
Opacity is special because it's expressed in Lottie as a percent, which is inconvenient for multiplication with alpha.
Giving an Opacity a strong type allows us to abstract the percent details, and also make the translator clearer because it's obvious where a value is being used for opacity.

### DIFF
--- a/source/LottieData/OpacityGradientStop.cs
+++ b/source/LottieData/OpacityGradientStop.cs
@@ -9,18 +9,18 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
 #endif
     sealed class OpacityGradientStop : GradientStop
     {
-        public OpacityGradientStop(double offset, double opacityPercent)
+        public OpacityGradientStop(double offset, Opacity opacity)
             : base(offset)
         {
-            OpacityPercent = opacityPercent;
+            Opacity = opacity;
         }
 
-        public double OpacityPercent { get; }
+        public Opacity Opacity { get; }
 
         /// <inheritdoc/>
         public override GradientStopKind Kind => GradientStopKind.Opacity;
 
         /// <inheritdoc/>
-        public override string ToString() => $"{OpacityPercent}%@{Offset}";
+        public override string ToString() => $"{Opacity.Value * 100}%@{Offset}";
     }
 }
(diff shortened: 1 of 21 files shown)


source/LottieData/OpacityGradientStop.cs
```diff
@@ -9,18 +9,18 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
...
 
-        public double OpacityPercent { get; }
+        public Opacity Opacity { get; }
 
         /// <inheritdoc/>
         public override GradientStopKind Kind => GradientStopKind.Opacity;
 
         /// <inheritdoc/>
-        public override string ToString() => $"{OpacityPercent}%@{Offset}";
+        public override string ToString() => $"{Opacity.Value * 100}%@{Offset}";
```
simeoncran: Opacity.Percent

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -953,13 +953,13 @@ ShapeOrVisual TranslateLayer(TranslationContext parentContext, Layer layer)
...
             // Get the opacity of the layer.
-            var layerOpacityPercent = context.TrimAnimatable(context.Layer.Transform.OpacityPercent);
+            var layerOpacity = context.TrimAnimatable(context.Layer.Transform.Opacity);
 
             // Convert the layer's in point and out point into absolute progress (0..1) values.
             var inProgress = GetInPointProgress(context);
             var outProgress = GetOutPointProgress(context);
 
-            if (inProgress > 1 || outProgress <= 0 || inProgress >= outProgress || layerOpacityPercent.AlwaysEquals(0))
+            if (inProgress > 1 || outProgress <= 0 || inProgress >= outProgress || layerOpacity.AlwaysEquals(LottieData.Opacity.Transparent))
```
eliezerpMS: >if (inProgress > 1 || outProgress <= 0 || inProgress >= outProgress || layerOpacity.AlwaysEquals(LottieData.Opacity.Transparent)) [](start = 12, length = 129)

Might as well start wrapping this line.

simeoncran: A lot of this code is about to change with the animated opacity stuff... mind if I leave it until later?

---
In reply to: [368164485](https://github.com/windows-toolkit/Lottie-Windows/pull/219#discussion_r368164485) [](ancestors = 368164485)

PR #223, opened 2020-01-21
Optimize generated code for collection population.

simeoncran:
If a collection (e.g. Shapes or Children) has only one child being added, don't both creating a local for it, just add it directly.

This change is designed to make the generated code a bit simpler.


source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1146,6 +1146,38 @@ void WriteCreateAssignment(CodeBuilder builder, ObjectData node, string createCa
                 }
             }
 
+            void WritePopulateShapesCollection(CodeBuilder builder, IList<CompositionShape> shapes, ObjectData node)
+            {
+                switch (shapes.Count)
```
derobinsMSFT: I'm curious about the choice of using a switch here, as opposed to something a little more concise like:

```
if (shapes.Count > 1)
{
     builder.WriteLine($"{Var} shapes = result{Deref}Shapes;");
}

foreach (shape in Shapes)
{
    builder.WriteComment(…);
    builder.WriteLine("${(shapes.Count == 1 ? "result{Deref}Shapes" : "shapes")} …);
}
```

Similar goes for the changes in `WriteAnimatedVisualCode`.

simeoncran: I love switches where they make it clear that you have exactly n cases, especially where each case is quite different.
The way you've written it there are 2 branches to consider (shapes.Count > 1) and (shapes.Count == 1). A reader trying to reason about the code has to track down each branch and consider all the cases they cover. 2 branches equals 4 possibilities. With the switch it's clear that there are only 3 cases to consider, and it's clear what happens in each case.
The downside is a little less code sharing.
Overall I think the switches are a win. Chat in person if you strongly disagree.
While I'm looking at it though, I dislike how one case returns and the others break. I'll fix that to be consistent.

---
In reply to: [369869664](https://github.com/windows-toolkit/Lottie-Windows/pull/223#discussion_r369869664) [](ancestors = 369869664)

PR #224, opened 2020-01-22
Optimizations for generated code.

simeoncran:
1. Call a method to bind an animator to Progress.
2. Call a method to bind a property to an expression.


source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -14,8 +14,10 @@
 using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Wui;
```
derobinsMSFT: Looks like this can be removed (cf. line 20).

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1569,51 +1658,82 @@ void StartAnimation(CodeBuilder builder, CompositionObject obj, ObjectData node,
...
+                if (referenceParameters.Length == 1 &&
+                    string.IsNullOrWhiteSpace(animation.Target))
                 {
-                    builder.WriteLine($"{SingletonExpressionAnimationName}{Deref}Target = {String(expressionAnimation.Target)};");
-                }
+                    var rp = referenceParameters[0];
+                    var referenceParameterName = GetReferenceParameterName(obj, localName, animationNode, rp);
 
-                foreach (var rp in animator.Animation.ReferenceParameters)
+                    // Special-case where there is exactly one reference parameter. Call a helper.
```
derobinsMSFT:
```suggestion
                    // Special case where there is exactly one reference parameter. Call a helper.
```

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -373,6 +375,9 @@ protected string GenerateCode()
             // Write the LoadedImageSurface byte arrays into the outer (IAnimatedVisualSource) class.
             WriteLoadedImageSurfaceArrays(builder);
 
+            // Write the method that binds AnimationController.Progress of animations to an expression.
```
derobinsMSFT: I would change the wording here a bit. How about:
```suggestion
            // Write the method that binds the AnimationController.Progress property of a given animation to an expression.
```

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: I think the grammar is correct as it is - "special-case" is a verb, "special case" is not.

---
In reply to: [369873338](https://github.com/windows-toolkit/Lottie-Windows/pull/224#discussion_r369873338) [](ancestors = 369873338)

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: It starts the animation as well.

---
In reply to: [369875297](https://github.com/windows-toolkit/Lottie-Windows/pull/224#discussion_r369875297) [](ancestors = 369875297)

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -1613,51 +1715,82 @@ void StartAnimation(CodeBuilder builder, CompositionObject obj, ObjectData node,
...
+                    if (propSetOwner == obj)
+                    {
+                        // Use the name of the local that is holding the property set.
+                        referenceParameterName = "propertySet";
                     }
                     else
                     {
-                        referenceParameterValueName = CallFactoryFromFor(animationNode, rp.Value);
+                        // Get the factory for the owner of the property set, and get the Properties object from it.
+                        referenceParameterName = CallFactoryFromFor(animationNode, propSetOwner);
```
scott-moore-ms: >CallFactoryFromFor [](start = 49, length = 18)

What does the "FromFor" denote here?  My original thought was because this was called from within a for loop, but that's not the case

scott-moore-ms:
What is the actual optimization that this change makes (I'm guessing it's related to when we call a factory vs not?)?

For me personally when looking at codegen changes, it's really super helpful to have a before/after sample. Without that, when trying to work through the logic it's extra mental tax very easy to get confused between what the *code generator code* is doing vs what the *generated code* is doing.

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: // Returns the code to call the factory for the given node from the given node.
A "node" represents an object that will be created by the generated code. The animationNode describes the factory we are currently generating code for, and the propSetOwner is the object that we want to get. CallFactoryFromFor will return code for an expression that will get the propSetOwner object. The reason we care about the "from" (i.e. the caller) is because we might return a method call, or we might return a field access, depending on whether another caller would have created the object and stored it in a field already.

---
In reply to: [373627550](https://github.com/windows-toolkit/Lottie-Windows/pull/224#discussion_r373627550) [](ancestors = 373627550)

simeoncran:
Example before:
                result.StartAnimation("t0", ScalarAnimation_0_to_1_0());
                var controller = result.TryGetAnimationController("t0");
                controller.Pause();
                controller.StartAnimation("Progress", RootProgress());
                result.StartAnimation("t1", ScalarAnimation_0_to_1_0());
                controller = result.TryGetAnimationController("t1");
                controller.Pause();
                controller.StartAnimation("Progress", RootProgress());
                result.StartAnimation("t2", t2ScalarAnimation_0_to_1_0());
                controller = result.TryGetAnimationController("t2");
                controller.Pause();
                controller.StartAnimation("Progress", RootProgress());
                result.StartAnimation("t3", ScalarAnimation_0_to_1_0());
                controller = result.TryGetAnimationController("t3");
                controller.Pause();
                controller.StartAnimation("Progress", RootProgress());


Example after:
                StartProgressBoundAnimation(result, "t0", ScalarAnimation_0_to_1_0(), RootProgress());
                StartProgressBoundAnimation(result, "t1", ScalarAnimation_0_to_1_0(), RootProgress());
                StartProgressBoundAnimation(result, "t2", t2ScalarAnimation_0_to_1_0(), RootProgress());
                StartProgressBoundAnimation(result, "t3", ScalarAnimation_0_to_1_0(), RootProgress());

...and the helper method that is generated once:
        static void StartProgressBoundAnimation(
                CompositionObject target,
                String animatedPropertyName,
                CompositionAnimation animation,
                ExpressionAnimation controllerProgressExpression)
        {
            target.StartAnimation(animatedPropertyName, animation);
            var controller = target.TryGetAnimationController(animatedPropertyName);
            controller.Pause();
            controller.StartAnimation("Progress", controllerProgressExpression);
        }




---
In reply to: [580861580](https://github.com/windows-toolkit/Lottie-Windows/pull/224#issuecomment-580861580) [](ancestors = 580861580)

PR #229, opened 2020-01-25
Transfer object descriptions and names during tree reduction.

simeoncran:
Tree reduction is an optimization on the Composition graph where we attempt to coalesce and remove redundant visuals and shapes. Previously this resulted in the loss of some descriptions and names on objects. Now we attempt to transfer that information out of the objects that are being removed into the objects that are taking over their roles.

The result is that some names that were being lost from the Lottie files now show up as comments in the generated code.


source/UIData/Tools/TreeReducer.cs
```diff
@@ -325,18 +325,26 @@ static void CoalesceContainerShapes(ObjectGraph<Node> graph)
                 container.Shapes.Clear();
 
                 // Insert the first child where the container was.
-                parent.Shapes[index] = children[0];
+                var child0 = children[0];
```
eliezerpMS: >child0 [](start = 20, length = 6)

style nit: could this just be name firstChild?

source/UIData/Tools/TreeReducer.cs
```diff
@@ -325,18 +325,26 @@ static void CoalesceContainerShapes(ObjectGraph<Node> graph)
...
 
                 // Fix the parent pointer in the graph.
-                graph[children[0]].Parent = (CompositionObject)parent;
+                graph[child0].Parent = (CompositionObject)parent;
 
                 // Insert the rest of the children.
                 for (var i = 1; i < children.Length; i++)
                 {
-                    parent.Shapes.Insert(index + i, children[i]);
+                    var childI = children[i];
```
eliezerpMS: >childI [](start = 24, length = 6)

style nit: could this be named something like currentChild?

source/UIData/Tools/TreeReducer.cs
eliezerpMS: also same for the below usages.

---
In reply to: [370906285](https://github.com/windows-toolkit/Lottie-Windows/pull/229#discussion_r370906285) [](ancestors = 370906285)

source/UIData/Tools/TreeReducer.cs
eliezerpMS: also same for the below usages.

---
In reply to: [370906307](https://github.com/windows-toolkit/Lottie-Windows/pull/229#discussion_r370906307) [](ancestors = 370906307)

source/UIData/Tools/TreeReducer.cs
```diff
@@ -392,18 +400,70 @@ static void CoalesceContainerVisuals(ObjectGraph<Node> graph)
...
+            if (!string.IsNullOrWhiteSpace(fromShortDescription))
+            {
+                var toShortDescription = to.ShortDescription;
+                if (string.IsNullOrWhiteSpace(toShortDescription))
+                {
+                    to.ShortDescription = fromShortDescription;
+                }
+                else
+                {
+                    to.ShortDescription = $"{toShortDescription} / {fromShortDescription}";
```
eliezerpMS: >$"{toShortDescription} / {fromShortDescription}" [](start = 42, length = 48)

How do we expect the short descriptions to be? Would it make sense to have them separated as different lines or would that make making them into comments too difficult?

source/UIData/Tools/TreeReducer.cs
simeoncran: Let's try N

---
In reply to: [370906349](https://github.com/windows-toolkit/Lottie-Windows/pull/229#discussion_r370906349) [](ancestors = 370906349,370906307)

source/UIData/Tools/TreeReducer.cs
simeoncran: Let's try N.

---
In reply to: [370906347](https://github.com/windows-toolkit/Lottie-Windows/pull/229#discussion_r370906347) [](ancestors = 370906347,370906285)

source/UIData/Tools/TreeReducer.cs
simeoncran: Short descriptions are supposed to be single lines. Long descriptions are allowed to be multi lines.
        /// Gets or sets a short description of the object, suitable for
        /// use in a single line comment.


---
In reply to: [370906520](https://github.com/windows-toolkit/Lottie-Windows/pull/229#discussion_r370906520) [](ancestors = 370906520)

PR #231, opened 2020-01-29
Use nameof to keep Yaml field names in sync with Lottie data property names

simeoncran:
Some of the Yaml field names were wrong. This change gets rid of the hand-written strings and replaces them with nameof(...) so that the names stay in sync with LottieData.
And did some small cleanup and refactoring while I was there.

This does change the Yaml output from what it was - that is intentional.

### DIFF
--- a/source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
+++ b/source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
@@ -326,8 +326,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
         }
 
         YamlObject FromMask(Mask mask)
-        {
-            var result = new YamlMap
+            => new YamlMap
             {
                 { nameof(mask.Name), mask.Name },
                 { nameof(mask.Inverted), mask.Inverted },
@@ -335,8 +334,6 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
                 { nameof(mask.Opacity), FromAnimatable(mask.Opacity) },
                 { nameof(mask.Points), FromAnimatable(mask.Points, p => FromSequence(p, FromBezierSegment)) },
             };
-            return result;
-        }
 
         YamlObject FromShapeGroup(ShapeGroup content, YamlMap superclassContent)
         {
--- a/source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
+++ b/source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
@@ -394,7 +394,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
             result.Add(nameof(content.Position), FromAnimatable(content.Position));
             result.Add(nameof(content.Anchor), FromAnimatable(content.Anchor));
             result.Add(nameof(content.Opacity), FromAnimatable(content.Opacity));
-            result.Add(nameof(content.RotationDegrees), FromAnimatable(content.RotationDegrees));
+            result.Add(nameof(content.Rotation), FromAnimatable(content.Rotation));
             return result;
         }
 
@@ -423,16 +423,9 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
 
         YamlObject FromAnimatable<T>(Animatable<T> animatable, Func<T, YamlObject> valueSelector)
             where T : IEquatable<T>
-        {
-            if (!animatable.IsAnimated)
-            {
-                return valueSelector(animatable.InitialValue);
-            }
-            else
-            {
-                return FromSpan<KeyFrame<T>>(animatable.KeyFrames, kf => FromKeyFrame(kf, valueSelector));
-            }
-        }
+            => animatable.IsAnimated
+                ? FromSpan<KeyFrame<T>>(animatable.KeyFrames, kf => FromKeyFrame(kf, valueSelector))
+                : valueSelector(animatable.InitialValue);
 
         YamlObject FromAnimatable(Animatable<Color> animatable) => FromAnimatable(animatable, Scalar);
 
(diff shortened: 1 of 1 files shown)


source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -440,22 +437,19 @@ YamlObject FromAnimatable<T>(Animatable<T> animatable, Func<T, YamlObject> value
             }
         }
 
-        YamlObject FromAnimatable(Animatable<double> animatable)
-            => FromAnimatable(animatable, FromDouble);
-
-        static YamlObject FromDouble(double value) => (YamlScalar)value;
+        YamlObject FromAnimatable(Animatable<Color> animatable) => FromAnimatable(animatable, Scalar);
```
janny10: Is it alright that this no longer checks for null?

simeoncran: I'm not seeing what you're referring to - where was this checking for null before?

---
In reply to: [372679165](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r372679165) [](ancestors = 372679165)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -39,7 +39,7 @@ YamlMap GetLottieObjectContent(LottieObject obj)
 
             var result = new YamlMap
             {
-                { "Name", name },
+                { nameof(obj.Name), name },
```
derobinsMSFT: >nameof(obj.Name) [](start = 18, length = 16)

Why not `nameof(name)`?

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -143,62 +143,59 @@ YamlSequence FromSpan<T>(ReadOnlySpan<T> collection, Func<T, YamlObject> selecto
...
 
-        YamlObject FromLayersAsset(LayerCollectionAsset asset)
+        YamlObject FromLayersAsset(LayerCollectionAsset asset, YamlMap superclassContent)
         {
-            var result = new YamlMap
-            {
-                { "Id", asset.Id },
-                { "Layers", FromLayerCollection(asset.Layers) },
-            };
+            var result = superclassContent;
```
derobinsMSFT: >var result = superclassContent; [](start = 12, length = 31)

Just for my own understanding-- why not just update and return `superclassContent`?

derobinsMSFT: Also, why is the return type specified as `YamlObject` when the actual object being returned is (or at least starts as) a `YamlMap`? (I noticed that `YamlMap` is derived from `YamlObject`.)

---
In reply to: [373274202](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r373274202) [](ancestors = 373274202)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -332,28 +329,28 @@ YamlObject FromMask(Mask mask)
         {
             var result = new YamlMap
```
derobinsMSFT: >var result = new YamlMap [](start = 12, length = 24)

Why not just return `new YamlMap`?

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -533,8 +527,8 @@ YamlMap FromKeyFrame<T>(KeyFrame<T> keyFrame, Func<T, YamlObject> valueSelector)
                 if (cp1 != Vector3.Zero || cp2 != Vector3.Zero)
                 {
                     // Spatial bezier
-                    result.Add("ControlPoint1", FromVector3(cp1));
-                    result.Add("ControlPoint2", FromVector3(cp2));
+                    result.Add(nameof(v3kf.SpatialControlPoint1), FromVector3(cp1));
```
derobinsMSFT: >v3kf.SpatialControlPoint1 [](start = 38, length = 25)

Why not `nameof(cp[1|2])`?

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
derobinsMSFT: Looks like @janny10 was referring to line 448 from the previous version of this file.

---
In reply to: [372701615](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r372701615) [](ancestors = 372701615,372679165)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
scott-moore-ms: [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof)

nameof(name) would return "name" where the desire is for this to be "Name"

---
In reply to: [373270644](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r373270644) [](ancestors = 373270644)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
simeoncran: These choices were all about patterns. 
a. All of the "From..." methods return YamlObject. Some of them might be maps (they probably all are) but the callers should never care - that's just an implementation detail of the "From..." methods. The (implied) contract is that if I give you a Lottie object, you'll give me a YamlObject that represents it. The details of how you package it are none of my business. 
b. The "result" pattern is a consistent way of saying to the reader "this is the thing I'm returning". If there was a "return superclassContent" it would suggest that we're returning the superclass content, but actually we're returning that plus some other stuff.

---
In reply to: [373275715](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r373275715) [](ancestors = 373275715,373274202)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
simeoncran: I have no good reason. Will fix.

---
In reply to: [373276939](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r373276939) [](ancestors = 373276939)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
simeoncran: cp1 is a Vector3, so the name would be "Vector3".

---
In reply to: [373279764](https://github.com/windows-toolkit/Lottie-Windows/pull/231#discussion_r373279764) [](ancestors = 373279764)

PR #234, opened 2020-01-30
Simplify node naming.

simeoncran:
Previously we special-cased the naming of the root node and the root progress animations during codegen, but did some other special naming during translation. With this change we do all of the special naming at codegen time.

This change also includes better handling of name collisions in the node namer. If 2 or more nodes get the same name, we have to append a counter value to the name to differentiate them. Previously we appended a counter even if there was only 1 node, in order to prevent C++ from getting confused when a method name was the same as a type name. Now we keep track of whether a name is also a type name, and can therefore avoid appending the counter more often. This makes for much nicer naming (i.e. they don't have a _0 suffix) for most singleton objects.

### DIFF
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -125,6 +125,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             if (_addDescriptions)
             {
                 Describe(_rootVisual, "The root of the composition.", string.Empty);
+                Name(_rootVisual, "Root");
             }
 
             // Add the master progress property to the visual.
@@ -2860,11 +2861,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             if (_addDescriptions)
             {
                 Describe(result, $"Color bound to root CompositionPropertySet value: {bindingName}", bindingName);
-            }
 
-            // Name the brush the same as the binding name. This will allow the code generator to
-            // give it a more meaningful name.
-            Name(result, bindingName);
+                // Name the brush with a name that includes the binding name. This will allow the code generator to
+                // give its factory a more meaningful name.
+                Name(result, $"BoundColor_{bindingName}");
+            }
 
             if (opacity.IsAnimated)
             {
@@ -3641,6 +3642,16 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             {
                 bindingAnimation = _c.CreateExpressionAnimation(ScaledAndOffsetRootProgress(scale, offset));
                 bindingAnimation.SetReferenceParameter(RootName, _rootVisual);
+                if (_addDescriptions)
+                {
+                    // Give the animation a nice readable name in codegen.
+                    var name = key.Offset != 0 || key.Scale != 1
+                        ? "RootProgressRemapped"
+                        : "RootProgress";
+
+                    Name(bindingAnimation, name);
+                }
+
                 _progressBindingAnimations.Add(key, bindingAnimation);
             }
 
@@ -4333,17 +4344,18 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         void Describe(IDescribable obj, string longDescription, string shortDescription = null)
         {
             Debug.Assert(_addDescriptions, "This method should only be called if the user wanted descriptions");
-            Debug.Assert(obj.ShortDescription == null, "Descriptions should never get set more than once");
-            Debug.Assert(obj.LongDescription == null, "Descriptions should never get set more than once");
+            Debug.Assert(obj.ShortDescription is null, "Descriptions should never get set more than once");
+            Debug.Assert(obj.LongDescription is null, "Descriptions should never get set more than once");
 
             obj.ShortDescription = shortDescription ?? longDescription;
             obj.LongDescription = longDescription;
         }
 
-        // Sets a name on an object.
-        void Name(IDescribable obj, string name)
+        // Sets a name on an object. This allows the code generator to give the object
+        // a more meaningful name.
+        static void Name(IDescribable obj, string name)
         {
-            Debug.Assert(obj.Name == null, "Names should never get set more than once");
+            Debug.Assert(obj.Name is null, "Names should never get set more than once");
             obj.Name = name;
         }
 
@@ -4418,14 +4430,15 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         // A pair of doubles used as a key in a dictionary.
         sealed class ScaleAndOffset
         {
-            readonly double _scale;
-            readonly double _offset;
-
             internal ScaleAndOffset(double scale, double offset)
             {
-                _scale = scale;
-                _offset = offset;
+                Scale = scale;
+                Offset = offset;
             }
+
+            internal double Scale { get; }
+
+            internal double Offset { get; }
 
             public override bool Equals(object obj)
             {
@@ -4435,10 +4448,10 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                     return false;
                 }
 
-                return other._scale == _scale && other._offset == _offset;
+                return other.Scale == Scale && other.Offset == Offset;
             }
 
-            public override int GetHashCode() => _scale.GetHashCode() ^ _offset.GetHashCode();
+            public override int GetHashCode() => Scale.GetHashCode() ^ Offset.GetHashCode();
         }
 
         // A type that is either a CompositionType or a Visual.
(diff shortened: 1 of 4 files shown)


source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -78,61 +67,71 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
                 //       the C++ compiler will complain that CompositionPath is not a type.
-                //       So to ensure we don't hit that case, always append a counter suffix.
-
-                // Use only as many digits as necessary to express the largest count.
-                var digitsRequired = (int)Math.Ceiling(Math.Log10(nodeList.Count + 1));
-                var counterFormat = new string('0', digitsRequired);
-
-                for (var i = 0; i < nodeList.Count; i++)
+                //       So to ensure we don't hit that case, append a counter suffix, unless the name
+                //       known to not be a type name.
```
derobinsMSFT: >known [](start = 25, length = 5)

"is known"

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -78,61 +67,71 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
-                CompositionObjectType.CompositionColorGradientStop => DescribeCompositionColorGradientStop((CompositionColorGradientStop)obj),
-                CompositionObjectType.StepEasingFunction => DescribeStepEasingFunction((StepEasingFunction)obj),
+                CompositionObjectType.ExpressionAnimation => NameExpressionAnimation((ExpressionAnimation)obj),
+                CompositionObjectType.CompositionColorBrush => NameCompositionColorBrush((CompositionColorBrush)obj),
+                CompositionObjectType.CompositionColorGradientStop => NameCompositionColorGradientStop((CompositionColorGradientStop)obj),
+                CompositionObjectType.StepEasingFunction => NameStepEasingFunction((StepEasingFunction)obj),
 
-                // All other cases, just ToString() the type name.
-                _ => obj.Type.ToString(),
+                _ => NameCompositionObjectType(obj.Type),
```
derobinsMSFT: >NameCompositionObjectType [](start = 21, length = 25)

Looks like we now remove the "Composition" prefix just for the default case; any chance "Composition" could still be in the name for any of the other cases?

simeoncran: No, but if it was, it's intentional. This branch is just a catch-all for when we don't have an algorithm for setting the name.

---
In reply to: [373658838](https://github.com/windows-toolkit/Lottie-Windows/pull/234#discussion_r373658838) [](ancestors = 373658838)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -31,45 +31,34 @@ static class NodeNamer<TNode>
...
+                };
+
+                if (!nodesByName.TryGetValue(nodeName, out var nodeList))
                 {
                     nodeList = new List<TNode>();
-                    nodesByTypeName.Add(baseName, nodeList);
+                    nodesByName.Add(nodeName, nodeList);
                 }
 
                 nodeList.Add(node);
```
janny10: I don't understand this.. does var nodeList somehow persist over iterations or is this a mistake..?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4373,22 +4384,23 @@ static CanvasGeometryCombine GeometryCombine(MergePaths.MergeMode mergeMode)
         void Describe(IDescribable obj, string longDescription, string shortDescription = null)
         {
             Debug.Assert(_addDescriptions, "This method should only be called if the user wanted descriptions");
-            Debug.Assert(obj.ShortDescription == null, "Descriptions should never get set more than once");
-            Debug.Assert(obj.LongDescription == null, "Descriptions should never get set more than once");
+            Debug.Assert(obj.ShortDescription is null, "Descriptions should never get set more than once");
```
janny10: why is this static?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4164,8 +4175,8 @@ protected override WinCompData.Expressions.Scalar Simplify()
 
         // Returns the name of a variable on the root property set that advances linearly from 0 to 1 over the
         // given range of Progress.
-        TimeRemap GetRemappedProgress(double tRangeLow, double tRangeHigh)
-            => new TimeRemap(tRangeLow, tRangeHigh, ExpressionFactory.RootProgress);
+        TimeRemap GetRemappedProgress(double tRangeLow, double tRangeHigh) =>
```
janny10: >is null [](start = 46, length = 7)

Should there be a TODO to convert all remaining == null to is null?

simeoncran: Yes, in a OneNote somewhere. Not in the code. We really shouldn't have any TODO's.

---
In reply to: [374410051](https://github.com/windows-toolkit/Lottie-Windows/pull/234#discussion_r374410051) [](ancestors = 374410051)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Not sure what you're pointing at (seems to have moved in a merge). But in general, if it can be static, we make it static. It tells the reader "I don't depend on the state of "this"" And it makes the call slightly faster.

---
In reply to: [374409524](https://github.com/windows-toolkit/Lottie-Windows/pull/234#discussion_r374409524) [](ancestors = 374409524)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -78,61 +67,71 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
-                var digitsRequired = (int)Math.Ceiling(Math.Log10(nodeList.Count + 1));
-                var counterFormat = new string('0', digitsRequired);
-
-                for (var i = 0; i < nodeList.Count; i++)
+                //       So to ensure we don't hit that case, append a counter suffix, unless the name
+                //       is known to not be a type name.
+                if (nodeList.Count == 1 && nodeName.IsNotATypeName)
                 {
-                    yield return (nodeList[i], $"{baseName}_{i.ToString(counterFormat)}");
+                    // The name is unique, is not a type name, or is in the whitelist, so no need for a suffix.
```
simeoncran: >or is in the whitelist, [](start = 63, length = 23)

There is no whitelist!

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -311,8 +297,34 @@ static string StripPrefix(string name, string prefix)
...
+        {
+            NodeName(string name, bool isNotATypeName)
+            {
+                Name = name;
+                IsNotATypeName = isNotATypeName;
+            }
+
+            internal string Name { get; }
+
+            // True iff the name is also the name of a type.
```
simeoncran: > True iff the name is also the name of a type. [](start = 14, length = 46)

Inverted.

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -78,61 +67,71 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
...
+        // Returns a name for the given CompositionObject, suitable for use as an identifier.
+        static NodeName NameCompositionObject(TNode node, CompositionObject obj)
         {
-            var result = obj.Type switch
+            var name = NameOf(obj);
+
+            if (name != null)
+            {
+                // The object has a name, so use it.
+                return NodeName.FromNonTypeName(name);
```
derobinsMSFT: >FromNonTypeName [](start = 32, length = 15)

Can a CompositionObject's name ever match that of a type?

simeoncran: Yep. If the translator decides to use a name that is also a type name. Or maybe someone #include'd a type declaration. So yes, things could go wrong, but we control the translator, and we trust the user will figure out the problem if they start modifying our generated code.

---
In reply to: [374434484](https://github.com/windows-toolkit/Lottie-Windows/pull/234#discussion_r374434484) [](ancestors = 374434484)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -142,106 +141,96 @@ static string DescribeCompositionColorBrush(CompositionColorBrush obj)
...
                 }
                 else
                 {
-                    // The color is bound to an expression.
-                    var objectName = ((IDescribable)obj).Name;
-
-                    return string.IsNullOrWhiteSpace(objectName)
-                        ? "BoundGradientStop"
-                        : $"{objectName}GradientStop";
+                    return NodeName.FromNonTypeName("GradientStop");
```
simeoncran: AnimatedGradientStop

PR #235, opened 2020-01-30
Add a type to represent rotation.

simeoncran:
Previously we just used a double and you had to know that it was represented in degrees. Now the type directly represents 2d rotation and has properties to express the rotation in degrees and in radians.

This change also fixed up a few places I noticed where the Opacity type was still talking about OpacityPercent.

### DIFF
--- a/source/LottieData/LottieData.projitems
+++ b/source/LottieData/LottieData.projitems
@@ -59,6 +59,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)\Repeater.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\RepeaterTransform.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\RoundedCorner.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)\Rotation.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Sequence.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Serialization\LottieCompositionXmlSerializer.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Serialization\LottieCompositionYamlSerializer.cs" />
--- a/source/LottieData/LottieData.projitems
+++ b/source/LottieData/LottieData.projitems
@@ -58,8 +58,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)\Rectangle.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Repeater.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\RepeaterTransform.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\RoundedCorner.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Rotation.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)\RoundedCorner.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Sequence.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Serialization\LottieCompositionXmlSerializer.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\Serialization\LottieCompositionYamlSerializer.cs" />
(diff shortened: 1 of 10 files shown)


source/LottieData/Rotation.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+#if PUBLIC_LottieData
+    public
+#endif
+    struct Rotation : IEquatable<Rotation>
+    {
+        Rotation(double degrees)
+        {
+            Degrees = degrees;
+        }
```
derobinsMSFT: What happens if I assume/guess that Rotation takes radians by default? Is using degrees (when creating a Rotation object) enforced somehow, or at least well-documented?

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -400,7 +400,7 @@ YamlObject FromTransform(Transform content, YamlMap superclassContent)
             result.Add("Position", FromAnimatable(content.Position));
             result.Add("Anchor", FromAnimatable(content.Anchor));
             result.Add("OpacityPercent", FromAnimatable(content.Opacity, FromOpacityPercent));
-            result.Add("RotationDegrees", FromAnimatable(content.RotationDegrees));
+            result.Add("RotationDegrees", FromAnimatable(content.Rotation, FromRotation));
```
derobinsMSFT: >RotationDegrees [](start = 24, length = 15)

Remove "Degrees".

source/LottieData/Rotation.cs
simeoncran: The constructor is private, so the only place you could make that mistake is from within this class. At the moment the only way for other classes to create a Rotation is via Rotation.FromDegrees, or Rotation.None.
As long as the constructor remains private and construction is done via named methods the contract will be clear to each user.

---
In reply to: [373204129](https://github.com/windows-toolkit/Lottie-Windows/pull/235#discussion_r373204129) [](ancestors = 373204129)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
simeoncran: This is addressed in [https://github.com/windows-toolkit/Lottie-Windows/pull/231](https://github.com/windows-toolkit/Lottie-Windows/pull/231) when the names will be synched to the property names.

---
In reply to: [373210280](https://github.com/windows-toolkit/Lottie-Windows/pull/235#discussion_r373210280) [](ancestors = 373210280)

source/LottieData/LottieData.projitems
```diff
@@ -59,6 +59,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)\Repeater.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\RepeaterTransform.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)\RoundedCorner.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)\Rotation.cs" />
```
simeoncran: Not alphabetical

source/LottieData/Rotation.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
+{
+    /// <summary>
+    /// A rotation value.
+    /// </summary>
+#if PUBLIC_LottieData
+    public
+#endif
+    struct Rotation : IEquatable<Rotation>
```
simeoncran: readonly

source/LottieData/Rotation.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public double Radians => Math.PI * Degrees / 180.0;
+
+        public static Rotation None => new Rotation(0);
+
+        public static Rotation FromDegrees(double value) => new Rotation(value);
+
+        public bool Equals(Rotation other) => other.Degrees == Degrees;
+
+        public override bool Equals(object obj) => obj is Rotation other ? other.Equals(this) : false;
```
simeoncran: obj is Rotation other && other.Equals(this);

PR #236, opened 2020-01-30
Add ability to get the type for a given CompositionPropertySet property name.

simeoncran:
This also gets rid of the CompositionPropertySet.IsEmpty property in order to reduce the number of non-standard (i.e. not in Windows.UI.Composition) apis on the class.

This change is in preparation for some new features coming to property binding.

### DIFF
--- a/source/UIData/Tools/Optimizer.cs
+++ b/source/UIData/Tools/Optimizer.cs
@@ -495,29 +495,48 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
 
             result = CacheAndInitializeCompositionObject(obj, result);
 
-            foreach (var prop in obj.ColorProperties)
+            foreach (var (name, type) in obj.Names)
             {
-                result.InsertColor(prop.Key, prop.Value);
-            }
+                switch (type)
+                {
+                    case WinCompData.MetaData.PropertySetValueType.Color:
+                        {
+                            obj.TryGetColor(name, out var value);
+                            result.InsertColor(name, value);
+                            break;
+                        }
 
-            foreach (var prop in obj.ScalarProperties)
-            {
-                result.InsertScalar(prop.Key, prop.Value);
-            }
+                    case WinCompData.MetaData.PropertySetValueType.Scalar:
+                        {
+                            obj.TryGetScalar(name, out var value);
+                            result.InsertScalar(name, value);
+                            break;
+                        }
 
-            foreach (var prop in obj.Vector2Properties)
-            {
-                result.InsertVector2(prop.Key, prop.Value);
-            }
+                    case WinCompData.MetaData.PropertySetValueType.Vector2:
+                        {
+                            obj.TryGetVector2(name, out var value);
+                            result.InsertVector2(name, value);
+                            break;
+                        }
 
-            foreach (var prop in obj.Vector3Properties)
-            {
-                result.InsertVector3(prop.Key, prop.Value);
-            }
+                    case WinCompData.MetaData.PropertySetValueType.Vector3:
+                        {
+                            obj.TryGetVector3(name, out var value);
+                            result.InsertVector3(name, value);
+                            break;
+                        }
 
-            foreach (var prop in obj.Vector4Properties)
-            {
-                result.InsertVector4(prop.Key, prop.Value);
+                    case WinCompData.MetaData.PropertySetValueType.Vector4:
+                        {
+                            obj.TryGetVector4(name, out var value);
+                            result.InsertVector4(name, value);
+                            break;
+                        }
+
+                    default:
+                        throw new InvalidOperationException();
+                }
             }
 
             StartAnimationsAndFreeze(obj, result);
--- a/source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
+++ b/source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
@@ -778,15 +778,15 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen
             nodes.OrderBy(n => n.Name, AlphanumericStringComparer.Instance);
 
         static string PropertySetValueType(PropertySetValueType value)
-    => value switch
-    {
-        WinCompData.MetaData.PropertySetValueType.Color => "Color",
-        WinCompData.MetaData.PropertySetValueType.Scalar => "Scalar",
-        WinCompData.MetaData.PropertySetValueType.Vector2 => "Vector2",
-        WinCompData.MetaData.PropertySetValueType.Vector3 => "Vector3",
-        WinCompData.MetaData.PropertySetValueType.Vector4 => "Vector4",
-        _ => throw new InvalidOperationException(),
-    };
+            => value switch
+            {
+                WinCompData.MetaData.PropertySetValueType.Color => "Color",
+                WinCompData.MetaData.PropertySetValueType.Scalar => "Scalar",
+                WinCompData.MetaData.PropertySetValueType.Vector2 => "Vector2",
+                WinCompData.MetaData.PropertySetValueType.Vector3 => "Vector3",
+                WinCompData.MetaData.PropertySetValueType.Vector4 => "Vector4",
+                _ => throw new InvalidOperationException(),
+            };
 
         string PropertySetValueInitializer(CompositionPropertySet propertySet, string propertyName, WinCompData.MetaData.PropertySetValueType propertyType)
             => propertyType switch
--- a/source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
+++ b/source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
@@ -10,6 +10,7 @@ using System.Text;
 using Microsoft.Toolkit.Uwp.UI.Lottie.GenericData;
 using Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools;
 using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData;
+using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.MetaData;
 using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Mgcg;
 using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Wui;
 using Expr = Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Expressions;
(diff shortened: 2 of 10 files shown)


source/WinCompData/CompositionPropertySet.cs
```diff
@@ -4,21 +4,25 @@
...
     public
 #endif
     sealed class CompositionPropertySet : CompositionObject
     {
-        HashSet<string> _names;
+        static readonly SortedDictionary<string, PropertySetValueType> s_empty = new SortedDictionary<string, PropertySetValueType>();
+
+        // Initialized to an empty sentinel value, then replaced with a new object
+        // when the first value is added.
+        SortedDictionary<string, PropertySetValueType> _names = s_empty;
```
janny10: >_names [](start = 55, length = 6)

Unsure why we're doing this?

source/WinCompData/CompositionPropertySet.cs
```diff
@@ -32,80 +36,96 @@ internal CompositionPropertySet(CompositionObject owner)
...
+        /// Returns the names and types of the values that have been added to this <see cref="CompositionPropertySet"/>.
+        /// The results are ordered by name.
+        /// </summary>
+        public IReadOnlyDictionary<string, PropertySetValueType> Names => _names;
 
         /// <inheritdoc/>
         public override CompositionObjectType Type => CompositionObjectType.CompositionPropertySet;
 
-        void Insert<T>(string propertyName, in T value, ref PropertyBag<T> bag)
+        void Insert<T>(string propertyName, in T value, PropertySetValueType type, ref PropertyBag<T> bag)
```
janny10: >Insert [](start = 13, length = 6)

is it worth checking Type isn't None?

janny10: >Insert [](start = 13, length = 6)

Don't CompositionPropertySets support storing values of same name/diff type? Could be mistaken..

source/WinCompData/CompositionPropertySet.cs
```diff
@@ -32,80 +36,96 @@ internal CompositionPropertySet(CompositionObject owner)
...
+            {
+                _names.Add(propertyName, type);
+            }
 
             // Set the value.
             bag.SetValue(propertyName, in value);
         }
 
-        CompositionGetValueStatus TryGet<T>(string propertyName, ref PropertyBag<T> bag, out T value)
+        CompositionGetValueStatus TryGet<T>(string propertyName, PropertySetValueType type, ref PropertyBag<T> bag, out T value)
```
janny10: Is this the function that achieves what's described in the description? If not, I'm missing something please enlighten. If so, I don't understand how this differs greatly from preexisting paradigm. This is still an internal method that returns the value associated with the propertyName, and its callers don't expose the property type. Moreover, the property type can be assumed by whether the function succeeds/returns a valid value. It seems the only change here is instead of not setting out T value, we are setting it to default(t) and returning a new error value..

source/WinCompData/CompositionPropertySet.cs
janny10: Also, nit - any particular reason for "Sorted"Dictionary vs Hashtable? Would accept that it's arbitrary

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -776,6 +777,53 @@ static LoadedImageSurfaceInfo LoadedImageSurfaceInfoFromObjectData(ObjectData no
...
+        static string PropertySetValueType(PropertySetValueType value)
+    => value switch
+    {
+        WinCompData.MetaData.PropertySetValueType.Color => "Color",
+        WinCompData.MetaData.PropertySetValueType.Scalar => "Scalar",
+        WinCompData.MetaData.PropertySetValueType.Vector2 => "Vector2",
+        WinCompData.MetaData.PropertySetValueType.Vector3 => "Vector3",
+        WinCompData.MetaData.PropertySetValueType.Vector4 => "Vector4",
+        _ => throw new InvalidOperationException(),
+    };
```
derobinsMSFT: Indentation

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -776,6 +777,53 @@ static LoadedImageSurfaceInfo LoadedImageSurfaceInfoFromObjectData(ObjectData no
...
+            };
+
+        string PropertySetColorValueInitializer(CompositionPropertySet propertySet, string propertyName)
+            => propertySet.TryGetColor(propertyName, out var value) == CompositionGetValueStatus.Succeeded
+                    ? _stringifier.Color(value)
+                    : throw new InvalidOperationException();
+
+        string PropertySetScalarValueInitializer(CompositionPropertySet propertySet, string propertyName)
+            => propertySet.TryGetScalar(propertyName, out var value) == CompositionGetValueStatus.Succeeded
+                    ? _stringifier.Float(value)
```
derobinsMSFT: >Float [](start = 35, length = 5)

Is it alright to use Float here? Any possible Scalar values that wouldn't be Floats or Float subtypes?

source/UIData/Tools/Optimizer.cs
```diff
@@ -495,29 +495,48 @@ CompositionPropertySet GetCompositionPropertySet(CompositionPropertySet obj)
...
-            foreach (var prop in obj.ColorProperties)
+            foreach (var (name, type) in obj.Names)
             {
-                result.InsertColor(prop.Key, prop.Value);
-            }
+                switch (type)
+                {
+                    case WinCompData.MetaData.PropertySetValueType.Color:
+                        {
+                            obj.TryGetColor(name, out var value);
```
derobinsMSFT: >TryGetColor [](start = 32, length = 11)

Should you be checking that the `Try*` calls here succeed?

source/WinCompData/CompositionPropertySet.cs
simeoncran: The contract (that we defined) on the "Names" property is that it returns the names ordered alphabetically. If we used a hashtable then we'd have to sort the names before handing them out.

---
In reply to: [373654743](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373654743) [](ancestors = 373654743)

simeoncran: This guarantees that _names is never null, so we never need to check for null.
It's particularly handy when someone asks for the Names in the CompositionPropertySet - we just hand out the dictionary _names value and it's guaranteed to be a valid dictionary. Keep in mind that the common case is the the CompositionPropertySet has no entries in it, so most of the time we'll be handing out that s_empty value when it's enumerated.
If we could use C# 8 nullables (which we will be able to eventually, when UWP catches up to C# 8) then this pattern will allow us to to declare _names non-nullable. Non-nullability is a nice property because it removes one of the failure cases (i.e. the compiler can guarantee that you'll never dereference a null).

---
In reply to: [373640601](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373640601) [](ancestors = 373640601)

source/WinCompData/CompositionPropertySet.cs
simeoncran: Because you can reason about that possibility by just looking at this one file I don't bother checking. But that's a good use for Debug.Assert. I'll add.

---
In reply to: [373642309](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373642309) [](ancestors = 373642309)

source/WinCompData/CompositionPropertySet.cs
janny10: Why not just set it to the value of s_empty instead of having s_empty? 

---
In reply to: [373733154](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373733154) [](ancestors = 373733154,373640601)

source/WinCompData/CompositionPropertySet.cs
simeoncran: They do not.

---
In reply to: [373643399](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373643399) [](ancestors = 373643399)

source/WinCompData/CompositionPropertySet.cs
simeoncran: Because that would require allocating a new SortedDictionary for every CompositionProperytSet. This way we get to share it.

---
In reply to: [373733461](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373733461) [](ancestors = 373733461,373733154,373640601)

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: The C# type system guarantees it to be safe. But anyway Scalar is just another name for Float in CompositionPropertySet.

---
In reply to: [373664308](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373664308) [](ancestors = 373664308)

source/WinCompData/CompositionPropertySet.cs
janny10: understood

---
In reply to: [373733864](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373733864) [](ancestors = 373733864,373733461,373733154,373640601)

source/UIData/Tools/Optimizer.cs
simeoncran: No need. The .Names property told us that there is a value with this name and type in there, so the Try will always succeed.
Note that the .Names property is something that we added - it's not part of the WinComp.CompositionPropertySet API. WinComp.CompositionPropertySet needs the "Try" because there's no way to ask it what properties are in it, so you just have to try to get a value out and see if it works. 

---
In reply to: [373665258](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373665258) [](ancestors = 373665258)

source/WinCompData/CompositionPropertySet.cs
```diff
@@ -32,80 +37,98 @@ internal CompositionPropertySet(CompositionObject owner)
...
+            if (existingPropertyType != type)
+            {
+                // The name is in the property set, but not for this type.
+                value = default(T);
+                return CompositionGetValueStatus.TypeMismatch;
+            }
+
+            if (!bag.TryGetValue(propertyName, out value))
+            {
+                throw new InvalidOperationException();
```
janny10: Shouldn't this never happen? Would this better as an assert?

source/WinCompData/CompositionPropertySet.cs
```diff
@@ -32,80 +37,98 @@ internal CompositionPropertySet(CompositionObject owner)
...
 
             // The name is in the property set - does it refer to the right type?
-            return bag.TryGetValue(propertyName, out value)
-                ? CompositionGetValueStatus.Succeeded
-                : CompositionGetValueStatus.TypeMismatch;
+            if (existingPropertyType != type)
+            {
+                // The name is in the property set, but not for this type.
+                value = default(T);
+                return CompositionGetValueStatus.TypeMismatch;
```
janny10: >CompositionGetValueStatus [](start = 23, length = 25)

Does this need to exist? Couldn't we throw here and remove this CompositionGetValueStatus type? Seems the other places TypeMismatch is checked, we throw. Happy to keep it, just thought it was worth commenting on.

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -786,6 +787,53 @@ static LoadedImageSurfaceInfo LoadedImageSurfaceInfoFromObjectData(ObjectData no
         static IEnumerable<ObjectData> OrderByName(IEnumerable<ObjectData> nodes) =>
             nodes.OrderBy(n => n.Name, AlphanumericStringComparer.Instance);
 
+        static string PropertySetValueType(PropertySetValueType value)
```
janny10: Could be just me, but it seems confusing to have the enum name and its de facto toString method have the same name. Maybe this could be PropertySetValueTypeName or PropertySetValueTypeToString. Would nameof work on this enum?

source/UIData/Tools/Optimizer.cs
janny10: Is it necessary for WinCompData CompositionPropertySets to have Try* methods? Perhaps I'm missing something but it looks like all the places they are called the value is guaranteed to exist.

---
In reply to: [373734712](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r373734712) [](ancestors = 373734712,373665258)

simeoncran: The API for CompositionPropertySet is not owned by us: [https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Composition.CompositionPropertySet](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Composition.CompositionPropertySet)
We try to make out the WinCompData APIs look the same as the Comp APIs so that code written for Comp will likely compile for WinCompData.

---
In reply to: [374246937](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r374246937) [](ancestors = 374246937,373734712,373665258)

source/WinCompData/CompositionPropertySet.cs
simeoncran: It should never happen. As to whether it should be an assert, that's a bit of a philosophical question that we should have and apply to the whole codebase.
You can think of our use of InvalidOperationException as "this will never happen". Going forward I want to replace all the InvalidOperationException usage with something more descriptive (like the Unreachable in the parser).

---
In reply to: [374231523](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r374231523) [](ancestors = 374231523)

source/WinCompData/CompositionPropertySet.cs
simeoncran: I believe this is so the behavior matches WinComp's behavior.

---
In reply to: [374233215](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r374233215) [](ancestors = 374233215)

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
simeoncran: I'm in 2 minds about that... yes, Nameof will work, but then we're taking a dependency on the name used by the enum.
Regarding the name of the type being the same as the name of thing that we're converting to string, that's a pattern through a lot of this codebase.

---
In reply to: [374236395](https://github.com/windows-toolkit/Lottie-Windows/pull/236#discussion_r374236395) [](ancestors = 374236395)

PR #237, opened 2020-01-31
Make our simple implementations of Equals more consistent across types.

simeoncran:


derobinsMSFT:
The `Equals` method definitions in `HoldEasing.cs` and `LinearEasing.cs` are currently inconsistent, even though the functionality is the same ("Is this object a nonnull [Hold|Linear]Easing?"). These should also be updated so that their implementations are consistent.

derobinsMSFT:
We can also simplify the `Equals` override in `Sequence.cs`:

`public override bool Equals(object obj) => Equals(obj as Sequence<T>);`

source/LottieData/Color.cs
```diff
@@ -47,7 +47,7 @@ public static Color FromArgb(double a, double r, double g, double b)
         public double B { get; }
 
         /// <inheritdoc/>
-        public override bool Equals(object obj) => obj is Color && Equals((Color)obj);
+        public override bool Equals(object obj) => obj is Color other && Equals(other);
 
         /// <inheritdoc/>
         public bool Equals(Color other) => A == other.A && R == other.R && G == other.G && B == other.B;
```
derobinsMSFT: >Equals [](start = 20, length = 6)

Should we check if `other` is null here? (`Color` is a reference type.)

derobinsMSFT: If so, we can also change the other `Equals` to:

`public override bool Equals(object obj) => Equals(obj as Color);`

---
In reply to: [373638380](https://github.com/windows-toolkit/Lottie-Windows/pull/237#discussion_r373638380) [](ancestors = 373638380)

source/WinCompData/Expressions/Expression.cs
```diff
@@ -87,7 +87,7 @@ protected static string Parenthesize(Expression expression)
 
         public bool Equals(Expression other) => !(other is null) && other.ToText() == ToText();
 
-        public override sealed bool Equals(object obj) => (obj is Expression other) && Equals(other);
+        public override sealed bool Equals(object obj) => obj is Expression other && Equals(other);
```
derobinsMSFT: >obj is Expression other [](start = 58, length = 23)

Necessary? `Equals(Expression other)` checks for null; maybe just call `Equals(obj as Expression)`.

source/WinCompData/Expressions/Expression.cs
```diff
@@ -87,7 +87,7 @@ protected static string Parenthesize(Expression expression)
 
         public bool Equals(Expression other) => !(other is null) && other.ToText() == ToText();
```
derobinsMSFT: >!(other is null) [](start = 48, length = 16)

`other is Expression`

derobinsMSFT:
`source\LottieToWinComp\LottieToWinCompTranslator.cs`, line 4430: Can probably simplify this override to something like:

```
public override bool Equals(object obj) =>
    obj is ScaleAndOffset other &&
    other._scale == _scale &&
    other._offset == _offset;
```

source/LottieData/Color.cs
derobinsMSFT: Still need a null check in the strongly-typed `Equals`.

---
In reply to: [373639449](https://github.com/windows-toolkit/Lottie-Windows/pull/237#discussion_r373639449) [](ancestors = 373639449,373638380)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4457,15 +4457,9 @@ internal ScaleAndOffset(double scale, double offset)
...
-            {
-                var other = obj as ScaleAndOffset;
-                if (other == null)
-                {
-                    return false;
-                }
-
-                return other._scale == _scale && other._offset == _offset;
-            }
+                => obj is ScaleAndOffset other &&
```
derobinsMSFT: >=> [](start = 16, length = 2)

Previous line (consistent with the rest of the file).

source/LottieData/Sequence.cs
```diff
@@ -43,11 +43,7 @@ public Sequence(IEnumerable<T> items)
             Enumerable.SequenceEqual(_items, other._items);
```
derobinsMSFT: `other is Sequence<T> && Enumerable.SequenceEqual(_items, other._items);`

source/LottieData/Color.cs
derobinsMSFT: `other is Color`

---
In reply to: [373744759](https://github.com/windows-toolkit/Lottie-Windows/pull/237#discussion_r373744759) [](ancestors = 373744759,373639449,373638380)

PR #239, opened 2020-02-04
Begin parser refactor to allow switching to System.Text.Json.

simeoncran:
Break up the huge LottieCompositionReader into separate files so make it easier to reason about. This is being done now so that we can get ready to switch to System.Text.Json.
This also permanently turns on checking for unparsed fields seeing as we were always doing that anyway.

Apart from some renaming, the parser should work exactly as it did before. Subsequent commits will attempt to improve the code and get it better aligned with what System.Text.Json will need, before we finally switch over to System.Text.Json.


derobinsMSFT:
> #pragma warning disable SA1205 // Partial elements should declare access

Periods.

---
Refers to: source/LottieReader/Serialization/LottieCompositionReader.cs:23 in a0cee1f. [](commit_id = a0cee1fae502f246a4f3283684ddf5059fb6fa84, deletion_comment = False)

derobinsMSFT:
> #pragma warning disable SA1205 // Partial elements should declare access

Same for all other files with these directives/comments.

---
In reply to: [582052510](https://github.com/windows-toolkit/Lottie-Windows/pull/239#issuecomment-582052510) [](ancestors = 582052510)

---
Refers to: source/LottieReader/Serialization/LottieCompositionReader.cs:23 in a0cee1f. [](commit_id = a0cee1fae502f246a4f3283684ddf5059fb6fa84, deletion_comment = False)

simeoncran:
> #pragma warning disable SA1205 // Partial elements should declare access

Please ignore these. They're inserted automatically by VS. Ours shouldn't look any different from anyone else's. 

---
In reply to: [582053550](https://github.com/windows-toolkit/Lottie-Windows/pull/239#issuecomment-582053550) [](ancestors = 582053550,582052510)

---
Refers to: source/LottieReader/Serialization/LottieCompositionReader.cs:23 in a0cee1f. [](commit_id = a0cee1fae502f246a4f3283684ddf5059fb6fa84, deletion_comment = False)

PR #240, opened 2020-02-04
Finish plumbing of property binding info to codegen.

simeoncran:
Gets the exposed type from the translator to the generator. This is necessary for the next stage: actually generating properties for the propertyset values.


source/UIDataCodeGen/CodeGen/SourceMetadata.cs
```diff
@@ -0,0 +1,79 @@
...
+        Lottie _lottieMetadata;
+
+        internal SourceMetadata(IReadOnlyDictionary<Guid, object> sourceMetadata)
+        {
+            _sourceMetadata = sourceMetadata;
+        }
+
+        internal Lottie LottieMetadata => _lottieMetadata ?? (_lottieMetadata = new Lottie(this));
+
+        internal IReadOnlyList<(string bindingName, PropertySetValueType actualType, PropertySetValueType exposedType)> ProperytBindings
```
derobinsMSFT: >Properyt [](start = 120, length = 8)

`PropertyBindings`?

source/LottieToWinComp/PropertyBindings.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    sealed class PropertyBindings
+    {
+        // Identifies the bound property names in TranslationResult.SourceMetadata.
+        static readonly Guid s_propertyBindingNamesKey = new Guid("A115C46A-254C-43E6-A3C7-9DE516C3C3C8");
+
+        readonly List<(string bindingName, PropertySetValueType actualType, PropertySetValueType exposedType)> _names =
+            new List<(string bindingName, PropertySetValueType actualType, PropertySetValueType exposedType)>();
+
+        // Adds the current list of property bindings to the source metatadata dictionary.
```
derobinsMSFT: >metata [](start = 68, length = 6)

"meta"

source/UIDataCodeGen/CodeGen/SourceMetadata.cs
```diff
@@ -0,0 +1,82 @@
...
+                        var endFrame = startFrame + durationInFrames;
+                        var endProgress = Clamp01(endFrame / DurationInFrames);
+                        var endTime = TimeSpan.FromSeconds(endFrame / FramesPerSecond);
+                        yield return (m.Name, (startTime, startProgress), (endTime, endProgress));
+                    }
+                }
+            }
+        }
+
+        static double Clamp01(double value)
```
derobinsMSFT: >Clamp01 [](start = 22, length = 7)

Could this method have a more descriptive name? In addition, maybe this should be part of the `Lottie` class; it appears to only be used in the `Lottie.Markers` getter.

simeoncran: Ah, yes, it could be moved inside the Lottie class. What do you suggest as a more descriptive name?


---
In reply to: [374982591](https://github.com/windows-toolkit/Lottie-Windows/pull/240#discussion_r374982591) [](ancestors = 374982591)

derobinsMSFT: Ah, I see why it's called `Clamp01` now; the name looked generated at first.

How about something like `EnsureProgressLimitBetweenZeroAndOne`?

---
In reply to: [374983835](https://github.com/windows-toolkit/Lottie-Windows/pull/240#discussion_r374983835) [](ancestors = 374983835,374982591)

simeoncran: Will address in next iteration.

---
In reply to: [374987987](https://github.com/windows-toolkit/Lottie-Windows/pull/240#discussion_r374987987) [](ancestors = 374987987,374983835,374982591)

source/LottieToWinComp/PropertyBindings.cs
simeoncran: Will address in next iteration.

---
In reply to: [374975468](https://github.com/windows-toolkit/Lottie-Windows/pull/240#discussion_r374975468) [](ancestors = 374975468)

PR #242, opened 2020-02-08
Add "Reader" struct to ease transition to System.Text.Json.

simeoncran:
The Reader is a ref struct, just like System.Text.Json.Utf8JsonReader, but under the covers still uses the Newtonsoft JsonReader. This gets us another step closer to switching over to System.Text.Json.


source/LottieReader/Serialization/Exceptions.cs
```diff
@@ -0,0 +1,14 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
+{
+    static class Exceptions
+    {
+        // The code we hit is supposed to be unreachable. This indicates a bug.
+        internal static Exception Unreachable => new InvalidOperationException("Unreachable code executed");
```
derobinsMSFT: >executed" [](start = 97, length = 9)

Nit: should error/exception messages have periods?

derobinsMSFT: Looks like this had a `.` originally (`JsonToGenericData.cs`).

---
In reply to: [377199848](https://github.com/windows-toolkit/Lottie-Windows/pull/242#discussion_r377199848) [](ancestors = 377199848)

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -100,15 +101,15 @@ out IReadOnlyList<(string Code, string Description)> issues)
         }
 
         static LottieComposition ReadLottieCompositionFromJson(
-            JsonReader jsonReader,
+            ref Reader jsonReader,
```
derobinsMSFT: >jsonReader [](start = 23, length = 10)

Can this be renamed to `reader` to be consistent with the other `Reader` objects in the surrounding methods? Consequently, the `LottieCompositionReader` below will need a new name-- naming it something besides `reader` would help better distinguish that it's not an object of the newer `Reader` type.

simeoncran: As discussed... the name is good for the new Utf8JsonReader, it just doesn't seem right for the interim Reader struct, but that struct only exists for staging.

---
In reply to: [377225348](https://github.com/windows-toolkit/Lottie-Windows/pull/242#discussion_r377225348) [](ancestors = 377225348)

source/LottieReader/Serialization/EmptyList.cs
```diff
@@ -0,0 +1,55 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
+{
+    // A list that never contains any items.
+    sealed class EmptyList<T> : IList<T>
```
derobinsMSFT: >EmptyList [](start = 17, length = 9)

Looking at the `IList` documentation, I notice that you have defined a few properties for `EmptyList` but omitted definitions for others (e.g., `IsFixedSize`). Is there a particular reason for this?

simeoncran: If I was omitting them then the compiler would complain. I suspect you're looking at the wrong IList. See: [https://docs.microsoft.com/en-us/dotnet/api/system.collections.ilist?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ilist?view=netframework-4.8) - that's the (old) non-generic IList. This one is the generic one.

---
In reply to: [377362369](https://github.com/windows-toolkit/Lottie-Windows/pull/242#discussion_r377362369) [](ancestors = 377362369)

source/LottieReader/Serialization/EmptyList.cs
```diff
@@ -0,0 +1,55 @@
...
+        IEnumerator IEnumerable.GetEnumerator() => Array.Empty<T>().GetEnumerator();
+
+        int IList<T>.IndexOf(T item) => -1;
+
+        void IList<T>.Insert(int index, T item)
+        {
+            throw new IndexOutOfRangeException();
+        }
+
+        bool ICollection<T>.Remove(T item) => false;
```
derobinsMSFT: >false [](start = 46, length = 5)

Nit: according to the [official documentation]([https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1.remove?view=netframework-4.8#System_Collections_Generic_ICollection_1_Remove__0](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1.remove?view=netframework-4.8#System_Collections_Generic_ICollection_1_Remove__0)_), `Remove` should return a `NotSupportedException` on read-only items.

source/LottieReader/Serialization/EmptyList.cs
derobinsMSFT: Sorry, my mistake! Yes, was looking at the non-generic documentation.

---
In reply to: [377369171](https://github.com/windows-toolkit/Lottie-Windows/pull/242#discussion_r377369171) [](ancestors = 377369171,377362369)

PR #243, opened 2020-02-08
More complete support for fills and strokes in the YAML serializer.

simeoncran:


source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -291,17 +291,13 @@ YamlObject FromShapeLayerContent(ShapeLayerContent content, YamlMap superclassCo
                 case ShapeContentType.Group:
                     return FromShapeGroup((ShapeGroup)content, superclassContent);
                 case ShapeContentType.SolidColorStroke:
```
derobinsMSFT: >case ShapeContentType.SolidColorStroke: [](start = 16, length = 39)

Move down a couple lines (to match the [alphabetical] ordering below).

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -346,44 +379,54 @@ YamlObject FromSolidColorStroke(SolidColorStroke content, YamlMap superclassCont
...
             return result;
         }
 
         YamlObject FromLinearGradientStroke(LinearGradientStroke content, YamlMap superclassContent)
+        {
+            var result = FromIGradient(content, superclassContent);
+            return result;
+        }
+
+        YamlMap FromIGradient(IGradient content, YamlMap superclassContent)
```
derobinsMSFT: >YamlMap [](start = 8, length = 7)

Just about all the other `From*` methods here are returning `YamlObject` (same for `FromIRadialGradient`.)

simeoncran: This needs to return a YamlMap because it is being used as a map by the callers.
Alternatively we could introduce a new pattern for interfaces that simply populates the given map and returns void. Let me try that...

---
In reply to: [377294571](https://github.com/windows-toolkit/Lottie-Windows/pull/243#discussion_r377294571) [](ancestors = 377294571)

source/LottieData/Serialization/LottieCompositionYamlSerializer.cs
```diff
@@ -50,6 +50,19 @@ YamlMap GetLottieObjectContent(LottieObject obj)
...
         }
 
+        void AddFromIGradient(IGradient content, YamlMap result)
+        {
+            result.Add(nameof(content.StartPoint), FromAnimatable(content.StartPoint));
+            result.Add(nameof(content.EndPoint), FromAnimatable(content.EndPoint));
+            result.Add(nameof(content.GradientStops), FromAnimatable(content.GradientStops, p => FromSequence(p, FromGradientStop)));
+        }
+
+        void AddFromIRadialGradient(IRadialGradient content, YamlMap result)
```
derobinsMSFT: >AddFromIRadialGradient [](start = 13, length = 22)

No longer a call to `AddFromIGradient`.

simeoncran: oops!

---
In reply to: [377342341](https://github.com/windows-toolkit/Lottie-Windows/pull/243#discussion_r377342341) [](ancestors = 377342341)

Issue #253, opened 2020-05-06
NuGet package description is not descriptive


chingucoding:
The package description for Microsoft.Toolkit.Uwp.UI.Lottie is just "Package Description". Instead this package should have a meaningful package description. See screenshot below:

![unknown](https://user-images.githubusercontent.com/16122379/81154387-222c3900-8f84-11ea-92d2-8e6fe2ccf444.png)

jamesmcroft:
Should be a simple fix too, just some missing csproj metadata that needs adding in here - https://github.com/windows-toolkit/Lottie-Windows/blob/master/Lottie-Windows/Lottie-Windows.csproj

simeoncran:
This won't show up until the 6.1 release (soon!)

PR #254, opened 2020-05-06
Added Lottie-Windows package description and tags

jamesmcroft:
## Summary

Changes have been made to the Lottie Windows csproj to include additional metadata for the generated NuGet package including a description (taken from the [README](https://github.com/windows-toolkit/Lottie-Windows/tree/master/Lottie-Windows)) and package tags.

This change is related to #253

### DIFF
--- a/Lottie-Windows/Lottie-Windows.csproj
+++ b/Lottie-Windows/Lottie-Windows.csproj
@@ -6,8 +6,12 @@
     <OutputType>winmdobj</OutputType>
     <IsPackable>True</IsPackable>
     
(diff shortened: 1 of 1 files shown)


jamesmcroft:
@simeoncran The description and tags here are used by NuGet.org. The description is displayed when a user searches or views the package on the site. The tags are used to describe the package and help it's discoverability in search and filtering on the site. 

Once built, the package metadata will include the description and tags fields as below:
NuGet package metadata

At the moment, when you search for the Lottie package in NuGet or through VS, you get this:
Lottie NuGet package

The new metadata will let users see something more like this:
Microsoft Toolkit package example description

The tags will also be displayed within the package description page like this:
Microsoft Toolkit package example tags

simeoncran:
Fantastic! We should do this for LottieGen too, but feel free to leave that to me. Good change! Thanks!

PR #265, opened 2020-05-22
Improve handling of grouped animated paths.

simeoncran:
This change will allow some (not very common) cases of grouped animated paths where the animation can be optimized away.
In the cases it can't handle, it now raises an issue to make it clear that there is a problem.

### DIFF
--- a/source/LottieData/Path.cs
+++ b/source/LottieData/Path.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
 
         /// <summary>
         /// Returns a path with the same properties except with the given
-        /// <paramref name="geometryData"/> in palse of <see cref="Data"/>.
+        /// <paramref name="geometryData"/> in place of <see cref="Data"/>.
         /// </summary>
         /// <param name="geometryData">The geometry to use in place of <see cref="Data"/>.</param>
         /// <returns>The cloned path.</returns>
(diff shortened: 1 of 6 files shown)


source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4297,6 +4300,27 @@ float GetOutPointProgress(TranslationContext context)
...
 
+        // Creates a CompositionPath from a group of paths.
+        CompositionPath CompositionPathFromPathGeometryGroup(
+            IEnumerable<Sequence<BezierSegment>> paths,
+            ShapeFill.PathFillType fillType,
+            bool optimizeLines)
+        {
+            var compositionPaths = paths.Select(p => CompositionPathFromPathGeometry(p, fillType, optimizeLines)).ToArray();
+
+            CompositionPath result;
```
derobinsMSFT, line 4311: >CompositionPath result; [](start = 12, length = 23)

No need for this variable. If you feel strongly about keeping it, at least combine the declaration and initialization.

source/LottieData/Path.cs
```diff
@@ -28,5 +28,22 @@ sealed class Path : ShapeLayerContent
 
         /// <inheritdoc/>
         public override LottieObjectType ObjectType => LottieObjectType.Shape;
+
+        /// <summary>
+        /// Returns a path with the same properties except with the given
+        /// <paramref name="geometryData"/> in palse of <see cref="Data"/>.
```
eliezerpMS, line 34: >palse [](start = 47, length = 5)

palse?

simeoncran: Time for glasses. :sunglasses: 

---
In reply to: [429382003](https://github.com/windows-toolkit/Lottie-Windows/pull/265#discussion_r429382003) [](ancestors = 429382003)

PR #268, opened 2020-05-26
We were ignoring the "closed" property on paths.

simeoncran:
In most cases, a closed path is simply one that ends where it starts, so we just left all our paths open and relied on After Effects to complete the path.
But that was wrong in some cases - specifically the closing of the path determines whether a miter or an end cap is used where the end meets the start.

### DIFF
--- a/source/LottieData/Optimization/Optimizer.cs
+++ b/source/LottieData/Optimization/Optimizer.cs
@@ -106,8 +106,8 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
             return new Animatable<PathGeometry>(hacked[0].Value, hacked, optimized.PropertyIndex);
         }
 
-        // Returns a KeyFrame<PathGeometry> that contains only the first bezier segment of the given
-        // KeyFrame<PathGeometry>
+        // Returns a KeyFrame<PathGeometry> that contains only the first Bezier segment of the given
+        // KeyFrame<PathGeometry>.
         static KeyFrame<PathGeometry> HackPathGeometry(KeyFrame<PathGeometry> value)
             => new KeyFrame<PathGeometry>(
                 value.Frame,
--- a/source/LottieData/Optimization/Optimizer.cs
+++ b/source/LottieData/Optimization/Optimizer.cs
@@ -21,22 +21,22 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
 #endif
     sealed class Optimizer
     {
-        static readonly AnimatableComparer<Sequence<BezierSegment>> PathGeometryComparer = new AnimatableComparer<Sequence<BezierSegment>>();
-        readonly Dictionary<Animatable<Sequence<BezierSegment>>, Animatable<Sequence<BezierSegment>>> _animatablePathGeometriesCache;
+        static readonly AnimatableComparer<PathGeometry> PathGeometryComparer = new AnimatableComparer<PathGeometry>();
+        readonly Dictionary<Animatable<PathGeometry>, Animatable<PathGeometry>> _animatablePathGeometriesCache;
 
         public Optimizer()
         {
-            _animatablePathGeometriesCache = new Dictionary<Animatable<Sequence<BezierSegment>>, Animatable<Sequence<BezierSegment>>>(PathGeometryComparer);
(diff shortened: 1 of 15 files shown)


source/LottieData/Optimization/Optimizer.cs
```diff
@@ -101,13 +103,18 @@ public Animatable<Sequence<BezierSegment>> GetOptimized(Animatable<Sequence<Bezi
...
             var hacked = optimized.KeyFrames.SelectToSpan(pg => HackPathGeometry(pg));
-            return new Animatable<Sequence<BezierSegment>>(hacked[0].Value, hacked, optimized.PropertyIndex);
+            return new Animatable<PathGeometry>(hacked[0].Value, hacked, optimized.PropertyIndex);
         }
 
-        static KeyFrame<Sequence<BezierSegment>> HackPathGeometry(KeyFrame<Sequence<BezierSegment>> value)
-        {
-            return new KeyFrame<Sequence<BezierSegment>>(value.Frame, new Sequence<BezierSegment>(new[] { value.Value.Items[0] }), Vector3.Zero, Vector3.Zero, value.Easing);
-        }
+        // Returns a KeyFrame<PathGeometry> that contains only the first bezier segment of the given
```
derobinsMSFT, line 109: >b [](start = 73, length = 1)

Capitalize.

source/LottieData/Optimization/Optimizer.cs
```diff
@@ -101,13 +103,18 @@ public Animatable<Sequence<BezierSegment>> GetOptimized(Animatable<Sequence<Bezi
...
+        // KeyFrame<PathGeometry>
```
derobinsMSFT, line 110: >> [](start = 32, length = 1)

Period.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2055,10 +2055,12 @@ CanvasGeometry CreateWin2dPathGeometry(Path path, ShapeFill.PathFillType fillTyp
                         }
                     }
 
-                    // Leave the figure open. If Lottie wanted it closed it will have defined
-                    // a final bezier segment back to the start.
-                    // Closed simply tells D2D to synthesize a final segment.
-                    builder.EndFigure(CanvasFigureLoop.Open);
+                    // Closed tells D2D to synthesize a final segment. In many cases Closed
+                    // will have no effect because After Effects will have included the final
+                    // segment however it can make a difference, for example if the figure is
```
derobinsMSFT, line 2060: >segment [](start = 23, length = 7)

Nit: semicolon or period.

source/LottieData/PathGeometry.cs
```diff
@@ -0,0 +1,44 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
+{
+    /// <summary>
+    /// A sequence of <see cref="BezierSegment"/> that describes the shape of a path.
```
derobinsMSFT, line 11: >> [](start = 48, length = 1)

"s"

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: You need to explain to me why. This is similar to one we were discussing previously but I don't understand the rule you are applying and why it matters.

---
In reply to: [430660286](https://github.com/windows-toolkit/Lottie-Windows/pull/268#discussion_r430660286) [](ancestors = 430660286)

derobinsMSFT: In this case, the sentence strikes me as a run-on/a tad long. Here would be a good spot to split it.

---
In reply to: [430694315](https://github.com/windows-toolkit/Lottie-Windows/pull/268#discussion_r430694315) [](ancestors = 430694315,430660286)

simeoncran: Got it. I'll simplify it a bit.

---
In reply to: [430700091](https://github.com/windows-toolkit/Lottie-Windows/pull/268#discussion_r430700091) [](ancestors = 430700091,430694315,430660286)

PR #270, opened 2020-05-27
PathGeometry didn't implement GetHashCode so it wasn't being canonicalized.

simeoncran:
Also, fixing a bug in the NodeNamer that caused it to add leading 0's to object counters unnecessarily when there were exactly 10/100/1000/etc objects.

### DIFF
--- a/source/UIDataCodeGen/CodeGen/NodeNamer.cs
+++ b/source/UIDataCodeGen/CodeGen/NodeNamer.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen
                 else
                 {
                     // Use only as many digits as necessary to express the largest count.
-                    var digitsRequired = (int)Math.Ceiling(Math.Log10(nodeList.Count + 1));
+                    var digitsRequired = (int)Math.Ceiling(Math.Log10(nodeList.Count));
                     var counterFormat = new string('0', digitsRequired);
 
                     for (var i = 0; i < nodeList.Count; i++)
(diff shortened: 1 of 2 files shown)


source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -77,7 +77,7 @@ public static IEnumerable<(TNode, string)> GenerateNodeNames(IEnumerable<TNode>
                 else
                 {
                     // Use only as many digits as necessary to express the largest count.
-                    var digitsRequired = (int)Math.Ceiling(Math.Log10(nodeList.Count + 1));
+                    var digitsRequired = (int)Math.Ceiling(Math.Log10(nodeList.Count));
```
derobinsMSFT, line 80: >nodeList.Count [](start = 70, length = 14)

If this is 0, there will be -Infinity digitsRequired. Not sure if that would cause any issues here.

simeoncran: Can't happen. nodesByName contains only non-empty lists. See line 55.

---
In reply to: [431440979](https://github.com/windows-toolkit/Lottie-Windows/pull/270#discussion_r431440979) [](ancestors = 431440979)

derobinsMSFT: Thanks for pointing that out.

---
In reply to: [431450689](https://github.com/windows-toolkit/Lottie-Windows/pull/270#discussion_r431450689) [](ancestors = 431450689,431440979)

PR #272, opened 2020-05-28
Add color and improved formatting to the output of LottieGen.

simeoncran:
This makes the output easier for humans to parse.

### DIFF
--- a/LottieGen/InfoType.cs
+++ b/LottieGen/InfoType.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+/// <summary>
+/// Categories of information that are displayed to the user on the console. These
+/// categories are used to color the output.
+/// </summary>
+enum InfoType
+{
+    Default = 0,
+
+    /// <summary>
+    /// Advice to the user about a follow up action they should take.
+    /// </summary>
+    Advice,
+
+    /// <summary>
+    /// A file path or file name.
+    /// </summary>
+    FilePath,
+
+    /// <summary>
+    /// An issue that may prevent the result from working as expected.
+    /// </summary>
+    Issue,
+
+    /// <summary>
+    /// An announcement about the tool name and version.
+    /// </summary>
+    Signon,
+}
--- a/LottieGen/Reporter.cs
+++ b/LottieGen/Reporter.cs
@@ -31,7 +31,8 @@ sealed class Reporter
         WriteError(errorMessage, ConsoleColor.Red, ConsoleColor.Black);
 
     // Helper for writing info lines to the info stream.
-    internal void WriteInfo(string infoMessage) => WriteInfo(InfoType.Default, infoMessage);
+    internal void WriteInfo(string infoMessage) =>
+        WriteInfo(InfoType.Default, infoMessage);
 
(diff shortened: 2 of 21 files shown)


LottieGen/Reporter.cs
```diff
@@ -28,15 +27,39 @@ internal Reporter(TextWriter infoStream, TextWriter errorStream)
...
+
+    // Helper for writing info lines to the info stream.
+    internal void WriteInfo(InfoType type, string infoMessage)
     {
-        InfoStream.WriteLine(infoMessage);
+        ConsoleColor foreground, background = ConsoleColor.Black;
+
+        switch (type)
+        {
+            case InfoType.Default:
```
derobinsMSFT, line 43: >InfoType.Default: [](start = 17, length = 17)

Wondering if the Default and Signon colors should be switched-- cf. line 113.

simeoncran: I think 113 is where the problem is - we shouldn't be doing defaults that way - they should always be explicit via an InfoType.

---
In reply to: [432174922](https://github.com/windows-toolkit/Lottie-Windows/pull/272#discussion_r432174922) [](ancestors = 432174922)

LottieGen/InfoType.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+/// <summary>
+/// Categories of information that are displayed to the user on the console. These
+/// categories are used to color the output.
+/// </summary>
+enum InfoType
```
derobinsMSFT, line 9: >InfoType [](start = 5, length = 8)

Based on how this enum is named/described, it seems strange that Error isn't included as a category, since errors are also pieces of info that can be displayed on the console. That said, Errors are handled a bit differently by the `Reporter` than other types of info (e.g., their `TextWriter`, and thus output stream, can be different from that of the other info types).

It would be good to add a comment explaining why Error isn't included here/that errors are a special case, and perhaps also update the enum name to reflect this (e.g., `InfoStreamType` or `NonErrorInfoType`).

simeoncran: I agree that it's non-intuitive, but this enum is just categories of information that we handle, it's not meant to document how the reporter is to be used.
Also I'm trying to keep the name terse so that it doesn't clutter up the callsites so much.

---
In reply to: [432560252](https://github.com/windows-toolkit/Lottie-Windows/pull/272#discussion_r432560252) [](ancestors = 432560252)

LottieGen/Reporter.cs
```diff
@@ -28,15 +27,39 @@ internal Reporter(TextWriter infoStream, TextWriter errorStream)
...
-    internal void WriteError(string errorMessage)
-    {
-        ErrorStream.WriteLine($"Error: {errorMessage}");
-    }
+    internal void WriteError(string errorMessage) =>
+        WriteError(errorMessage, ConsoleColor.Red, ConsoleColor.Black);
 
     // Helper for writing info lines to the info stream.
-    internal void WriteInfo(string infoMessage)
+    internal void WriteInfo(string infoMessage) => WriteInfo(InfoType.Default, infoMessage);
```
derobinsMSFT, line 34: >WriteInfo(InfoType.Default, infoMessage); [](start = 51, length = 41)

Nit: move to next line (consistent with `WriteError`).

PR #276, opened 2020-06-01
Support reading of drawing direction.

simeoncran:
Drawing direction determines where the path of the shape starts and ends. This matters when trimming the path. I think that if the shape is closed (e.g. an ellipse) and the drawing is in reverse
then the trim start and trim end get swapped. Still trying to figure out all the details.
There will be at least one more change to support drawing direction. This is just to get the
bones of the parsing in place. Following changes will use it during translation.

Note that the enum values don't have descriptive names because I'm not sure what they should be yet.

### DIFF
--- a/source/LottieData/LottieData.projitems
+++ b/source/LottieData/LottieData.projitems
@@ -6,8 +6,6 @@
     <SharedGUID>b3db16ee-a821-4474-a188-e64926529bbd</SharedGUID>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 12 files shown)


source/LottieData/LottieData.projitems
```diff
@@ -6,6 +6,8 @@
     <SharedGUID>b3db16ee-a821-4474-a188-e64926529bbd</SharedGUID>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)DrawingDirection.cs" />
```
simeoncran, line 9: Reorder

PR #277, opened 2020-06-01
Make writes to the console transactional where necessary.

simeoncran:
The recent output coloring made the problem obvious - when we run Lottiegen over multiple files there are numerous threads that are writing to the console and can interleave their writes.
With this change, the writes that need to be kept together have a way of being kept together.

### DIFF
--- a/LottieGen/LottieFileProcessor.cs
+++ b/LottieGen/LottieFileProcessor.cs
@@ -362,8 +362,14 @@ sealed class LottieFileProcessor
             outputFilePath,
             LottieCompositionXmlSerializer.ToXml(_lottieComposition).ToString());
 
-        _reporter.WriteInfo("Lottie XML written to:");
-        _reporter.WriteInfo(InfoType.FilePath, $" {outputFilePath}");
+        if (result)
+        {
+            using (_reporter.InfoStream.Lock())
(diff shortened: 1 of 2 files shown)


derobinsMSFT:
>         InfoStream.WriteLine(infoMessage);

There's a lock in `Color`, and a lock in `WriteLine`, but-- particularly for the `WriteInfo` calls not protected by a `LockGuard`-- what's stopping another thread from calling `Color` and changing the colors after this thread calls `Color` but before it calls `WriteLine`? 

---
Refers to: LottieGen/Reporter.cs:119 in 92e97bc. [](commit_id = 92e97bca6d1967cd2d0272fde24c9ebb9833db59, deletion_comment = False)

simeoncran:
>         InfoStream.WriteLine(infoMessage);

Excellent point, that would be bad.

---
In reply to: [637753945](https://github.com/windows-toolkit/Lottie-Windows/pull/277#issuecomment-637753945) [](ancestors = 637753945)

---
Refers to: LottieGen/Reporter.cs:119 in 92e97bc. [](commit_id = 92e97bca6d1967cd2d0272fde24c9ebb9833db59, deletion_comment = False)

PR #279, opened 2020-06-03
Support reverse drawing direction.

simeoncran:


source/LottieData/DrawingDirection.cs
```diff
@@ -5,16 +5,26 @@
...
+        /// The path is to be drawn in the forward direction. For rectangles and
+        /// ellipses the forward direction is clockwise.
+        /// </summary>
+        Forward = 1,
+
+        /// <summary>
+        /// The path is to be drawn in the reverse direction. For rectangles and
+        /// ellipses the reverse direction is counter-clockwise.
+        /// </summary>
+        Reverse = 3,
```
derobinsMSFT, line 28: >3 [](start = 18, length = 1)

2?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -1814,7 +1814,7 @@ CompositionShape TranslateMergePathsContent(TranslationContext context, ShapeCon
             {
                 var result = _c.CreateSpriteShape();
                 result.Geometry = _c.CreatePathGeometry(new CompositionPath(mergedGeometry));
-                TranslateAndApplyShapeContentContext(context, shapeContext, result);
+                TranslateAndApplyShapeContentContext(context, shapeContext, result, reverseDirection: false, trimOffsetDegrees: 0);
```
derobinsMSFT, line 1817: >reverseDirection: [](start = 84, length = 17)

The calls to this function in this file are inconsistent about whether the explicitly name this parameter.

source/LottieData/DrawingDirection.cs
simeoncran: Ah, that shouldn't be there. The values are what After Effects uses, but there's no need to leak that info here. The only value that matters is the Unknown value.

---
In reply to: [435358833](https://github.com/windows-toolkit/Lottie-Windows/pull/279#discussion_r435358833) [](ancestors = 435358833)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: The goal is not for that to be consistent, just readable. When the value is just "false" then it's impossible to know from the context what it means. But if you are passing a variable called "reverseDirection" or some such thing then it's clear.

---
In reply to: [435368983](https://github.com/windows-toolkit/Lottie-Windows/pull/279#discussion_r435368983) [](ancestors = 435368983)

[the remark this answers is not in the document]
simeoncran: Why? Not saying you're wrong, I just don't understand why you are suggesting it.

---
In reply to: [435370325](https://github.com/windows-toolkit/Lottie-Windows/pull/279#discussion_r435370325) [](ancestors = 435370325)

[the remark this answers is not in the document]
derobinsMSFT: Ah, sorry-- consistency with the other calls to this function.

---
In reply to: [435607181](https://github.com/windows-toolkit/Lottie-Windows/pull/279#discussion_r435607181) [](ancestors = 435607181,435370325)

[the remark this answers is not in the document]
simeoncran: Not sure that I agree that callsites should have consistent shapes. My goal here was to keep it compact so as not to draw attention to the parameters which aren't very interesting in this case. But I don't feel that strongly, so I'll try it.

---
In reply to: [435619269](https://github.com/windows-toolkit/Lottie-Windows/pull/279#discussion_r435619269) [](ancestors = 435619269,435607181,435370325)

PR #284, opened 2020-06-08
Multi dimensional easing

simeoncran:
We were doing cubic bezier incorrectly. After Effects allows different cubic Bezier easing values for each component of a multi-component value. For example a Vector2 has 2 components, x and y, and their values can be eased at different rates.

We were also being too aggressive about optimizing away keyframes. Previously we said that if a keyframe had the same value as the previous keyframe, then it was redundant. However that's wrong in some cases where the easing between the keyframes is cubic Bezier.

Note that so far this change fixes the optimization, and the parsing, but the translation works the way it did before - the fix to have separate easings for component in the translator will come after this.


source/LottieData/LottieData.projitems
```diff
@@ -6,6 +6,7 @@
     <SharedGUID>b3db16ee-a821-4474-a188-e64926529bbd</SharedGUID>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)CubicBezier.cs" />
```
simeoncran, line 9: Reorder

source/LottieData/CubicBezier.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// The 3rd control point.
+        /// </summary>
+        public Vector2 ControlPoint2 { get; }
+
+        /// <summary>
+        /// True if the <see cref="CubicBezier"/> represents a linear function from
+        /// 0,0 to 1,1. A linear function requires that <see cref="ControlPoint1"/>
+        /// and <see cref="ControlPoint2"/> line on the line that passes through 0,0, and 1,1.
```
derobinsMSFT, line 37: >line [](start = 44, length = 4)

"lie"?

source/LottieData/LottieData.projitems
```diff
@@ -6,82 +6,83 @@
...
-    <Compile Include="$(MSBuildThisFileDirectory)\TextLayer.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\Tools\Stats.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\Transform.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\Trim.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\TrimPath.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\ValidationIssues.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\Vector2.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\Vector3.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Animatable.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)ExtensionMethods.cs" />
```
derobinsMSFT, line 10: >)ExtensionMethods.cs [](start = 48, length = 20)

Reorder?

source/LottieData/Optimization/Optimizer.cs
```diff
@@ -111,6 +111,14 @@ public Animatable<PathGeometry> GetOptimized(Animatable<PathGeometry> value)
         static KeyFrame<PathGeometry> HackPathGeometry(KeyFrame<PathGeometry> value) =>
             value.CloneWithNewValue(new PathGeometry(new Sequence<BezierSegment>(new[] { value.Value.BezierSegments.Items[0] }), isClosed: false));
 
+        static bool HasNonLinearCubicBezierEasing<T>(KeyFrame<T> keyFrame)
+            where T : IEquatable<T>
+        {
+            var easing = keyFrame.Easing;
+
+            return easing.Type == Easing.EasingType.CubicBezier && !((CubicBezierEasing)easing).Beziers[0].IsLinear;
```
derobinsMSFT, line 119: >Beziers[0] [](start = 96, length = 10)

Can this ever be empty?

source/LottieData/Optimization/Optimizer.cs
```diff
@@ -381,8 +406,10 @@ static IEnumerable<KeyFrame<T>> RemoveRedundantKeyFrames<T>(KeyFrame<T>[] keyFra
             for (var i = 0; i < keyFrames.Length - 1; i++)
             {
                 // Only include the key frame if it has a frame value that is different
-                // from the next key frame's frame.
-                if (keyFrames[i].Frame != keyFrames[i + 1].Frame)
+                // from the next key frame's frame or if it has a non-linear cubic Bezier easing
```
derobinsMSFT, line 409: >it [](start = 57, length = 2)

If the current frame has a non-linear cubic Bezier, or the next frame? The code suggests the next frame, but the wording here implies the current frame.

source/LottieData/Optimization/Optimizer.cs
simeoncran: No, a CubicBezierEasing is defined as an easing with one or more Beziers.

---
In reply to: [439045606](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439045606) [](ancestors = 439045606)

source/LottieData/CubicBezier.cs
simeoncran: That's an off by n error.

---
In reply to: [439040571](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439040571) [](ancestors = 439040571)

source/LottieData/Optimization/Optimizer.cs
simeoncran: Looking at this closer I believe they're both wrong. This is designed to get rid of keyframes that are at the same offset. It never makes sense to have 2 key frames at the same offset. I think what was there originally was correct.

---
In reply to: [439052463](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439052463) [](ancestors = 439052463)

source/LottieData/CubicBezier.cs
```diff
@@ -0,0 +1,60 @@
...
+                ControlPoint1.Equals(other.ControlPoint1) &&
+                ControlPoint2.Equals(other.ControlPoint2);
+
+        /// <inheritdoc/>
+        public override int GetHashCode() => ControlPoint1.GetHashCode() ^ ControlPoint2.GetHashCode();
+
+        public static bool operator ==(CubicBezier a, CubicBezier b) =>
+            a.ControlPoint1 == b.ControlPoint1 && a.ControlPoint2 == b.ControlPoint2;
+
+        public static bool operator !=(CubicBezier a, CubicBezier b) => !(a == b);
```
derobinsMSFT, line 55: >!(a == b); [](start = 72, length = 10)

Nit; next line? (Cf. `CubicBezierEasing::operator !=`).

simeoncran: Why? Part of the beauty of the => is that things can be compacted onto the one line.

---
In reply to: [439517448](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439517448) [](ancestors = 439517448)

source/LottieData/Optimization/Optimizer.cs
```diff
@@ -271,9 +292,12 @@ public static ReadOnlySpan<KeyFrame<T>> RemoveRedundantKeyFrames<T>(in ReadOnlyS
             if (optimizedFrames is null)
             {
                 // No redundant key frames found yet.
-                // If the final 2 key frames have the same value, the final key frame is redundant.
-                if (keyFrames[keyFrames.Length - 1].Value.Equals(keyFrames[keyFrames.Length - 2].Value))
+                // If the final 2 key frames have the same value, the final key frame is redundant,
+                // unless it has a non-linear cubic Bezier easing.
+                if (keyFrames[keyFrames.Length - 1].Value.Equals(keyFrames[keyFrames.Length - 2].Value) &&
+                    !HasNonLinearCubicBezierEasing(keyFrames[keyFrames.Length - 1]))
```
derobinsMSFT, line 298: >!HasNonLinearCubicBezierEasing(keyFrames[keyFrames.Length - 1] [](start = 20, length = 62)

For this case as well as the case with some optimizedFrames below-- what if the second-to-last frame has a non-linear cubic Bezier easing? Above, a frame is not considered redundant if either it or the previous frame has one.

source/LottieReader/Serialization/LottieJsonObjectElement.cs
```diff
@@ -32,6 +32,52 @@ internal LottieJsonObjectElement(LottieCompositionReader owner, JsonElement wrap
                 Array.Sort(_properties, Comparer.Instance);
             }
 
+            // Parses an array of Vector2 stored as 2 arrays of equal-length
+            // "x" and "y" values.
+            // If "x" or "y" properties are missing, returns an empty array.
```
derobinsMSFT, line 37: >If "x" or "y" properties are missing, returns an empty array. [](start = 15, length = 61)

This doesn't quite match what the code's doing-- in particular, if `x` is a valid non-array value, this will return a single `Vector2` in an array, regardless of whether `y` has a value.

source/LottieData/CubicBezier.cs
derobinsMSFT: Very true-- in that case, condense the `CubicBezierEasing::operator !=` onto one line, for consistency.

---
In reply to: [439530193](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439530193) [](ancestors = 439530193,439517448)

source/LottieReader/Serialization/Animatables.cs
```diff
@@ -560,8 +559,8 @@ static bool HasKeyframes(in LottieJsonArrayElement? array)
                 Easing easing = HoldEasing.Instance;
 
                 // SpatialBeziers.
-                var ti = default(Vector3);
-                var to = default(Vector3);
+                var ti = default(Vector2);
```
derobinsMSFT, line 562: >ti [](start = 20, length = 2)

What do "ti' and "to" stand for exactly? Do they correspond to names of properties in Bodymovin? A comment here might be helpful.

simeoncran: They're the names of the properties on the object. See line 615. I think the "i" stands for "in" and "o" stands for "out" as in "ease-in" and "ease-out". Lottie uses lots of these terse names to (presumably) save space.
The comment already describes them as the SpatialBeziers values. Maybe there's a better name you can think of. I don't know what comment here would make it clearer.

---
In reply to: [439538419](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439538419) [](ancestors = 439538419)

source/LottieData/Optimization/Optimizer.cs
simeoncran: If the second to last frame has a non-linear cubic Bezier easing then it should still be there. Are you seeing a place where it gets removed?

---
In reply to: [439531373](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439531373) [](ancestors = 439531373)

derobinsMSFT: My mistake-- above, a frame is not redundant if either it or the _following_ frame has one.

---
In reply to: [439578256](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439578256) [](ancestors = 439578256,439531373)

source/LottieReader/Serialization/LottieJsonObjectElement.cs
simeoncran: Yeah, that's too complicated. I've changed how AsVector2() works to make this contract simpler. I don't know why the AsVector2() was biased towards the x value rather than caring about both and y. We'll see if it triggers any issues with tests.

---
In reply to: [439533964](https://github.com/windows-toolkit/Lottie-Windows/pull/284#discussion_r439533964) [](ancestors = 439533964)

PR #287, opened 2020-06-10
CppWinrt CodeGen Output Fixes

karenbtlai:
Fixes for some manual changes I had to do for CppWinrt Output:

- Adding `winrt` namespace for `Windows::UI::Color`, `Windows::Foundation::Metadata::ApiInformation`, `CompositionPropertySet`, `Microsoft::UI::Controls`
- Add `ColorAsVector4` method in header file
- Modify conditional statement for `GetThemeProperties` method in header file
- remove extra `compositionPropertySet` for ThemePropertyImpls methods


source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
```diff
@@ -111,7 +113,7 @@ protected override void WriteThemePropertyImpls(CodeBuilder builder)
                 builder.WriteLine();
             }
 
-            builder.WriteLine($"{T.CompositionPropertySet} {sourceClassQualifier}EnsureThemeProperties({T.Compositor} compositor)");
+            builder.WriteLine($"winrt::{T.CompositionPropertySet} {sourceClassQualifier}EnsureThemeProperties({T.Compositor} compositor)");
```
simeoncran, line 116: I'm wondering why the prefix is needed here. It looks like other winrt:: things work without a prefix in the .cpp file. I would have thought you'd only need the prefix in the header.

source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
```diff
@@ -81,6 +81,8 @@ private protected override void WriteThemeHeader(HeaderBuilder builder)
                 }
             }
 
+            builder.Private.WriteLine();
+            builder.Private.WriteLine($"winrt::float4 ColorAsVector4(winrt::Windows::UI::Color color);");
```
simeoncran, line 85: >winrt::float4 ColorAsVector4(winrt::Windows::UI::Color color);"); [](start = 40, length = 65)

Because this effectively makes ColorAsVector4 public, it should be changed in the C# and CX generators to be consistent.
But maybe it shouldn't be public in the first place. In CX and C# it's "internal".
Also it's a very simple function, so maybe it doesn't have much value being public - it's cheap and easy to just copy the implementation. I'd like to know more about your scenario. Note that if you're just setting color values on theme properties, you can do it with the properties that are generated. You only need this conversion if you are setting values in the CompositionPropertySet returned by EnsureThemeProperties. Is that what you're doing? Can we make the generated properties easier to use instead?

source/UIDataCodeGen/CodeGen/CppInstantiatorGeneratorBase.cs
```diff
@@ -416,7 +416,7 @@ protected override void WriteFileEnd(CodeBuilder builder)
             }
 
             // Generate the method that creates an instance of the composition on the IAnimatedVisualSource.
-            builder.WriteLine($"{S.Hatted(AnimatedVisualTypeName)} {_s.Namespace(SourceInfo.Namespace)}::{SourceClassName}::TryCreateAnimatedVisual(");
+            builder.WriteLine($"winrt::{S.Hatted(AnimatedVisualTypeName)} {_s.Namespace(SourceInfo.Namespace)}::{SourceClassName}::TryCreateAnimatedVisual(");
```
simeoncran, line 419: What is special about your case that you need the winrt:: prefix but others don't?

source/UIDataCodeGen/CodeGen/CppInstantiatorGeneratorBase.cs
```diff
@@ -365,7 +365,7 @@ IEnumerable<string> GetConstructorParameters(IAnimatedVisualInfo info)
             // Write the IsRuntimeCompatible static method.
             builder.WriteLine("static bool IsRuntimeCompatible()");
             builder.OpenScope();
-            builder.WriteLine($"return Windows::Foundation::Metadata::ApiInformation::IsApiContractPresent({_s.String("Windows.Foundation.UniversalApiContract")}, {info.RequiredUapVersion});");
+            builder.WriteLine($"return winrt::Windows::Foundation::Metadata::ApiInformation::IsApiContractPresent({_s.String("Windows.Foundation.UniversalApiContract")}, {info.RequiredUapVersion});");
```
simeoncran, line 368: If winrt:: prefixes are added in this file, you first need to check whether you are generating CX or winrtcpp. winrt:: makes no sense for CX.

source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
karenbtlai: Hm, it seems as I resolved other namespace issues, CompositionPropertySet is not longer causing build errors. Will remove.

source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
karenbtlai: The method ColorAsVector4 was simply giving build errors for not having it declared in the header file, which is why I added it there.

source/UIDataCodeGen/CodeGen/CppInstantiatorGeneratorBase.cs
karenbtlai: It seems `Microsoft::UI::Xaml::Controls` is not recognized as a namespace name by itself. It would either require adding `winrt::` in `WriteLine()`, or declare it with using namespace at the top with `using namespace winrt::Microsoft::UI::Xaml::Controls;`

source/UIDataCodeGen/CodeGen/CppInstantiatorGeneratorBase.cs
karenbtlai: I see, didn't realize this file also generates output for CppCx. Fixed.

---
In reply to: [438457622](https://github.com/windows-toolkit/Lottie-Windows/pull/287#discussion_r438457622) [](ancestors = 438457622)

source/UIDataCodeGen/CodeGen/CppInstantiatorGeneratorBase.cs
```diff
@@ -890,8 +909,11 @@ void WriteIAnimatedVisualSourceHeaderText(HeaderBuilder builder, IAnimatedVisual
             // type of the IAnimatedVisualSource is not known. It relies on a custom interface that declares
             // it, so if we're not generating code for a custom interface, there's no reason to generate
             // the method.
-            if (IsInterfaceCustom && SourceInfo.IsThemed)
```
simeoncran, line 893: You're changing the semantics here, but the comment doesn't reflect that. I don't think this is right anyway - isn't this making GetThemeProperties always implemented, but virtual if the interface is no IAnimatedVisualSource?
Can you reply with more info about what you're trying to have it do for the (I think) 2 cases: IAnimatedVisualSource, custom interface.

source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
simeoncran: Got it now, but make it a static member.

---
In reply to: [439606946](https://github.com/windows-toolkit/Lottie-Windows/pull/287#discussion_r439606946) [](ancestors = 439606946)

source/UIDataCodeGen/CodeGen/CppInstantiatorGeneratorBase.cs
```diff
@@ -246,7 +246,14 @@ IEnumerable<string> GetConstructorParameters(IAnimatedVisualInfo info)
 
             if (info.AnimatedVisualSourceInfo.IsThemed)
             {
-                yield return $"{_typeName.CompositionPropertySet} themeProperties";
+                if (_isCppwinrtMode)
+                {
+                    yield return $"winrt::{_typeName.CompositionPropertySet} themeProperties";
```
simeoncran, line 251: If the prefix is necessary here, why wouldn't it be necessary on _typeName.Compositor?

simeoncran:
This isn't getting any traction, and I've rewritten the cppwinrt support in an upcoming PR that will supersede this.

PR #288, opened 2020-06-10
Handle rounded rectangles more accurately

simeoncran:
Composition and After Effects use different models for expression rounded rectangles. Composition has a Vector2 CornerRadius that can be set to anything. As the radius trends towards half the size the rectangle looks more like an ellipse, but when the CornerRadius gets bigger than half the size it starts looking like a weird 4-sided star thing. After Effects uses a scalar Roundness value and always caps the value so that you can't go beyond a circle.

Previously we were not accounting for this difference in models.

This is not a complete fix, but it fixes a lot of egregious cases. It's not complete because there is another After Effects concept called "Rounded corners" which interacts with rounded rectangles in a complex fashion, and we don't yet support that correctly. It seems to be really rare though, so not a big deal.

This change also undoes part of a previous "fix" to the code for stopping an animation on a WinCompData.CompositionObject. Due to a misunderstanding about how that's supposed to work, we would stop some animations that shouldn't have been stopped.

### DIFF
--- a/source/LottieData/RoundedCorner.cs
+++ b/source/LottieData/RoundedCorner.cs
@@ -18,12 +18,15 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
         }
 
         /// <summary>
-        /// The radius of the rounding. If the shape to which this applies is a rectangle
-        /// the rounding will only apply if the rectangle a non-0 roundess value.
+        /// The radius of the rounding.
+        /// </summary>
+        /// <remarks>
+        /// If the shape to which this applies is a rectangle
+        /// the rounding will only apply if the rectangle has a 0 roundess value.
         /// Once the radius value reaches half of the largest dimension of the
         /// rectangle, the result will be equivalent to an ellipse of the same size, and
         /// increasing the radius further will have no further effect.
-        /// </summary>
+        /// </remarks>
         public Animatable<double> Radius { get; }
 
         /// <inheritdoc/>
--- a/source/LottieData/RoundedCorner.cs
+++ b/source/LottieData/RoundedCorner.cs
@@ -21,10 +21,10 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
         /// The radius of the rounding.
         /// </summary>
         /// <remarks>
-        /// If the shape to which this applies is a rectangle
-        /// the rounding will only apply if the rectangle has a 0 roundess value.
-        /// Once the radius value reaches half of the largest dimension of the
-        /// rectangle, the result will be equivalent to an ellipse of the same size, and
+        /// If the shape to which this applies is a rectangle, the rounding will
+        /// only apply if the rectangle has a 0 roundness value. Once the radius
+        /// value reaches half of the largest dimension of the rectangle, the
+        /// result will be equivalent to an ellipse of the same size, and
         /// increasing the radius further will have no further effect.
         /// </remarks>
         public Animatable<double> Radius { get; }
--- a/source/LottieData/RoundedCorner.cs
+++ b/source/LottieData/RoundedCorner.cs
@@ -17,6 +17,13 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
             Radius = radius;
         }
 
+        /// <summary>
+        /// The radius of the rounding. If the shape to which this applies is a rectangle
+        /// the rounding will only apply if the rectangle a non-0 roundess value.
+        /// Once the radius value reaches half of the largest dimension of the
+        /// rectangle, the result will be equivalent to an ellipse of the same size, and
+        /// increasing the radius further will have no further effect.
+        /// </summary>
         public Animatable<double> Radius { get; }
 
         /// <inheritdoc/>
--- a/source/LottieToWinComp/ExpressionFactory.cs
+++ b/source/LottieToWinComp/ExpressionFactory.cs
@@ -69,11 +69,14 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
         internal static Scalar RootScalar(string propertyName) => Scalar(RootProperty(propertyName));
 
-        internal static Scalar ConstrainedCornerRadiusScalar(double roundness) => Min(roundness, Min(MySize.X, MySize.Y) / 2);
+        internal static Vector2 ConstrainedCornerRadiusScalar(double roundness)
+            => Vector2(Min(roundness, Min(MySize.X, MySize.Y) / 2), Min(roundness, Min(MySize.X, MySize.Y) / 2));
 
-        internal static Scalar ConstrainedCornerRadiusScalar() => Min(MyRoundness, Min(MySize.X, MySize.Y) / 2);
+        internal static Vector2 ConstrainedCornerRadiusScalar()
+            => Vector2(Min(MyRoundness, Min(MySize.X, MySize.Y) / 2), Min(MyRoundness, Min(MySize.X, MySize.Y) / 2));
 
-        internal static Scalar ConstrainedCornerRadiusScalar(Sn.Vector2 size) => Min(MyRoundness, Math.Min(size.X, size.Y) / 2);
+        internal static Vector2 ConstrainedCornerRadiusScalar(Sn.Vector2 size)
+            => Vector2(Min(MyRoundness, Math.Min(size.X, size.Y) / 2), Min(MyRoundness, Math.Min(size.X, size.Y) / 2));
 
         // The value of a Color property stored as a Vector4 on the theming property set.
         static Vector4 ThemedColor4Property(string propertyName) => Vector4(ThemeProperty(propertyName));
--- a/source/LottieToWinComp/ExpressionFactory.cs
+++ b/source/LottieToWinComp/ExpressionFactory.cs
@@ -22,11 +22,13 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         internal static readonly Vector2 MyAnchor = MyVector2("Anchor");
         internal static readonly Vector3 MyAnchor3 = Vector3(MyAnchor.X, MyAnchor.Y, 0);
         internal static readonly Vector4 MyColor = MyVector4("Color");
+        internal static readonly Scalar MyCornerRadiusX = MyScalar("CornerRadius.X");
         internal static readonly Scalar MyInheritedOpacity = MyScalar("InheritedOpacity");
         internal static readonly Scalar MyOpacity = MyScalar("Opacity");
         internal static readonly Vector2 MyPosition = MyVector2("Position");
         internal static readonly Vector2 MySize = MyVector2("Size");
         internal static readonly Matrix3x2 MyTransformMatrix = MyMatrix3x2("TransformMatrix");
+        static readonly Scalar MyRoundness = MyScalar("Roundness");
         static readonly Scalar MyTStart = MyScalar("TStart");
         static readonly Scalar MyTEnd = MyScalar("TEnd");
 
@@ -66,6 +68,12 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         internal static Vector2 PositionToOffsetExpression(Sn.Vector2 position) => Vector2(position) - HalfMySize;
 
         internal static Scalar RootScalar(string propertyName) => Scalar(RootProperty(propertyName));
+
+        internal static Scalar ConstrainedCornerRadiusScalar(double roundness) => Min(roundness, Min(MySize.X, MySize.Y) / 2);
+
+        internal static Scalar ConstrainedCornerRadiusScalar() => Min(MyRoundness, Min(MySize.X, MySize.Y) / 2);
+
+        internal static Scalar ConstrainedCornerRadiusScalar(Sn.Vector2 size) => Min(MyRoundness, Math.Min(size.X, size.Y) / 2);
 
         // The value of a Color property stored as a Vector4 on the theming property set.
         static Vector4 ThemedColor4Property(string propertyName) => Vector4(ThemeProperty(propertyName));
--- a/source/LottieToWinComp/ExpressionFactory.cs
+++ b/source/LottieToWinComp/ExpressionFactory.cs
@@ -22,7 +22,6 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         internal static readonly Vector2 MyAnchor = MyVector2("Anchor");
         internal static readonly Vector3 MyAnchor3 = Vector3(MyAnchor.X, MyAnchor.Y, 0);
         internal static readonly Vector4 MyColor = MyVector4("Color");
-        internal static readonly Scalar MyCornerRadiusX = MyScalar("CornerRadius.X");
         internal static readonly Scalar MyInheritedOpacity = MyScalar("InheritedOpacity");
         internal static readonly Scalar MyOpacity = MyScalar("Opacity");
         internal static readonly Vector2 MyPosition = MyVector2("Position");
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -23,7 +23,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Runtime.InteropServices;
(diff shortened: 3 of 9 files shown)


source/LottieData/RoundedCorner.cs
```diff
@@ -17,6 +17,16 @@ sealed class RoundedCorner : ShapeLayerContent
             Radius = radius;
         }
 
+        /// <summary>
+        /// The radius of the rounding.
+        /// </summary>
+        /// <remarks>
+        /// If the shape to which this applies is a rectangle
+        /// the rounding will only apply if the rectangle has a 0 roundess value.
```
eliezerpMS, line 25: >roundess [](start = 66, length = 8)

roundness

source/LottieData/RoundedCorner.cs
```diff
@@ -17,6 +17,16 @@ sealed class RoundedCorner : ShapeLayerContent
             Radius = radius;
         }
 
+        /// <summary>
+        /// The radius of the rounding.
+        /// </summary>
+        /// <remarks>
+        /// If the shape to which this applies is a rectangle
```
eliezerpMS, line 24: Just curious what is the reasoning behind the comment wrapping length here?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2231,141 +2231,200 @@ CompositionShape TranslateEllipseContent(TranslationContext context, ShapeConten
...
+            // Use a non-rounded rectangle geometry.
+            if (_targetUapVersion <= 7)
+            {
+                // V7 did not reliably draw non-rounded rectangles.
+                // Work around the problem by using a rounded rectangle with a tiny corner radius.
+                var roundedRectangleGeometry = _c.CreateRoundedRectangleGeometry();
+                geometry = roundedRectangleGeometry;
+
+                // NOTE: magic tiny corner radius number - do not change!
+                roundedRectangleGeometry.CornerRadius = new Sn.Vector2(0.000001F);
```
eliezerpMS, line 2283: >0.000001F [](start = 71, length = 9)

Does an epsilon work or is this value from something in dcomp or dwmcore?

eliezerpMS, line 2283: > roundedRectangleGeometry.CornerRadius = new Sn.Vector2(0.000001F); [](start = 15, length = 67)

Nit: Is this the only difference with the if/else contents? If so then maybe just consolidate the two somehow?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2231,141 +2231,200 @@ CompositionShape TranslateEllipseContent(TranslationContext context, ShapeConten
...
+            var geometry = _c.CreateRoundedRectangleGeometry();
+            compositionShape.Geometry = geometry;
+
+            // If a RoundedRectangle is in the context, use it to override the roundness unless the roundness is non-0.
+            var cornerRadius = context.TrimAnimatable(
+                shapeContext.RoundedCorner != null && shapeContent.Roundness.AlwaysEquals(0)
+                    ? shapeContext.RoundedCorner.Radius
+                    : shapeContent.Roundness);
+
+            // In After Effects, the rectangle Roundness has no further affect once it reaches min(Size.X, Size.Y)/2.
```
eliezerpMS, line 2333: >affect [](start = 72, length = 6)

Effect?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2231,141 +2231,200 @@ CompositionShape TranslateEllipseContent(TranslationContext context, ShapeConten
...
-                    ApplyScalarKeyFrameAnimation(context, cornerRadius, geometry, "CornerRadius.X");
-                    ApplyScalarKeyFrameAnimation(context, cornerRadius, geometry, "CornerRadius.Y");
-                }
-                else
-                {
-                    geometry.CornerRadius = Vector2((float)cornerRadius.InitialValue);
-                }
+            // Convert size and position into offset. This is necessary because a geometry's offset is for
+            // its top left corner, whereas a Lottie position is for its centerpoint.
+            geometry.Offset = Vector2(position.InitialValue - (size.InitialValue / 2));
```
eliezerpMS, line 2380: >Offset = Vector2(position.InitialValue - (size.InitialValue / 2)); [](start = 21, length = 66)

Perhaps a nit but should this just be a helper function with a name like "ConvertCenterPoint" or "LottiePositionToDcompPosition" and then you would not have to keep typing out this comment?

source/LottieData/RoundedCorner.cs
simeoncran: It's art. OK, I'll switch back to engineering and line things up better.

---
In reply to: [438454421](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438454421) [](ancestors = 438454421)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Well, it's magic, so who knows. :-)
I'm frightened to touch it. It was an emergency workaround when the problem with rectangles was discovered. The fear is that there might be some optimization in the DWM code that decides to branch to non-rounded rectangle code if the corner radius is small enough. Even if that optimization didn't exist when this change was made, maybe it was going to turn up before shipping. Maybe this value doesn't work if the rectangle is large enough. Anyhow, we tried this value, it looked good on everything we threw at it, and then vowed to never speak of it again. Until today . . . :-)
Epsilon is rarely a useful value because as soon as you use it in an expression you risk having it evaporate, e.g. 10+epsilon == 10  is true.

---
In reply to: [438455366](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438455366) [](ancestors = 438455366)

simeoncran: You mean between the V8 and V7 branches? The problem is with the "var rectangleGeometry" - it has unrelated types in each branch (RoundedRectangle and Rectangle). I tried. Maybe I'm missing something but I don't see it yet.

---
In reply to: [438455853](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438455853) [](ancestors = 438455853)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: You're right!

---
In reply to: [438456471](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438456471) [](ancestors = 438456471)

source/WinCompData/CompositionObject.cs
```diff
@@ -141,15 +141,10 @@ public void StartAnimation(string target, CompositionAnimation animation)
...
             // For example, if the property is TransformMatrix we must also stop animations
-            // on TransformMatrix.M11, TransformMatrix.M12, etc; and if the property is
-            // TransformMatrix.M11 we must also stop animations on TransformMatrix,
-            // TransformMatrix.M12, etc.
-            //
-            // If there's a dot in the name it is a sub-channel name.
-            var firstDotIndex = propertyName.IndexOf('.');
-            var rootPropertyPrefix = $"{(firstDotIndex >= 0 ? propertyName.Substring(0, firstDotIndex) : propertyName)}.";
+            // on TransformMatrix.M11, TransformMatrix.M12, etc..
+            var rootPropertyPrefix = $"{propertyName}.";
```
derobinsMSFT, line 147: >rootPropertyPrefix [](start = 16, length = 18)

Can this be renamed just `propertyPrefix`? The provided `propertyName` may no longer be the actual root property.

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Not a nit at all- I'll do it.

---
In reply to: [438457505](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438457505) [](ancestors = 438457505)

source/WinCompData/CompositionObject.cs
simeoncran: Both names seem imperfect to me. But it's not the prefix of the property, it's a prefix created using the property name as a root, so rootPropertyPrefix is slightly better. Maybe there's a third option.

---
In reply to: [438461767](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438461767) [](ancestors = 438461767)

source/LottieToWinComp/LottieToWinCompTranslator.cs
eliezerpMS: True I guess one is rounded and one is not.

---
In reply to: [438461123](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r438461123) [](ancestors = 438461123,438455853)

source/WinCompData/CompositionObject.cs
raymond-wh-leung, line 147: >$"{propertyName}."; [](start = 37, length = 19)

The prefix could be wrong if the propertyName already has a subchannel in it? (i.e. if propertyname is TransformMatrix.M11)

raymond-wh-leung:
>                     animatorPropertyName.StartsWith(rootPropertyPrefix))

If there are two animations running, one for TransformMatrix.M11, and the other for TransformMatrix.M12, calling StopAnimation(TransformMatrix.M12) would not stop TransformMatrix.M11 coz they animate on two different channels. (they don't intersect each other).

---
Refers to: source/WinCompData/CompositionObject.cs:154 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

raymond-wh-leung:
>             StopAnimation(target);

Likewise, if there is an animation running of "Offset.X", starting an vector3 animation on "Offset" would stop that "Offset.X' animation because the vector3 animations will intersect with that "Offset.X" animation for the X channel.

---
Refers to: source/WinCompData/CompositionObject.cs:119 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

source/WinCompData/CompositionObject.cs
```diff
@@ -141,22 +141,24 @@ public void StartAnimation(string target, CompositionAnimation animation)
...
         {
-            // We also need to stop animations on any sub-channels and super-channels.
-            // For example, if the property is TransformMatrix we must also stop animations
-            // on TransformMatrix.M11, TransformMatrix.M12, etc; and if the property is
-            // TransformMatrix.M11 we must also stop animations on TransformMatrix,
-            // TransformMatrix.M12, etc.
-            //
+            // We also need to stop animations on any sub-channels and the root property.
+            // Examples:
+            //  Sub-channels: stopping TransformMatrix must also stop
```
raymond-wh-leung, line 146: >TransformMatrix [](start = 39, length = 15)

nit: consider using another property for the example (coz we don't have an animation to animate the matrix property).

raymond-wh-leung:
>                     animatorPropertyName.StartsWith(rootPropertyPrefix))

Looks like it could still be an issue in the latest. (i.e. in the latest change, stopping Offset.X would also stop Offset.Y?). For the comp animation's case, "Offset.Y" would not be stopped because they target on different channels.

---
In reply to: [642769088](https://github.com/windows-toolkit/Lottie-Windows/pull/288#issuecomment-642769088) [](ancestors = 642769088)

---
Refers to: source/WinCompData/CompositionObject.cs:154 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

simeoncran:
>                     animatorPropertyName.StartsWith(rootPropertyPrefix))

I think it's correct. If you are stopping Offset.X:  subChannelPrefix, "Offset.X." and rootPropertyName == "Offset", and propertyName="Offset.X".  So we'd stop Offset.X and Offset and Offset.X.* but Offset.Y would not be affected. Note the stopping Offset will not stop Offset.Y.

Are there any cases where we can have 2 dots in the property name, e.g. Foo.Bar.X ?  We don't handle that case.


---
In reply to: [642851655](https://github.com/windows-toolkit/Lottie-Windows/pull/288#issuecomment-642851655) [](ancestors = 642851655,642769088)

---
Refers to: source/WinCompData/CompositionObject.cs:154 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

source/LottieToWinComp/ExpressionFactory.cs
```diff
@@ -22,11 +22,13 @@ static class ExpressionFactory
         internal static readonly Vector2 MyAnchor = MyVector2("Anchor");
         internal static readonly Vector3 MyAnchor3 = Vector3(MyAnchor.X, MyAnchor.Y, 0);
         internal static readonly Vector4 MyColor = MyVector4("Color");
+        internal static readonly Scalar MyCornerRadiusX = MyScalar("CornerRadius.X");
```
derobinsMSFT, line 25: >MyCornerRadiusX [](start = 40, length = 15)

This looks unused.

raymond-wh-leung:
>                     animatorPropertyName.StartsWith(rootPropertyPrefix))

Ah right. From the last part of your statement. ("Note the stopping Offset will not stop Offset.Y"). From the changes, stopping "Offset" will stop "Offset.Y" (which is correct), so want to make sure I understand that statement correctly?

There shouldn't be 2 dots for most of the compositionobjects. The tricky one is the composition effect, I think it can store property in the property set (like saturation.saturation) in the propertyset which is animatable, I was trying to find for the effects case there are cases with property with 2 dots but couldn't find it.  (it seems that it might be possible for the effects to have 2 dots in the property. fyi).

---
In reply to: [642874355](https://github.com/windows-toolkit/Lottie-Windows/pull/288#issuecomment-642874355) [](ancestors = 642874355,642851655,642769088)

---
Refers to: source/WinCompData/CompositionObject.cs:154 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

source/LottieToWinComp/ExpressionFactory.cs
simeoncran: That's true!

---
In reply to: [439033602](https://github.com/windows-toolkit/Lottie-Windows/pull/288#discussion_r439033602) [](ancestors = 439033602)

simeoncran:
>                     animatorPropertyName.StartsWith(rootPropertyPrefix))

Yeah, I think I meant stopping Offset.X will not stop Offset.Y.

For the 2 dots, I know it won't affect Lottie because we aren't using effects like that yet, so it's probably ok to not get that correct yet. If it's as simple as us looking for the last dot instead of the first dot then I could do that now, but I don't understand the case yet (saturation.saturation... ?) so I'm fine with ignoring it for now. Thoughts?

---
In reply to: [642897292](https://github.com/windows-toolkit/Lottie-Windows/pull/288#issuecomment-642897292) [](ancestors = 642897292,642874355,642851655,642769088)

---
Refers to: source/WinCompData/CompositionObject.cs:154 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

raymond-wh-leung:
>                     animatorPropertyName.StartsWith(rootPropertyPrefix))

Yup agreed.

---
In reply to: [642914865](https://github.com/windows-toolkit/Lottie-Windows/pull/288#issuecomment-642914865) [](ancestors = 642914865,642897292,642874355,642851655,642769088)

---
Refers to: source/WinCompData/CompositionObject.cs:154 in 70334d1. [](commit_id = 70334d10be8001ea85aa244fdf2bab03a8296b93, deletion_comment = False)

source/WinCompData/CompositionObject.cs
```diff
@@ -141,22 +141,24 @@ public void StartAnimation(string target, CompositionAnimation animation)
...
-            // on TransformMatrix.M11, TransformMatrix.M12, etc; and if the property is
-            // TransformMatrix.M11 we must also stop animations on TransformMatrix,
-            // TransformMatrix.M12, etc.
-            //
+            // We also need to stop animations on any sub-channels and the root property.
+            // Examples:
+            //  Sub-channels: stopping Offset must also stop
+            //    Offset.X, Offset.Y, etc..
+            //  Root property: stopping Offset.X must also
+            //    stop OFfset.
```
derobinsMSFT, line 149: >F [](start = 24, length = 1)

Lowercase.

PR #290, opened 2020-06-12
Update nuget packages to the latest versions.

simeoncran:


scott-moore-ms:
Wait, I didn't say that!

---
In reply to: [429982327](https://github.com/windows-toolkit/Lottie-Windows/pull/290#issuecomment-429982327) [](ancestors = 429982327)

PR #294, opened 2020-06-15
Prevent duplicated keyframe progress

simeoncran:
I found a Lottie file that during translation would generate 2 keyframes at the same progress. That's illegal for Composition.
The Lottie file that triggered this had one path in a geometry group at frame 48 and another in the same group at 48.000000357.

This issue is caused by the fact that we convert unique key frame numbers as doubles to progress values as floats. Floats have less resolution, so if 2 keyframes were really close together then they might end up with identical progress values.

It's also an issue when we group animated paths because the 2 paths might have key frame numbers that are slightly different but end up at the same progress value. As long as we prevent identical progress values that's fine, but if we can be a bit smarter about how we compare frame numbers during grouping we can group more effectively.

### DIFF
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -2475,7 +2475,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         // Groups multiple Shapes into a D2D geometry group.
         CompositionShape TranslatePathGroupContent(TranslationContext context, ShapeContentContext shapeContext, IEnumerable<Path> paths)
         {
-            var groupingSucceeded = PathGeometryGroup.TryGroupPaths(paths, out var grouped);
+            var groupingSucceeded = PathGeometryGroup.TryGroupPaths(context, paths, out var grouped);
 
             // If any of the paths have different directions we may not get the translation
             // right, so check that case and warn the user.
@@ -4293,7 +4293,16 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
             foreach (var keyFrame in trimmedKeyFrames)
             {
+                // The progress of the current key frame, but adjusted as extra key frames
+                // are added to ensure we never add a key frame with the same frame number twice.
                 var adjustedProgress = (keyFrame.Frame - animationStartTime) / animationDuration;
+
+                if ((float)adjustedProgress <= previousProgress)
+                {
+                    // Prevent the next key frame from being inserted at the same progress value
+                    // as the one we just inserted.
+                    adjustedProgress = Float32.NextLargerThan(previousProgress);
+                }
 
                 if (keyFrame.SpatialControlPoint1 != default(Vector3) || keyFrame.SpatialControlPoint2 != default(Vector3))
                 {
@@ -4392,7 +4401,15 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                     if (previousKeyFrameWasExpression)
                     {
                         // Ensure the previous expression doesn't continue being evaluated during the current keyframe.
-                        insertKeyFrame(compositionAnimation, Float32.NextLargerThan(previousProgress), previousValue, _c.CreateStepThenHoldEasingFunction());
+                        var nextLargerThanPrevious = Float32.NextLargerThan(previousProgress);
+                        insertKeyFrame(compositionAnimation, nextLargerThanPrevious, previousValue, _c.CreateStepThenHoldEasingFunction());
+
+                        if ((float)adjustedProgress <= nextLargerThanPrevious)
+                        {
+                            // Prevent the next key frame from being inserted at the same progress value
+                            // as the one we just inserted.
+                            adjustedProgress = Float32.NextLargerThan(nextLargerThanPrevious);
+                        }
                     }
 
                     insertKeyFrame(compositionAnimation, (float)adjustedProgress, keyFrame.Value, _c.CreateCompositionEasingFunction(keyFrame.Easing));
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -4454,12 +4454,12 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             {
                 AdjustProgress(ref progress);
 
-                if (progress > 1)
-                {
-                    // We ran out of room. The key frame will be ignored.
-                    Debug.Fail("Progress > 1");
-                }
-                else
+                // If progress is > 1 then we have no more room to add key frames.
+                // This can happen as a result of extra key frames being added for
+                // various reasons. The dropped key frames shouldn't matter as they
+                // would only affect a very small amount of time at the end of the
+                // animation.
+                if (progress <= 1)
                 {
                     keyFrameInserter(animation, progress, value, easing);
                 }
@@ -4473,12 +4473,12 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             {
                 AdjustProgress(ref progress);
 
-                if (progress > 1)
-                {
-                    // We ran out of room. The key frame will be ignored.
-                    Debug.Fail("Progress > 1");
-                }
-                else
+                // If progress is > 1 then we have no more room to add key frames.
+                // This can happen as a result of extra key frames being added for
+                // various reasons. The dropped key frames shouldn't matter as they
+                // would only affect a very small amount of time at the end of the
+                // animation.
+                if (progress <= 1)
                 {
                     expressionKeyFrameInserter(animation, progress, expression, easing);
                 }
--- a/source/LottieToWinComp/LottieToWinCompTranslator.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.cs
@@ -2368,7 +2368,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 geometry.CornerRadius = Vector2((float)cornerRadiusValue);
             }
 
-            geometry.Offset = InitialOffset(size:size, position:position);
+            geometry.Offset = InitialOffset(size: size, position: position);
 
             if (!size.IsAnimated)
             {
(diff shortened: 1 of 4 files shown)


source/LottieToWinComp/TranslationContext.cs
```diff
@@ -137,5 +156,27 @@ internal Root(LottieComposition lottieComposition)
             {
             }
         }
+
+        /// <summary>
+        /// Compares frame numbers for equality. This takes into account the lossiness of the conversion
+        /// that is done from <see cref="float"/> frame numbers to <see cref="double"/> progress values.
```
derobinsMSFT, line 162: ><see cref="float"/> [](start = 30, length = 19)

"double frame numbers". . .  "float progress values"?

simeoncran: OK, that's backward. But I think you're asking what they mean. They're Lottie and Comp concepts so I don't think they need explaining here. Lottie specifies the location on the timeline as a frame number with double precision. Comp specified the location on the timeline as progress values [0..1] with single precision.

---
In reply to: [441047085](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441047085) [](ancestors = 441047085)

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -52,6 +56,7 @@ internal static bool TryGroupPaths(IEnumerable<Path> paths, out Animatable<PathG
                 // Add any keyframes.
                 foreach (var kf in p.Data.KeyFrames.ToArray().Skip(1))
                 {
+                    // See if there's a key frame at the frame number already.
```
derobinsMSFT, line 59: >// See if there's a key frame at the frame number already. [](start = 20, length = 58)

If two KeyFrame values map to the same progress value, the GeometryKeyFrame for a given path (`i`) and its corresponding KeyFrame Value and Easing data will be overwritten by the second KeyFrame. Wanted to make sure-- is this okay? When creating a KeyFrameAnimation in the translator, it appears you're careful to adjust the progress for each KeyFrame as necessary so that no KeyFrames/KeyFrame data is lost or overwritten.

simeoncran: No, that's not ok. I don't see it though. Can you explain where that is happening?

---
In reply to: [441088286](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441088286) [](ancestors = 441088286)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4293,8 +4293,17 @@ void StartKeyframeAnimation(CompositionObject compObject, string target, KeyFram
...
             {
+                // The progress of the current key frame, but adjusted as extra key frames
+                // are added to ensure we never add a key frame with the same frame number twice.
                 var adjustedProgress = (keyFrame.Frame - animationStartTime) / animationDuration;
 
+                if ((float)adjustedProgress <= previousProgress)
+                {
+                    // Prevent the next key frame from being inserted at the same progress value
+                    // as the one we just inserted.
+                    adjustedProgress = Float32.NextLargerThan(previousProgress);
```
derobinsMSFT, line 4304: >adjustedProgress [](start = 20, length = 16)

Should there be a check to make sure this doesn't exceed a value of 1 (e.g., if there were several KeyFrames right at the end of an animation that had the same progress value)? Perhaps this would be rare/unlikely in practice.

source/LottieToWinComp/TranslationContext.cs
```diff
@@ -137,5 +156,27 @@ internal Root(LottieComposition lottieComposition)
...
+                _context = context;
+            }
+
+            public bool Equals(double x, double y) => ProgressOf(x) == ProgressOf(y);
+
+            public int GetHashCode(double obj) => ProgressOf(obj).GetHashCode();
+
+            // Converts a frame number into a progress value.
+            float ProgressOf(double value) =>
+                (float)((value - _context.StartTime) / _context.DurationInFrames);
```
scott-moore-ms, line 179: >(float) [](start = 16, length = 7)

In the comment above you mention that this takes into account the lossiness of the conversion - that's achieved by doing the math and inducing this cast to a float on both x and y?

source/LottieToWinComp/TranslationContext.cs
derobinsMSFT: Thanks for the explanation! Sorry for the question mark-- was pointing out that these were flipped.

---
In reply to: [441085286](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441085286) [](ancestors = 441085286,441047085)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: That's a good thought. I don't know what I would do here if that case arose, so for now I think the easiest thing is to check for that in CompositionObject.StartAnimation and blow up.

---
In reply to: [441093524](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441093524) [](ancestors = 441093524)

source/LottieToWinComp/TranslationContext.cs
simeoncran: Correct.

---
In reply to: [441094912](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441094912) [](ancestors = 441094912)

scott-moore-ms:
>                         insertKeyFrame(compositionAnimation, (float)adjustedProgress, keyFrame.Value, _c.CreateCompositionEasingFunction(easing));

I don't understand why the fixups aren't required here on these?

---
Refers to: source/LottieToWinComp/LottieToWinCompTranslator.cs:4354 in 7b15cda. [](commit_id = 7b15cda6decfabef95781f5a1c723f31a2749f49, deletion_comment = False)

simeoncran:
>                         insertKeyFrame(compositionAnimation, (float)adjustedProgress, keyFrame.Value, _c.CreateCompositionEasingFunction(easing));

Yeah, me either. This method is waaaay complicated.
I think I'll rewrite this a bit so that inserting a key frame will ensure that we never create 2 key frames at the same progress value. That way we can remove the checks and adjustments from this loop.

---
In reply to: [644977147](https://github.com/windows-toolkit/Lottie-Windows/pull/294#issuecomment-644977147) [](ancestors = 644977147)

---
Refers to: source/LottieToWinComp/LottieToWinCompTranslator.cs:4354 in 7b15cda. [](commit_id = 7b15cda6decfabef95781f5a1c723f31a2749f49, deletion_comment = False)

source/LottieToWinComp/PathGeometryGroup.cs
simeoncran: I'll add a comment about why this is imperfect.

---
In reply to: [441092706](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441092706) [](ancestors = 441092706,441088286)

raymond-wh-leung:
>                         insertKeyFrame(compositionAnimation, (float)adjustedProgress, keyFrame.Value, _c.CreateCompositionEasingFunction(easing));

Btw on comp side, when you called InsertKeyFrame multiple times with the same keyframe time, the last one wins (the previous one at that keyframe time will get overwritten) on the animation template.

---
In reply to: [645000723](https://github.com/windows-toolkit/Lottie-Windows/pull/294#issuecomment-645000723) [](ancestors = 645000723,644977147)

---
Refers to: source/LottieToWinComp/LottieToWinCompTranslator.cs:4354 in 7b15cda. [](commit_id = 7b15cda6decfabef95781f5a1c723f31a2749f49, deletion_comment = False)

raymond-wh-leung:
Fyi, on the comp side, you can call InsertKeyFrame() multiple times with the same keyframe time. (the previous one will get overwritten).
(curious if you are seeing a failure or something?)

simeoncran:
I don't remember whether I saw a failure or not. But that's good to know.

---
In reply to: [645033715](https://github.com/windows-toolkit/Lottie-Windows/pull/294#issuecomment-645033715) [](ancestors = 645033715)

simeoncran:
>                         insertKeyFrame(compositionAnimation, (float)adjustedProgress, keyFrame.Value, _c.CreateCompositionEasingFunction(easing));

WinCompData didn't allow that, but I'll change it so it does.
Here though we should never add key frames at the same progress value.

---
In reply to: [645030772](https://github.com/windows-toolkit/Lottie-Windows/pull/294#issuecomment-645030772) [](ancestors = 645030772,645000723,644977147)

---
Refers to: source/LottieToWinComp/LottieToWinCompTranslator.cs:4354 in 7b15cda. [](commit_id = 7b15cda6decfabef95781f5a1c723f31a2749f49, deletion_comment = False)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4420,20 +4431,58 @@ void StartKeyframeAnimation(CompositionObject compObject, string target, KeyFram
...
-        float GetOutPointProgress(TranslationContext context)
-        {
-            var result = (context.Layer.OutPoint - context.StartTime) / context.DurationInFrames;
+            // Local method to ensure we never insert more than 1 key frame with
+            // the same progress value. This relies on the key frames being inserted
+            // in order, so if we get a key frame with the same progress value as
+            // the previous one we'll just adjust the progress value up slightly.
+            void InsertKeyFrame(TCA animation, float progress, T value, CompositionEasingFunction easing)
+            {
+                AdjustProgress(ref progress);
```
raymond-wh-leung, line 4455: >AdjustProgress [](start = 16, length = 14)

Is this still needed?

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4420,20 +4431,58 @@ void StartKeyframeAnimation(CompositionObject compObject, string target, KeyFram
...
-        float GetInPointProgress(TranslationContext context)
-        {
-            var result = (context.Layer.InPoint - context.StartTime) / context.DurationInFrames;
+            // If the given progress value is equal to a progress value that was already
+            // inserted into the animation, adjust it up to ensure we never try to
+            // insert a key frame on top of an existing key frame. This relies on the
+            // key frames being inserted in order.
+            void AdjustProgress(ref float progress)
+            {
+                if (progress == highestProgressValueSoFar)
```
raymond-wh-leung, line 4441: >highestProgressValueSoFar [](start = 32, length = 25)

Looks like this is a local variable in another function?

simeoncran: Yes, it's in the enclosing function. This is a local function, meaning it is local to the function it is contained in.

---
In reply to: [441642355](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441642355) [](ancestors = 441642355)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Yes. Why not?

---
In reply to: [441640064](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441640064) [](ancestors = 441640064)

source/LottieToWinComp/LottieToWinCompTranslator.cs
raymond-wh-leung: Ah ok didn't realize that.

---
In reply to: [441659405](https://github.com/windows-toolkit/Lottie-Windows/pull/294#discussion_r441659405) [](ancestors = 441659405,441642355)

PR #297, opened 2020-06-17
Remove the ability to output XML from LottieGen.

simeoncran:
The XML outputs aren't being used by anyone AFAIK. They have a maintenance cost and they unnecessarily increase the number of files we generate during testing which slows down testing.
So I'm removing XML completely. For WinComp the C# outputs are higher fidelity and easier to read. For Lottie the YAML outputs are much easier to read.

These old XML outputs were useful in the early development of this project but they have no value now.


source/WinCompData/WinCompData.projitems
```diff
@@ -94,4 +93,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Wui\Color.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Wui\Colors.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="$(MSBuildThisFileDirectory)Serialization\" />
```
derobinsMSFT, line 97: >$(MSBuildThisFileDirectory)Serialization\ [](start = 21, length = 41)

Why did this get added? It looks like you've removed this directory altogether (CompositionObjectXmlSerializer.cs was the only file in this directory).

simeoncran: That's VS being helpful.

---
In reply to: [442341613](https://github.com/windows-toolkit/Lottie-Windows/pull/297#discussion_r442341613) [](ancestors = 442341613)

PR #298, opened 2020-06-22
Ignore the "meta" field in Lottie files.

simeoncran:
Previously we would complain that this field is unexpected. It's an object added by the LottieFiles.com [plugin](https://lottiefiles.com/plugins/after-effects)/ We don't do anything interesting with it yet, but at least we shouldn't frighten the user with a warning.


source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -186,7 +186,20 @@ LottieComposition ParseLottieComposition(ref Reader reader)
                                 // Treat any other property as an extension of the BodyMovin format.
                                 default:
                                     {
-                                        _issues.UnexpectedField(currentProperty);
+                                        // Report the extension as an issue, unless it is well-known
```
derobinsMSFT, line 189: >is  [](start = 87, length = 3)

"is a"

PR #299, opened 2020-06-22
Update the version of nuget.

simeoncran:
Necessary to start consuming cswinrt.


azure-pipelines.yml
```diff
@@ -1,3 +1,4 @@
+
 trigger:
```
derobinsMSFT, line 2: Was this newline added on purpose?

PR #305, opened 2020-06-30
Stop using CompositionRectangleGeometry

simeoncran:
CompositionRectangleGeoemetry doesn't anti-alias correctly. Rounded rectangles with tiny corner radiuses work fine, so use them instead.

Rectangles are expected to be fixed eventually, and when they are we can take advantage of their better (than rounded) perf, but until then they aren't reliable.

### DIFF
--- a/source/LottieToWinComp/CompositionObjectFactory.cs
+++ b/source/LottieToWinComp/CompositionObjectFactory.cs
@@ -5,7 +5,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Reflection.Metadata.Ecma335;
 using Microsoft.Toolkit.Uwp.UI.Lottie.LottieData;
 using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData;
 using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Mgce;
--- a/source/LottieToWinComp/CompositionObjectFactory.cs
+++ b/source/LottieToWinComp/CompositionObjectFactory.cs
@@ -112,16 +112,19 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 roundedRectangleGeometry.CornerRadius = new Sn.Vector2(0.000001F);
                 roundedRectangleGeometry.Size = size;
                 roundedRectangleGeometry.Offset = offset;
+
                 result = roundedRectangleGeometry;
             }
             else
             {
                 // Later versions do not need the rounded rectangle workaround.
+                ConsumeVersionFeature(c_rectangleGeometryIsUnreliableUntil);
+
                 var rectangleGeometry = _compositor.CreateRectangleGeometry();
                 rectangleGeometry.Size = size;
                 rectangleGeometry.Offset = offset;
+
                 result = rectangleGeometry;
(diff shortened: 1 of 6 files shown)


source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -94,12 +95,36 @@ internal uint GetUapVersionForApi(string apiName)
...
+                rectangleGeometry.Size = size;
+                rectangleGeometry.Offset = offset;
+                result = rectangleGeometry;
+                ConsumeVersionFeature(c_rectangleGeometryIsUnreliableUntil);
+            }
+
+            return result;
         }
 
         internal CompositionRoundedRectangleGeometry CreateRoundedRectangleGeometry() => _compositor.CreateRoundedRectangleGeometry();
```
derobinsMSFT, line 130: >CreateRoundedRectangleGeometry [](start = 53, length = 30)

This doesn't appear to be in use anymore.

source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -94,12 +95,36 @@ internal uint GetUapVersionForApi(string apiName)
...
+                // <= V7 did not reliably draw non-rounded rectangles.
+                // <= V11 draws non-rounded rectangles with aliased edges.
+                // Work around the problem by using a rounded rectangle with a tiny corner radius.
+                var roundedRectangleGeometry = _compositor.CreateRoundedRectangleGeometry();
+
+                // NOTE: magic tiny corner radius number - do not change!
+                roundedRectangleGeometry.CornerRadius = new Sn.Vector2(0.000001F);
+                roundedRectangleGeometry.Size = size;
+                roundedRectangleGeometry.Offset = offset;
+                result = roundedRectangleGeometry;
```
derobinsMSFT, line 115: >result = roundedRectangleGeometry; [](start = 16, length = 34)

Nit: add an extra newline above this line, just to help with readability.

source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -94,12 +95,36 @@ internal uint GetUapVersionForApi(string apiName)
...
+            }
+            else
+            {
+                // Later versions do not need the rounded rectangle workaround.
+                var rectangleGeometry = _compositor.CreateRectangleGeometry();
+                rectangleGeometry.Size = size;
+                rectangleGeometry.Offset = offset;
+                result = rectangleGeometry;
```
derobinsMSFT, line 123: >result = rectangleGeometry; [](start = 16, length = 27)

Nit: extra newline above this line as well.

source/LottieToWinComp/CompositionObjectFactory.cs
simeoncran: What made you think that?

---
In reply to: [447948768](https://github.com/windows-toolkit/Lottie-Windows/pull/305#discussion_r447948768) [](ancestors = 447948768)

source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Reflection.Metadata.Ecma335;
```
derobinsMSFT, line 8: Not needed-- remove.

PR #306, opened 2020-07-01
Update player sample page

### LINKED ISSUE #304: Tracking issue: Replace Lottie-Windows sample with new XAML Controls Gallery Sample
This is a tracking issue for [this](https://github.com/microsoft/Xaml-Controls-Gallery/issues/441#issuecomment-651419807) (by @simeoncran)

Once https://github.com/microsoft/Xaml-Controls-Gallery/pull/480 is finished and approved, I'll port the changes over here.

chingucoding:
As per #304 updating the player sample to behave more naturally.

Closes #304

PS: The startup animation of the sample app is awesome!

### DIFF
--- a/samples/LottieSamples/Scenarios/PlaybackPage.xaml
+++ b/samples/LottieSamples/Scenarios/PlaybackPage.xaml
@@ -35,46 +35,45 @@
 
             <!--  Playback Buttons  -->
             <Grid Width="400"
-                  VerticalAlignment="Center">
+                      VerticalAlignment="Center">
                 <Grid.ColumnDefinitions>
                     <ColumnDefinition />
                     <ColumnDefinition />
                     <ColumnDefinition />
-                    <ColumnDefinition />
                 </Grid.ColumnDefinitions>
-                <Grid.Resources>
-                    <Style x:Key="ButtonBaseStyle" TargetType="ButtonBase">
-                        <Setter Property="HorizontalAlignment" Value="Stretch"/>
-                        <Setter Property="Margin" Value="10"/>
-                    </Style>
-                    <Style TargetType="Button" BasedOn="{StaticResource ButtonBaseStyle}"/>
-                    <Style TargetType="ToggleButton" BasedOn="{StaticResource ButtonBaseStyle}"/>
-                </Grid.Resources>
-                <!--  Play  -->
-                <Button x:Name="PlayButton"
-                        Grid.Column="0"
-                        Click="PlayButton_Click">
-                    <SymbolIcon Symbol="Play" />
+
+                <!-- Play/Pause -->
+                <Button x:Name="PlayPauseButton"
+                                Grid.Column="0"
+                                Margin="10"
+                                HorizontalAlignment="Stretch"
+                                Click="PlayPauseButton_Click">
+                    <StackPanel>
+                        <SymbolIcon x:Name="PlayIcon" Symbol="Play"/>
+                        <SymbolIcon x:Name="PauseIcon" Symbol="Pause" Visibility="Collapsed"/>
+                    </StackPanel>
                 </Button>
-                <!--  Pause  -->
-                <ToggleButton x:Name="PauseButton"
-                              Grid.Column="1"
-                              Checked="PauseButton_Checked"
-                              IsThreeState="False"
-                              Unchecked="PauseButton_Unchecked">
-                    <SymbolIcon Symbol="Pause" />
-                </ToggleButton>
-                <!--  Resume  -->
+
+                <!--  Stop/Reset -->
                 <Button x:Name="StopButton"
-                        Grid.Column="2"
-                        Click="StopButton_Click">
-                    <SymbolIcon Symbol="Stop" />
+                            Grid.Column="1"
+                            Margin="10"
+                            HorizontalAlignment="Stretch"
+                            Click="StopButton_Click"
+                            ToolTipService.ToolTip="Stop"
+                            AutomationProperties.Name="Stop">
+                    <SymbolIcon Symbol="Previous" />
                 </Button>
+
                 <!--  Reverse  -->
                 <Button x:Name="ReverseButton"
-                        Grid.Column="3"
-                        Click="ReverseButton_Click">
-                    <SymbolIcon Symbol="Previous" />
+                            Grid.Column="2"
+                            Margin="10"
+                            HorizontalAlignment="Stretch"
+                            Click="ReverseButton_Click"
+                            ToolTipService.ToolTip="Reverse"
+                            AutomationProperties.Name="Reverse">
+                    <FontIcon Glyph="&#xEB9E;" />
                 </Button>
             </Grid>
 
--- a/samples/LottieSamples/Scenarios/PlaybackPage.xaml
+++ b/samples/LottieSamples/Scenarios/PlaybackPage.xaml
@@ -18,7 +18,7 @@
                     playback in response to user interaction.
                     <Span FontWeight="SemiBold">Press</Span>
                     the buttons below to
-                    play, pause, stop, and reverse the Lottie animation.
+                    play/pause, stop, and switch the play direction (either forwards or backwards).
                     <LineBreak />
                 </Paragraph>
             </RichTextBlock>
@@ -54,7 +54,7 @@
                     </StackPanel>
                 </Button>
 
-                <!--  Stop/Reset -->
+                <!-- Reset -->
                 <Button x:Name="StopButton"
                             Grid.Column="1"
                             Margin="10"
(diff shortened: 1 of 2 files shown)


samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
+using System;
 using Windows.UI.Xaml;
+using Windows.UI.Xaml.Automation;
 using Windows.UI.Xaml.Controls;
 
 namespace LottieSamples.Scenarios
 {
     public sealed partial class PlaybackPage : Page
     {
+        private bool wasPaused = false;
```
simeoncran, line 14: >false [](start = 33, length = 5)

Remove unnecessary initializations.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
-            // Resume playing current animation.
-            Playback_Player.Resume();
-        }
-
         private void StopButton_Click(object sender, RoutedEventArgs e)
         {
             // Stop the animation, which completes PlayAsync and resets to initial frame. 
             Playback_Player.Stop();
-            PauseButton.IsChecked = false;
+            Playback_Player.SetProgress(0);
```
simeoncran, line 25: Why is this necessary? Stopping should go back to 0 progress automatically.

samples/LottieSamples/Scenarios/PlaybackPage.xaml
```diff
@@ -35,46 +35,45 @@
...
-                              Grid.Column="1"
-                              Checked="PauseButton_Checked"
-                              IsThreeState="False"
-                              Unchecked="PauseButton_Unchecked">
-                    <SymbolIcon Symbol="Pause" />
-                </ToggleButton>
-                <!--  Resume  -->
+
+                <!--  Stop/Reset -->
                 <Button x:Name="StopButton"
```
simeoncran, line 58: The icon doesn't look like a stop button. The naming should reflect the intention. Reset maybe?

simeoncran:
>                     play, pause, stop, and reverse the Lottie animation.

Update the text to reflect what the code now does.

---
Refers to: samples/LottieSamples/Scenarios/PlaybackPage.xaml:21 in 878feaf. [](commit_id = 878feaf102e25b9cd0e263b955b5e28cd41b82df, deletion_comment = False)

samples/LottieSamples/Scenarios/PlaybackPage.xaml
```diff
@@ -35,46 +35,45 @@
...
-                        Grid.Column="3"
-                        Click="ReverseButton_Click">
-                    <SymbolIcon Symbol="Previous" />
+                            Grid.Column="2"
+                            Margin="10"
+                            HorizontalAlignment="Stretch"
+                            Click="ReverseButton_Click"
+                            ToolTipService.ToolTip="Reverse"
+                            AutomationProperties.Name="Reverse">
+                    <FontIcon Glyph="&#xEB9E;" />
```
simeoncran, line 76: This icon makes me think the the animation will play quickly in reverse when I click it, but it doesn't quite work like that. It's non-intuitive to me. It's like the play button in that it starts things playing, but now that the play button toggles between pause and play, this button feels weird because it doesn't have the toggling ability. Maybe it could be a toggle with an arrow that faces forward or backward.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
+
+        private async void StartAnimation()
+        {
+            if (!Playback_Player.IsPlaying)
+            {
+                // Play the animation at the currently specified playback rate.
+                Playback_Player.PlaybackRate = 1;
+                SetIsPlayingIndicator(true);
+                await Playback_Player.PlayAsync(fromProgress: 0, toProgress: 1, looped: false);
+                SetIsPlayingIndicator(false);
```
simeoncran, line 46: > SetIsPlayingIndicator(false); [](start = 15, length = 30)

You might get here while the player is playing. If the player is stopped then started again, there is the possibility that this statement will execute while the second play is underway.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
+            if (Playback_Player.IsPlaying && !wasPaused)
             {
-                return;
+                Playback_Player.Pause();
+                SetIsPlayingIndicator(false);
+                wasPaused = true;
             }
+            else
+            {
+                // Either not playing or paused
```
simeoncran, line 61: Comments should be terminated with periods.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
+                // If paused, resume and set paused to false
+                if (wasPaused)
+                {
+                    Playback_Player.Resume();
+                    SetIsPlayingIndicator(true);
+                }
+                // Not playing, start animation now
```
simeoncran, line 68: >// Not playing, start animation now [](start = 16, length = 35)

Comments that describe a state should be placed inside blocks where the state is true. In this case, move it into the "else" block.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
simeoncran, line 14: >wasPaused [](start = 21, length = 9)

This name is confusing me a lot. If you can have an _isPaused field instead it would be clearer to me. Otherwise I don't know whether this means "it has ever been paused", or "it's not currently paused", or "it was paused, but n state changes happened since", etc.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
         private void ReverseButton_Click(object sender, RoutedEventArgs e)
         {
             // Set reverse playback rate.
             // NOTE: This property is live, which means it takes effect even if the animation is playing.
             Playback_Player.PlaybackRate = -1;
-            StartPlaying();
+            StartAnimation();
+        }
+
+        private async void StartAnimation()
```
simeoncran, line 38: Rename to StartPlay() because the player has a "play" concept rather than an "animation" concept. Animation makes me think of some keyframe animation on the page.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
```diff
@@ -2,71 +2,93 @@
...
+                // Play the animation at the currently specified playback rate.
+                Playback_Player.PlaybackRate = 1;
+                SetIsPlayingIndicator(true);
+                await Playback_Player.PlayAsync(fromProgress: 0, toProgress: 1, looped: false);
+                SetIsPlayingIndicator(false);
+            }
         }
 
-        private void StartPlaying()
+        private void PlayPauseButton_Click(object sender, RoutedEventArgs e)
```
simeoncran, line 50: Please add a comment to the method explaining the three states and what is supposed to happen when this is called for each state.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: No idea why this was in the sample, removed it now.

samples/LottieSamples/Scenarios/PlaybackPage.xaml
chingucoding: I've updated the icon to be a stop button since this is a common icon to use in this scenario and is a "learned" behavior.

samples/LottieSamples/Scenarios/PlaybackPage.xaml
chingucoding: I've updated the button to toggle between an arrow pointing to the left or right depending on the playback rate.

The question is if the wording I chose is good here: "(playback) direction" What do you think @simeoncran ?

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: I've added a check that verifies that the player is not playing anymore after the await, Is that the right approach here?

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: Fixed

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: Fixed

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: GOod point, updated the variable name to `_isPaused`.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: Renamed the function.

samples/LottieSamples/Scenarios/PlaybackPage.xaml.cs
chingucoding: Added a comment explaining what the idea of the function is.

chingucoding:
@stmoy There have been quite a few improvements on this sample, would you like me to port them back to the XCG once this PR gets merged?

simeoncran:
I'm rejecting this PR now, but @chingucoding you should definitely push back the improvements from the CR to XCG.
The reasons for rejecting here are that, as much as this makes the playing sample more interesting, it also makes it more complicated and that replaces some concepts that are fundamental to the Lottie API with higher level concepts. Also, having discussed this with @stmoy, I've come to the conclusion that it's actually good to have XCG diverge from the samples here. XCG enables discovery and inspiration, whereas the samples here are about getting people started on the fundamental concepts.
Apologies @chingucoding  for encouraging you down this path then switching on you, but it took me some time to realize how this should fit.

chingucoding:
All right, thanks for the explanation @simeoncran . At least, we can use the improved sample in the XCG :)

PR #307, opened 2020-07-01
Ignore the "ml2" field in Lottie files.

simeoncran:
It's a mystery as to what this field does. It appears to be related to miter limit, but that's already handled by "ml". Some tool is adding it, although BodyMovin importer doesn't seem to know what it is.

### DIFF
--- a/source/LottieReader/Serialization/ShapeLayerContents.cs
+++ b/source/LottieReader/Serialization/ShapeLayerContents.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
         {
             // Not clear whether we need to read these properties.
             // "ml2" is some sort of extra miter limit value that does not seem to be supported by
-            // BodyMovin. It's a mystery as to what it means or how hit is getting into the file,
+            // BodyMovin. It's a mystery as to what it means or how it is getting into the file,
             // but quite a few files have it.
             obj.IgnorePropertyThatIsNotYetSupported("fillEnabled", "hd", "ml2");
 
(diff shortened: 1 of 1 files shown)


source/LottieReader/Serialization/ShapeLayerContents.cs
```diff
@@ -85,7 +85,10 @@ ShapeLayerContent ReadShapeContent(in LottieJsonObjectElement obj)
             in ShapeLayerContent.ShapeLayerContentArgs shapeLayerContentArgs)
         {
             // Not clear whether we need to read these properties.
-            obj.IgnorePropertyThatIsNotYetSupported("fillEnabled", "hd");
+            // "ml2" is some sort of extra miter limit value that does not seem to be supported by
+            // BodyMovin. It's a mystery as to what it means or how hit is getting into the file,
```
derobinsMSFT, line 89: >hit [](start = 68, length = 3)

"it"

source/LottieReader/Serialization/ShapeLayerContents.cs
```diff
@@ -85,7 +85,10 @@ ShapeLayerContent ReadShapeContent(in LottieJsonObjectElement obj)
...
+            // but quite a few files have it.
+            obj.IgnorePropertyThatIsNotYetSupported("fillEnabled", "hd", "ml2");
```
derobinsMSFT, line 91: >"fillEnabled", "hd" [](start = 52, length = 19)

Seeing as you commented on why "ml2" is in this list, why not also add comments for these properties?

simeoncran: I don't have any more information about those. I did a bit of an investigation about ml2 so I wanted to capture what I'd learnt.

---
In reply to: [449296854](https://github.com/windows-toolkit/Lottie-Windows/pull/307#discussion_r449296854) [](ancestors = 449296854)

PR #309, opened 2020-07-07
Add ability to automatically bind to the colors in a Lottie's palette.

simeoncran:
When enabled with -GenerateColorBindings, LottieGen will create bindings
for each non-animated color it finds in a fill or stroke. This will
allow the colors to be changed at runtime.

This is particularly handy for animations that have a single color such
as those found at https://www.lottieflow.com.

The feature is less flexible than theme property bindings because:
a) it only supports colors
b) it doesn't give the designer much control over which parts of the
animation are bindable.
c) it doesn't allow the designer to choose the names used for the
bindings.

The binding name is based on the color, so a black color will generate a property
called "ColorBlack".

For single-color Lottie files we could be clever and name the property
"Foreground". I'm not doing that right now because we don't yet know
how the feature will be consumed.

### DIFF
--- a/LottieGen/CommandLineOptions.cs
+++ b/LottieGen/CommandLineOptions.cs
@@ -35,6 +35,8 @@ sealed class CommandLineOptions
     // The error should be a sentence (starts with a capital letter, and ends with a period).
     internal string ErrorDescription { get; private set; }
 
+    internal bool GenerateColorBindings { get; private set; }
+
     internal bool GenerateDependencyObject { get; private set; }
 
     internal bool HelpRequested { get; private set; }
@@ -83,6 +85,11 @@ sealed class CommandLineOptions
         if (DisableTranslationOptimizer)
         {
             sb.Append($" -{nameof(DisableTranslationOptimizer)}");
+        }
+
+        if (GenerateColorBindings)
+        {
+            sb.Append($" -{nameof(GenerateColorBindings)}");
         }
 
         if (GenerateDependencyObject)
@@ -134,6 +141,7 @@ sealed class CommandLineOptions
         Ambiguous,
         DisableCodeGenOptimizer,
         DisableTranslationOptimizer,
+        GenerateColorBindings,
         GenerateDependencyObject,
         Help,
         InputFile,
@@ -156,14 +164,14 @@ sealed class CommandLineOptions
 
         // Convert the language strings to language values.
         var languageTokenizer = new CommandlineTokenizer<Lang>(Lang.Ambiguous)
-                .AddKeyword("csharp", Lang.CSharp)
-                .AddKeyword("cppcx", Lang.Cx)
-                .AddKeyword("cx", Lang.Cx)
-                .AddKeyword("cppwinrt", Lang.Cppwinrt)
-                .AddKeyword("winrtcpp", Lang.Cppwinrt)
-                .AddKeyword("lottieyaml", Lang.LottieYaml)
-                .AddKeyword("dgml", Lang.WinCompDgml)
-                .AddKeyword("stats", Lang.Stats);
+                .AddKeyword(Lang.CSharp)
+                .AddKeyword(Lang.Cx, "cppcx")
+                .AddKeyword(Lang.Cx)
+                .AddKeyword(Lang.Cppwinrt)
+                .AddKeyword(Lang.Cppwinrt, "winrtcpp")
+                .AddKeyword(Lang.LottieYaml)
+                .AddKeyword(Lang.WinCompDgml, "dgml")
+                .AddKeyword(Lang.Stats);
 
         var languages = new List<Lang>();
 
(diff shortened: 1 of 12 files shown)


source/LottieToWinComp/LottieToMultiVersionWinCompTranslator.cs
```diff
@@ -23,22 +23,19 @@ sealed class LottieToMultiVersionWinCompTranslator
...
         /// <returns>The results of the translation and the issues.</returns>
         public static MultiVersionTranslationResult TryTranslateLottieComposition(
             LottieComposition lottieComposition,
+            TranslationOptions options,
             uint targetUapVersion,
             uint minimumUapVersion,
-            bool strictTranslation,
-            bool addCodegenDescriptions,
-            bool translatePropertyBindings)
+            bool strictTranslation)
```
derobinsMSFT, line 38: >bool strictTranslation [](start = 12, length = 22)

Why is this not in `TranslationOptions`? Seems odd to have this stand out when there's now a dedicated struct.

simeoncran: The others determine what the translator produces. "strictTranslation" determines whether the method will fail when there are difficulties. One says "this is what I want to make", the other says "this is what I want you to do if you can't make it the way I want".
I think "strictTranslation" is likely to disappear. It made sense as we were bringing the code up, but nobody has a use for it now.

---
In reply to: [451190382](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451190382) [](ancestors = 451190382)

simeoncran: Actually, now that you make me think about it, the versions should probably be part of the options.

---
In reply to: [451203341](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451203341) [](ancestors = 451203341,451190382)

simeoncran: Only the target one could be added. The minimum version is specific to the multi-version translator only.

---
In reply to: [451215511](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451215511) [](ancestors = 451215511,451203341,451190382)

simeoncran: Argh... maybe that's a mistake. It's not really an option because it's not a bool. What do you think?

---
In reply to: [451223350](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451223350) [](ancestors = 451223350,451215511,451203341,451190382)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3018,9 +3043,15 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
...
+                    var trimmed = context.TrimAnimatable(animatable);
                     var propertyName = name;
-                    result.Properties.InsertScalar(propertyName, Opacity(animatable.InitialValue));
-                    ApplyOpacityKeyFrameAnimation(context, context.TrimAnimatable(animatable), result.Properties, propertyName, propertyName, null);
+                    result.Properties.InsertScalar(propertyName, Opacity(trimmed.InitialValue));
+
+                    // The opacity is animated, but it might be non-animated after trimming.
+                    if (trimmed.IsAnimated)
+                    {
+                        ApplyOpacityKeyFrameAnimation(context, context.TrimAnimatable(animatable), result.Properties, propertyName, propertyName, null);
```
derobinsMSFT, line 3053: >context.TrimAnimatable(animatable) [](start = 63, length = 34)

`trimmed`

source/LottieToWinComp/LottieToMultiVersionWinCompTranslator.cs
derobinsMSFT: I would revert it back. It seems like `TranslationOptions` was meant to bring together a set of bool translation options, in particular; that's part of why `strictTranslation` not being included caught my eye. It also feels odd to have the target version in `TranslationOptions` but not the minimum version (where that gets used).

---
In reply to: [451226153](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451226153) [](ancestors = 451226153,451223350,451215511,451203341,451190382)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Good catch.

---
In reply to: [451233222](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451233222) [](ancestors = 451233222)

source/LottieToWinComp/LottieToMultiVersionWinCompTranslator.cs
simeoncran: Minimum version is not a public-facing parameter - that's why it's separated. I'll try a different name for this struct and see how that looks.

---
In reply to: [451238264](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451238264) [](ancestors = 451238264,451226153,451223350,451215511,451203341,451190382)

derobinsMSFT: I think the name change helps quite a bit-- along with removing `strictTranslation`.

---
In reply to: [451662312](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r451662312) [](ancestors = 451662312,451238264,451226153,451223350,451215511,451203341,451190382)

source/LottieToWinComp/LottieToMultiVersionWinCompTranslator.cs
```diff
@@ -128,19 +112,23 @@ sealed class LottieToMultiVersionWinCompTranslator
             // that satisfies minimumUapVersion.
             while (translationResult.MinimumRequiredUapVersion > minimumUapVersion)
             {
-                var nextLowerTarget = translationResult.MinimumRequiredUapVersion - 1;
+                // Copy the configuration but change the target version to the next value less than
+                // the version supported by the previous translation.
+                var nextLowerTargetConfiguration = configuration;
```
derobinsMSFT, line 117: >var nextLowerTargetConfiguration = configuration; [](start = 16, length = 49)

Nit: move this out of the loop? It only needs to be set to the original `configuration` once. Not sure if the C# compiler optimizes such that `nextLowerTargetConfiguration` is only initialized once anyway.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3021,16 +3048,17 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
                 anim.SetReferenceParameter(ThemePropertiesName, _themePropertySet);
 
                 StartExpressionAnimation(result, nameof(result.Color), anim);
-                return result;
             }
             else
             {
-                // Opacity isn't animated. Multiply the alpha channel of the color by the non-animated opacity value.
+                // Opacity isn't animated.
+                // Create an expression that multiples the alpha channel of the color by the opacity value.
```
derobinsMSFT, line 3055: >multiples [](start = 45, length = 9)

"multiplies"?

LottieGen/CommandLineOptions.cs
```diff
@@ -192,21 +200,22 @@ void ParseCommandLineStrings(string[] args)
...
-            .AddPrefixedKeyword("interface", Keyword.Interface)
-            .AddPrefixedKeyword("language", Keyword.Language)
-            .AddPrefixedKeyword("minimumuapversion", Keyword.MinimumUapVersion)
-            .AddPrefixedKeyword("namespace", Keyword.Namespace)
-            .AddPrefixedKeyword("outputfolder", Keyword.OutputFolder)
-            .AddPrefixedKeyword("public", Keyword.Public)
-            .AddPrefixedKeyword("strict", Keyword.Strict)
-            .AddPrefixedKeyword("targetuapversion", Keyword.TargetUapVersion)
-            .AddPrefixedKeyword("testmode", Keyword.TestMode);
+            .AddPrefixedKeyword(Keyword.Help, "?")
```
derobinsMSFT, line 203: >.AddPrefixedKeyword(Keyword.Help, "?") [](start = 12, length = 38)

Nit: alphabetize.

source/LottieToWinComp/LottieToMultiVersionWinCompTranslator.cs
simeoncran: I did it this way on purpose. It's true that it will copy the 8-12 bytes each time through the loop, but I'm not concerned about that cost here. In fact it will do that again when it makes the call to TryTranslateLottieComposition (which reminds me - I'll convert that to an "in" if possible to save that unnecessary cost). It's nice to have it inside the loop because it makes it clear that we are starting fresh with a copy of "configuration" then modifying it slightly. If it's outside the loop I'd have to read the whole loop to know what state it is in. 

---
In reply to: [452467907](https://github.com/windows-toolkit/Lottie-Windows/pull/309#discussion_r452467907) [](ancestors = 452467907)

PR #312, opened 2020-07-11
Support animated vector2/vector3 with easings that differ between channels

simeoncran:
Lottie allows Vector2/Vector3 to have different easings for each channel. When that happens we cannot translate
them as Vector2/Vector3 animations, because WinComp only supports an easing for the whole Vector2/Vector3.
Now we will recognize this case and first translate to the AnimatableVector3 to an AnimatableVectorXYZ and
create separate animations for each channel.

This change only supports separate easings in that cases that we've seen in the corpus. Complete support is
a lot more expensive code (i.e. expensive to maintain because it requires a lot of repeating of the same
patterns over and over with slight changes).
Current support includes:
* Size of a shape
* Size of a rectangle
* Radius of an ellipse
* Position of a shape

### DIFF
--- a/source/LottieToWinComp/AnimatableVector3Rewriter.cs
+++ b/source/LottieToWinComp/AnimatableVector3Rewriter.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         {
             // CubicBezierEasings can have multiple easings - one for each channel (X, Y, Z).
             // See if there are any CubicBezierEasings that have different easing for
-            // each channel. It there aren't any, there's nothing to do.
+            // each channel. If there aren't any, there's nothing to do.
             if (!HasMultipleEasings(animatableVector3))
             {
                 return animatableVector3;
@@ -39,6 +39,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             var xKeyFrames = ExtractKeyFrames(animatableVector3.KeyFrames, v => v.X, 0);
             var yKeyFrames = ExtractKeyFrames(animatableVector3.KeyFrames, v => v.Y, 1);
             var zKeyFrames = ExtractKeyFrames(animatableVector3.KeyFrames, v => v.Z, 2);
+
             return new AnimatableXYZ(
                 new Animatable<double>(xKeyFrames, animatableVector3.PropertyIndex),
                 new Animatable<double>(yKeyFrames, animatableVector3.PropertyIndex),
@@ -48,19 +49,27 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         // Extracts the key frames from a single channel of a Vector3.
         static KeyFrame<double>[] ExtractKeyFrames(
             ReadOnlySpan<KeyFrame<Vector3>> keyFrames,
-            Func<Vector3, double> valueExtractor,
-            int valueIndex)
+            Func<Vector3, double> channelValueSelector,
+            int channelIndex)
         {
             var result = new KeyFrame<double>[keyFrames.Length];
 
             for (var i = 0; i < keyFrames.Length; i++)
             {
                 var kf = keyFrames[i];
-                var value = valueExtractor(kf.Value);
+                var value = channelValueSelector(kf.Value);
                 var easing = kf.Easing;
                 if (easing is CubicBezierEasing cubicBezierEasing)
                 {
-                    var bezier = cubicBezierEasing.Beziers.Count > valueIndex ? cubicBezierEasing.Beziers[valueIndex] : cubicBezierEasing.Beziers[0];
+                    // Get the bezier for the channel, if there is one, otherwise
+                    // use the easing from the X channel. Not all Vector3s have
+                    // multiple easings, and some are really representing Vector2s
+                    // and might have easings for X and Y but probably not for Z,
+                    // so for those cases the X easing will do well enough.
+                    var bezier = cubicBezierEasing.Beziers.Count > channelIndex
+                                        ? cubicBezierEasing.Beziers[channelIndex]
+                                        : cubicBezierEasing.Beziers[0];
+
                     easing = new CubicBezierEasing(new[] { bezier });
                 }
 
--- a/source/LottieToWinComp/AnimatableVector3Rewriter.cs
+++ b/source/LottieToWinComp/AnimatableVector3Rewriter.cs
@@ -3,8 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Diagnostics;
-using System.Linq;
 using Microsoft.Toolkit.Uwp.UI.Lottie.LottieData;
 
 namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
@@ -13,21 +11,20 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
     /// Takes an <see cref="AnimatableVector3"/> and if the easings on the X, Y, or Z channels
     /// are different, returns an equivalent <see cref="AnimatableXYZ"/>.
     /// </summary>
-    static class AnimatableVector3ToAnimatableXYZ
+    static class AnimatableVector3Rewriter
     {
-        // Asserts that the given AnimatableVector3 has the same easing for each channel.
-        internal static void AssertSingleEasing(IAnimatableVector3 value)
-        {
-            if (value is AnimatableVector3 animatableVector3 && HasMultipleEasings(animatableVector3))
-            {
-                Debug.Fail("Multiple easings!");
-            }
-        }
+        /// <summary>
+        /// If the given <see cref="IAnimatableVector3"/> is an <see cref="AnimatableVector3"/> with multiple
+        /// easings, returns an equivalent <see cref="AnimatableXYZ"/> with one easing per channel.
+        /// </summary>
+        /// <returns>An equivalent <see cref="IAnimatableVector3"/> with only one easing per
+        /// channel.</returns>
+        internal static IAnimatableVector3 EnsureOneEasingPerChannel(IAnimatableVector3 animatableVector3) =>
+            animatableVector3 is AnimatableVector3 value ? EnsureOneEasingPerChannel(value) : animatableVector3;
 
-        internal static IAnimatableVector3 SeparateEasings(IAnimatableVector3 animatableVector3) =>
-            animatableVector3 is AnimatableVector3 value ? SeparateEasings(value) : animatableVector3;
-
-        internal static IAnimatableVector3 SeparateEasings(AnimatableVector3 animatableVector3)
+        // If the given AnimatableVector3 has multiple easings, returns an equivalent
+        // AnimatableXYZ with one easing per channel, otherwise returns the given AnimatableVector3.
+        static IAnimatableVector3 EnsureOneEasingPerChannel(AnimatableVector3 animatableVector3)
         {
             // CubicBezierEasings can have multiple easings - one for each channel (X, Y, Z).
             // See if there are any CubicBezierEasings that have different easing for
@@ -73,8 +70,8 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             return result;
         }
 
-        // Returns true iff the given AnimatableVector3 does not use the same easing for each
-        // of its channels (X, Y, Z).
+        // Returns true iff the key frames in the given AnimatableVector3 do not use the
+        // same easing for each of their channels (X, Y, Z).
         static bool HasMultipleEasings(AnimatableVector3 value)
         {
             if (value.IsAnimated)
@@ -83,24 +80,33 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 {
                     if (kf.Easing is CubicBezierEasing cubicBezierEasing)
                     {
+                        // Nothing to check if there is only one easing.
                         if (cubicBezierEasing.Beziers.Count > 1)
                         {
                             var xEasing = cubicBezierEasing.Beziers[0];
 
                             if (xEasing != cubicBezierEasing.Beziers[1])
                             {
+                                // The X and Y easings are different from each other.
                                 return true;
                             }
 
+                            // The X easing and the Y easing are the same.
+                            // If there's a Z easing, check whether it's the same too.
                             if (cubicBezierEasing.Beziers.Count > 2 && xEasing != cubicBezierEasing.Beziers[2])
                             {
+                                // The Z easing is different from the X and Y easings.
                                 return true;
                             }
                         }
+
+                        // The easings for each channel in that key frame are the same. Keep
+                        // looking until a key frame is found where the easings are not the same.
                     }
                 }
             }
 
+            // The easings for each channel are the same in each key frame.
             return false;
         }
     }
--- a/source/LottieToWinComp/AnimatableVector3Rewriter.cs
+++ b/source/LottieToWinComp/AnimatableVector3Rewriter.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 var easing = kf.Easing;
                 if (easing is CubicBezierEasing cubicBezierEasing)
                 {
-                    // Get the bezier for the channel, if there is one, otherwise
+                    // Get the Bezier for the channel, if there is one, otherwise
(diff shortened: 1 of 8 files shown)


source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Linq;
+using Microsoft.Toolkit.Uwp.UI.Lottie.LottieData;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
+{
+    /// <summary>
+    /// Takes an <see cref="AnimatableVector3"/> and if the easings on the X, Y, or Z channels
+    /// are different, returns an equivalent <see cref="AnimatableXYZ"/>.
+    /// </summary>
+    static class AnimatableVector3ToAnimatableXYZ
```
derobinsMSFT, line 16: >static class [](start = 4, length = 12)

Add `internal`?

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
```diff
@@ -0,0 +1,107 @@
...
+            var yKeyFrames = ExtractKeyFrames(animatableVector3.KeyFrames, v => v.Y, 1);
+            var zKeyFrames = ExtractKeyFrames(animatableVector3.KeyFrames, v => v.Z, 2);
+            return new AnimatableXYZ(
+                new Animatable<double>(xKeyFrames, animatableVector3.PropertyIndex),
+                new Animatable<double>(yKeyFrames, animatableVector3.PropertyIndex),
+                new Animatable<double>(zKeyFrames, animatableVector3.PropertyIndex));
+        }
+
+        // Extracts the key frames from a single channel of a Vector3.
+        static KeyFrame<double>[] ExtractKeyFrames(
```
derobinsMSFT, line 52: >static [](start = 8, length = 6)

Add `private`?

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
derobinsMSFT, line 16: >AnimatableVector3ToAnimatableXYZ [](start = 17, length = 32)

There's really only one method in this class involved in creating an `AnimatableXYZ` from an `AnimatableVector3` (`SeparateEasings`). More generally, the methods in this class provide info on the easing functions of an `AnimatableVector3`, or change how those functions are represented.  I think the name of the class should reflect this (e.g., `AnimatableVector3Easings`), and the "to `AnimatableXYZ`" notion instead deferred to just the `SeparateEasings` method (e.g., `CreateAnimatableXYZWithSeparateEasings`).

In particular, the name of this class sticks out with the `AssertSingleEasing` call added in TranslationContext.cs-- that method in itself has nothing to do with the `AnimatableXYZ` class.

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
```diff
@@ -0,0 +1,107 @@
...
+
+                            if (xEasing != cubicBezierEasing.Beziers[1])
+                            {
+                                return true;
+                            }
+
+                            if (cubicBezierEasing.Beziers.Count > 2 && xEasing != cubicBezierEasing.Beziers[2])
+                            {
+                                return true;
+                            }
```
derobinsMSFT, line 98: Maybe add a comment here pointing out that at this point, we know the X- and Y-channel easings are the same, so there's no need to compare the Y- and Z-channel easings directly here. It took me a second look to realize why those channels aren't directly compared.

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
```diff
@@ -0,0 +1,107 @@
...
+            var result = new KeyFrame<double>[keyFrames.Length];
+
+            for (var i = 0; i < keyFrames.Length; i++)
+            {
+                var kf = keyFrames[i];
+                var value = valueExtractor(kf.Value);
+                var easing = kf.Easing;
+                if (easing is CubicBezierEasing cubicBezierEasing)
+                {
+                    var bezier = cubicBezierEasing.Beziers.Count > valueIndex ? cubicBezierEasing.Beziers[valueIndex] : cubicBezierEasing.Beziers[0];
```
derobinsMSFT, line 66: >cubicBezierEasing.Beziers.Count > valueIndex [](start = 33, length = 44)

If a keyframe has two cubic Bezier easings, should the Z channel use the easing of the X channel or Y channel? Does it/should it have an easing function at all in that case? Right now, the `AnimatableXYZ` will assign the X channel's easing to the Z channel if there are only two easings-- making sure this is correct/expected.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2199,14 +2199,37 @@ CompositionShape TranslateEllipseContent(TranslationContext context, ShapeConten
                 compositionEllipseGeometry.Center = Vector2(position.InitialValue);
             }
 
-            var diameter = context.TrimAnimatable(shapeContent.Diameter);
-            if (diameter.IsAnimated)
+            var diameter3OrXYZ = AnimatableVector3ToAnimatableXYZ.SeparateEasings(shapeContent.Diameter);
```
derobinsMSFT, line 2202: >diameter3OrXYZ [](start = 16, length = 14)

How about `diameterVector3OrXYZ`? (Use the full name of what comes after "Animatable" for clarity-- same elsewhere that uses this naming convention.)

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2417,6 +2520,96 @@ CompositionShape TranslateRectangleContent(TranslationContext context, ShapeCont
...
+                        }
+
+                        if (height.IsAnimated)
+                        {
+                            ApplyScalarKeyFrameAnimation(context, height, geometry, $"{nameof(Rectangle.Size)}.Y");
+                        }
+                    }
+                    else
+                    {
+                        // Only Position is animated
```
derobinsMSFT, line 2556: >animated [](start = 44, length = 8)

Period.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2417,6 +2520,96 @@ CompositionShape TranslateRectangleContent(TranslationContext context, ShapeCont
...
+
+            // TODO - this only works correctly if Size and TrimOffset are not animated. A complete solution requires
+            //        adding another property.
+            var isPartialTrimPath = shapeContext.TrimPath != null &&
+                (shapeContext.TrimPath.Start.IsAnimated || shapeContext.TrimPath.End.IsAnimated || shapeContext.TrimPath.Offset.IsAnimated ||
+                shapeContext.TrimPath.Start.InitialValue.Value != 0 || shapeContext.TrimPath.End.InitialValue.Value != 1);
+
+            if ((width.IsAnimated || height.IsAnimated) && isPartialTrimPath)
+            {
+                // Warn that we might be getting things wrong
```
derobinsMSFT, line 2591: >wrong [](start = 56, length = 5)

Period.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2286,54 +2323,114 @@ CompositionShape TranslateRectangleContent(TranslationContext context, ShapeCont
...
-                        StartExpressionAnimation(geometry, "CornerRadius", cornerRadiusExpression);
+                        geometry.Properties.InsertScalar("Roundness", Float(cornerRadius.InitialValue));
+                        ApplyScalarKeyFrameAnimation(context, cornerRadius, geometry.Properties, "Roundness");
+
+                        if (width.IsAnimated || height.IsAnimated)
+                        {
+                            // Both size and cornerRadius are animated.
+                            var cornerRadiusExpression = _c.CreateExpressionAnimation(ConstrainedCornerRadiusScalar());
+                            cornerRadiusExpression.SetReferenceParameter("my", geometry);
+                            StartExpressionAnimation(geometry, "CornerRadius", cornerRadiusExpression);
```
derobinsMSFT, line 2347: The only difference between this code and the two blocks below are the arguments to `ConstrainedCornerRadiusScalar`. Moving this to a separate function might help a little with code repetition.

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
```diff
@@ -0,0 +1,107 @@
...
+            if (value is AnimatableVector3 animatableVector3 && HasMultipleEasings(animatableVector3))
+            {
+                Debug.Fail("Multiple easings!");
+            }
+        }
+
+        internal static IAnimatableVector3 SeparateEasings(IAnimatableVector3 animatableVector3) =>
+            animatableVector3 is AnimatableVector3 value ? SeparateEasings(value) : animatableVector3;
+
+        internal static IAnimatableVector3 SeparateEasings(AnimatableVector3 animatableVector3)
```
derobinsMSFT, line 30: >SeparateEasings [](start = 43, length = 15)

This doesn't always separate the easings of my `AnimatableVector3`. The name should reflect this (following up from a previous comment, perhaps `CreateAnimatableXYZIfMultipleEasings`?).

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -3579,20 +3772,48 @@ CompositionSubGraph TranslateTextLayer(TranslationContext.For<TextLayer> context
             }
 
 #if !NoScaling
-            if (scalePercent.IsAnimated)
+            // If the channels have separate easings, convert to an AnimatableXYZ;
```
derobinsMSFT, line 3775: >; [](start = 81, length = 1)

Period.

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -2199,14 +2199,37 @@ CompositionShape TranslateEllipseContent(TranslationContext context, ShapeConten
...
-            if (diameter.IsAnimated)
+            var diameter3OrXYZ = AnimatableVector3ToAnimatableXYZ.SeparateEasings(shapeContent.Diameter);
+            if (diameter3OrXYZ is AnimatableXYZ diameterXYZ)
             {
-                ApplyScaledVector2KeyFrameAnimation(context, diameter, 0.5, compositionEllipseGeometry, "Radius");
+                var diameterX = context.TrimAnimatable(diameterXYZ.X);
+                var diameterY = context.TrimAnimatable(diameterXYZ.Y);
+                if (diameterX.IsAnimated)
+                {
+                    ApplyScaledScalarKeyFrameAnimation(context, diameterX, 0.5, compositionEllipseGeometry, "Radius.X", "Radius.X");
```
derobinsMSFT, line 2209: >"Radius.X", "Radius.X" [](start = 108, length = 22)

Are these intentionally identical?

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
simeoncran: The style for Lottie-Windows (except in samples) it to never use redundant visibility modifiers.

---
In reply to: [453816383](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r453816383) [](ancestors = 453816383)

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
simeoncran: The style for Lottie-Windows (except in samples) it to never use redundant visibility modifiers.

---
In reply to: [453817943](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r453817943) [](ancestors = 453817943)

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
simeoncran: I don't know for sure, but I think it won't matter in practice. We use IAnimatableVector3 for both Vector3 and Vector2 cases. For the most part Lottie only deals with Vector2 (because it is mostly a 2D system) but occasionally there are Vector3 values passed around. I wasn't able to predict when I would need a Vector2 vs a Vector3 in some cases, so I just use Vector2. In the translator we ignore the Z.
It's all a bit dirty, but the alternative (having an IAnimatableVector2/AnimatableVector2/AnimatableXY) would be even more code, so that seems worse.

---
In reply to: [453834204](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r453834204) [](ancestors = 453834204)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: That makes it harder for me to red. I think the trick is to just call it "diameter".

---
In reply to: [453835875](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r453835875) [](ancestors = 453835875)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Yes, but I'll make that clearer.

---
In reply to: [453844977](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r453844977) [](ancestors = 453844977)

source/LottieToWinComp/AnimatableVector3Rewriter.cs
```diff
@@ -0,0 +1,113 @@
...
+        internal static IAnimatableVector3 EnsureOneEasingPerChannel(IAnimatableVector3 animatableVector3) =>
+            animatableVector3 is AnimatableVector3 value ? EnsureOneEasingPerChannel(value) : animatableVector3;
+
+        // If the given AnimatableVector3 has multiple easings, returns an equivalent
+        // AnimatableXYZ with one easing per channel, otherwise returns the given AnimatableVector3.
+        static IAnimatableVector3 EnsureOneEasingPerChannel(AnimatableVector3 animatableVector3)
+        {
+            // CubicBezierEasings can have multiple easings - one for each channel (X, Y, Z).
+            // See if there are any CubicBezierEasings that have different easing for
+            // each channel. It there aren't any, there's nothing to do.
```
derobinsMSFT, line 31: >It [](start = 29, length = 2)

"If"

source/LottieToWinComp/LottieToWinCompTranslator.cs
```diff
@@ -4107,7 +4329,7 @@ static bool IsFuzzyEqual(in Vector2 a, in Vector2 b, in double acceptableError)
                 null,
                 targetObject,
                 targetPropertyName,
-                longDescription,
+                longDescription ?? targetPropertyName,
```
derobinsMSFT, line 4332: >longDescription [](start = 16, length = 15)

These descriptions are already checked for null in `GenericCreateCompositionKeyFrameAnimation` and assigned to `targetPropertyName` if null-- the check here isn't necessary.

source/LottieToWinComp/AnimatableVector3ToAnimatableXYZ.cs
derobinsMSFT: I think the facts that the translator ignores the Z easing, the easing is hard to predict, and/or the easing doesn't matter for currently-implemented cases merits a comment here.

---
In reply to: [453956394](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r453956394) [](ancestors = 453956394,453834204)

source/LottieToWinComp/LottieToWinCompTranslator.cs
simeoncran: Good catch. I don't know why I thought it was necessary to fix that here.

---
In reply to: [454536180](https://github.com/windows-toolkit/Lottie-Windows/pull/312#discussion_r454536180) [](ancestors = 454536180)

source/LottieToWinComp/AnimatableVector3Rewriter.cs
```diff
@@ -0,0 +1,122 @@
...
+            var result = new KeyFrame<double>[keyFrames.Length];
+
+            for (var i = 0; i < keyFrames.Length; i++)
+            {
+                var kf = keyFrames[i];
+                var value = channelValueSelector(kf.Value);
+                var easing = kf.Easing;
+                if (easing is CubicBezierEasing cubicBezierEasing)
+                {
+                    // Get the bezier for the channel, if there is one, otherwise
```
derobinsMSFT, line 64: >b [](start = 31, length = 1)

Nit: capitalize.

PR #315, opened 2020-07-16
Recognize animation subchannels when discovering non-default properties.

simeoncran:
This problem has always existed but it didn't matter until we started using subchanel animations for Scale (Scale.X and Scale.Y) in order to support multiple easings.
The issue was that we ask which properties have non-default values by looking at whether the property is set, and whether there is an animation for that property.
But when we find animations for a property we need to ignore the subchannel part of the name (the ".X" part).

### DIFF
--- a/source/UIData/Tools/GraphCompactor.cs
+++ b/source/UIData/Tools/GraphCompactor.cs
@@ -127,7 +127,8 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
             PushContainerShapeTransformsDown(graph, containerShapes);
             CoalesceContainerShapes2(graph, containerShapes);
             PushPropertiesDownToSpriteShape(graph, containerShapes);
-            PushShapeVisbilityDown(graph, containerShapes);
+
+            //PushShapeVisbilityDown(graph, containerShapes);
         }
 
         // Finds sibling shape containers that have the same properties and combines them.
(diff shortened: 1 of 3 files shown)


source/UIData/Tools/GraphCompactor.cs
```diff
@@ -127,7 +127,8 @@ void OptimizeContainerShapes(ObjectGraph<Node> graph)
             PushContainerShapeTransformsDown(graph, containerShapes);
             CoalesceContainerShapes2(graph, containerShapes);
             PushPropertiesDownToSpriteShape(graph, containerShapes);
-            PushShapeVisbilityDown(graph, containerShapes);
+
+            //PushShapeVisbilityDown(graph, containerShapes);
```
eliezerpMS, line 131: >PushShapeVisbilityDown [](start = 14, length = 22)

Is this supposed to be left commented out?

source/UIData/Tools/Properties.cs
```diff
@@ -34,7 +34,16 @@ internal static PropertyId GetNonDefaultCompositionObjectProperties(CompositionO
 
             foreach (var animator in obj.Animators)
             {
-                result |= PropertyIdFromName(animator.AnimatedProperty);
+                var animatedPropertyName = animator.AnimatedProperty;
+
+                // The property name may contain subchannels. Trim those off.
+                var dotIndex = animatedPropertyName.IndexOf('.');
```
eliezerpMS, line 40: >var dotIndex = animatedPropertyName.IndexOf('.'); [](start = 16, length = 49)

do we ever have the case where it can be ".property.subproperty"?

source/UIData/Tools/GraphCompactor.cs
simeoncran: Oooo.. thanks. That's the thing that was triggering the problem so it was commented for my testing. Definitely needs to be uncommented.

---
In reply to: [456061305](https://github.com/windows-toolkit/Lottie-Windows/pull/315#discussion_r456061305) [](ancestors = 456061305)

source/UIData/Tools/Properties.cs
simeoncran: We could, but it should still be right.   "Prop.Foo.Bar"  and "Prop.Foo" should both report that "Prop" has a non-default value. Or am I missing something?

---
In reply to: [456066723](https://github.com/windows-toolkit/Lottie-Windows/pull/315#discussion_r456066723) [](ancestors = 456066723)

PR #316, opened 2020-07-16
Add display of markers to Lottie Viewer.

simeoncran:
And doing a bunch of refactoring to reduce code-behind with the benefit of hindsight.


LottieViewer/LottieVisualDiagnosticsViewModel.cs
```diff
@@ -0,0 +1,134 @@
...
+        internal static (double, double) FloatToRatio(double value)
+        {
+            const int maxRatioProduct = 200;
+            var candidateN = 1.0;
+            var candidateD = Math.Round(1 / value);
+            var error = Math.Abs(value - (candidateN / candidateD));
+
+            for (double n = candidateN, d = candidateD; n * d <= maxRatioProduct && error != 0;)
+            {
+                if (value > n / d)
```
raymond-wh-leung, line 97: >(value > n / d) [](start = 19, length = 15)

Seems that "d" could be zero?

simeoncran: Yep, if value > 2. Not a problem because we're using floating point numbers (doubles).

---
In reply to: [456618039](https://github.com/windows-toolkit/Lottie-Windows/pull/316#discussion_r456618039) [](ancestors = 456618039)

PR #318, opened 2020-07-23
Adding markers to the scrubber.

simeoncran:
This is a rewrite of the Scrubber to support markers. It also is a much nicer implementation of the
scrubber that reuses more of the XAML Slider functionality, and reduces the amount of copied XAML from
the default Slider style (from generic.xaml).

As I was adding this change, it became apparent that the way the XAML slider draws its track would
not work well with markers at 0.0 or 1.0 because the track ends up extending before and after the
0.0 and 1.0 positions which would make the markers look like they were incorrectly placed. So now
we do all of the drawing (track, decrease rectangle, and thumb) using Composition.

We also now mimic the mouse-over behavior of Slider so that the control feels a little more engaging.
This required hooking Composition into the VisualStateManager, which is achieved by replacing the
default VisualStateManager with our own implementation.

### DIFF
--- a/LottieViewer/Scrubber.xaml
+++ b/LottieViewer/Scrubber.xaml
@@ -1,262 +1,222 @@
-﻿<UserControl x:Class="LottieViewer.Scrubber"
-             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-             xmlns:local="using:LottieViewer"
-             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-             d:DesignHeight="300"
-             d:DesignWidth="400"
-             mc:Ignorable="d">
-    <UserControl.Resources>
-        <Style x:Key="ScrubberSliderStyle"
-               TargetType="Slider">
-            <Setter Property="Background" Value="{ThemeResource SliderTrackFill}" />
-            <Setter Property="BorderThickness" Value="{ThemeResource SliderBorderThemeThickness}" />
-            <Setter Property="Foreground" Value="{ThemeResource SliderTrackValueFill}" />
-            <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
-            <Setter Property="FontSize" Value="{ThemeResource ControlContentThemeFontSize}" />
-            <Setter Property="ManipulationMode" Value="None" />
-            <Setter Property="UseSystemFocusVisuals" Value="{StaticResource UseSystemFocusVisuals}" />
-            <Setter Property="FocusVisualMargin" Value="-7,0,-7,0" />
-            <Setter Property="IsFocusEngagementEnabled" Value="True" />
-            <Setter Property="Template">
-                <Setter.Value>
-                    <ControlTemplate TargetType="Slider">
-                        <Grid Margin="{TemplateBinding Padding}">
-                            <Grid.Resources>
-                                <Style x:Key="SliderThumbStyle"
-                                       TargetType="Thumb">
-                                    <Setter Property="BorderThickness" Value="0" />
-                                    <!--<Setter Property="Background" Value="{ThemeResource SliderThumbBackground}"/>-->
-                                    <Setter Property="Background" Value="{StaticResource LottieBasicBrush}" />
-                                    <Setter Property="Template">
-                                        <Setter.Value>
-                                            <ControlTemplate TargetType="Thumb">
-                                                <Ellipse x:Name="ellipse"
-                                                         Fill="{TemplateBinding Foreground}"
-                                                         Stroke="{TemplateBinding Background}"
-                                                         StrokeThickness="2" />
-                                            </ControlTemplate>
-                                        </Setter.Value>
-                                    </Setter>
-                                </Style>
-                            </Grid.Resources>
-                            <Grid.RowDefinitions>
-                                <RowDefinition Height="Auto" />
-                                <RowDefinition Height="*" />
-                            </Grid.RowDefinitions>
-                            <ContentPresenter x:Name="HeaderContentPresenter"
-                                              Margin="{ThemeResource SliderHeaderThemeMargin}"
-                                              x:DeferLoadStrategy="Lazy"
-                                              Content="{TemplateBinding Header}"
-                                              ContentTemplate="{TemplateBinding HeaderTemplate}"
-                                              FontWeight="{ThemeResource SliderHeaderThemeFontWeight}"
-                                              Foreground="{ThemeResource SliderHeaderForeground}"
-                                              TextWrapping="Wrap"
-                                              Visibility="Collapsed" />
-                            <Grid x:Name="SliderContainer"
-                                  Grid.Row="1"
-                                  Background="{ThemeResource SliderContainerBackground}"
-                                  Control.IsTemplateFocusTarget="True">
-                                <Grid x:Name="HorizontalTemplate"
-                                      MinHeight="44">
-                                    <Grid.ColumnDefinitions>
-                                        <ColumnDefinition Width="Auto" />
-                                        <ColumnDefinition Width="Auto" />
-                                        <ColumnDefinition Width="*" />
-                                    </Grid.ColumnDefinitions>
-                                    <Grid.RowDefinitions>
-                                        <RowDefinition Height="18" />
-                                        <RowDefinition Height="Auto" />
-                                        <RowDefinition Height="18" />
-                                    </Grid.RowDefinitions>
-                                    <Rectangle x:Name="HorizontalTrackRect"
-                                               Grid.Row="1"
-                                               Grid.ColumnSpan="3"
-                                               Height="{ThemeResource SliderTrackThemeHeight}"
-                                               Fill="{TemplateBinding Background}" />
-                                    <Rectangle x:Name="HorizontalDecreaseRect"
-                                               Grid.Row="1"
-                                               Fill="{StaticResource LottieBasicBrush}" />
-                                    <TickBar x:Name="TopTickBar"
-                                             Grid.ColumnSpan="3"
-                                             Height="{ThemeResource SliderOutsideTickBarThemeHeight}"
-                                             Margin="0,0,0,4"
-                                             VerticalAlignment="Bottom"
-                                             Fill="{ThemeResource SliderTickBarFill}"
-                                             Visibility="Collapsed" />
-                                    <TickBar x:Name="HorizontalInlineTickBar"
-                                             Grid.Row="1"
-                                             Grid.ColumnSpan="3"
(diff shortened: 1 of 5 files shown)


LottieViewer/Scrubber.xaml
```diff
@@ -1,262 +1,222 @@
...
+                        </Grid.RowDefinitions>
+                        <ContentPresenter x:Name="HeaderContentPresenter"
+                                Grid.Row="0"
+                                Content="{TemplateBinding Header}"
+                                ContentTemplate="{TemplateBinding HeaderTemplate}"
+                                FontWeight="{ThemeResource SliderHeaderThemeFontWeight}"
+                                Foreground="{ThemeResource SliderHeaderForeground}"
+                                Margin="{ThemeResource SliderTopHeaderMargin}"
+                                TextWrapping="Wrap"
+                                Visibility="Collapsed"
```
raymond-wh-leung, line 138: >  Visibility="Collapsed" [](start = 30, length = 24)

For the elements that are not visible (or with opacity 0), they need to remain in the template because the slider control will reference them?

LottieViewer/Scrubber.xaml
```diff
@@ -1,262 +1,222 @@
...
+                                <TickBar x:Name="RightTickBar"
+                                        Visibility="Collapsed"
+                                        Fill="{ThemeResource SliderTickBarFill}"
+                                        Height="{ThemeResource SliderOutsideTickBarThemeHeight}"
+                                        VerticalAlignment="Top"
+                                        Margin="0,4,0,0"
+                                        Grid.Row="2"
+                                        Grid.ColumnSpan="3" />
+
+                                <!-- The Width of the Thumb has been modified to make it round. -->
```
raymond-wh-leung, line 203: >modified to make it round. [](start = 69, length = 26)

Is it needed coz looks like the thumb is transparent?

LottieViewer/Scrubber.xaml.cs
```diff
@@ -134,33 +144,253 @@ public double Value
...
+            // slider for its size.
+            var result = base.ArrangeOverride(finalSize);
 
-            return base.ArrangeOverride(finalSize);
+            var sliderWidth = _slider.ActualWidth;
+
+            // Update the size of the progress rectangle and the position of the thumb.
+            _properties.InsertScalar("Width", (float)sliderWidth);
+
+            // Update the positio of the markers.
```
raymond-wh-leung, line 192: >positio [](start = 26, length = 7)

position

LottieViewer/Scrubber.xaml
simeoncran: The thumb is receiving the input. It just isn't visible. So yes, it's needed.
Also, some components in the template need to be there even if they're not visible because the Slider will try to reference them.

---
In reply to: [463150427](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r463150427) [](ancestors = 463150427)

LottieViewer/Scrubber.xaml
simeoncran: Yes. I've removed some that I was confident about. I removed some others and got bitten by crashes when they were referenced by the Slider. There are probably more that can be removed but I don't feel confident about them. We're already cheaper than a normal slider as a result of removing some of the things from the template.

---
In reply to: [463148127](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r463148127) [](ancestors = 463148127)

LottieViewer/Scrubber.xaml
```diff
@@ -1,262 +1,222 @@
...
+                                <Thumb x:Name="HorizontalThumb"
+                                        Opacity="0"
+                                        Style="{StaticResource SliderThumbStyle}"
+                                        DataContext="{TemplateBinding Value}"
+                                        Height="20"
```
derobinsMSFT, line 208: >Height="20" [](start = 40, length = 11)

Is there a default `ThemeResource` height you can use for this? Or, is this a modified height to keep the thumb round (along with the width)?

LottieViewer/Scrubber.xaml.cs
```diff
@@ -134,33 +144,253 @@ public double Value
...
+                        track: GetResourceBrushColor("SliderTrackFillDisabled"),
+                        decreaseRectangle: GetResourceBrushColor("SliderTrackValueFillDisabled"));
+                    break;
+                case "Pressed":
+                case "PointerOver":
+                    SetColors(
+                        markers: Colors.White,
+                        thumbFill: GetResourceBrushColor("LottieBasicBrush"),
+                        thumbStroke: GetResourceBrushColor("LottieBasicBrush"),
+                        track: GetResourceBrushColor("SliderTrackFillPointerOver"),
```
derobinsMSFT, line 255: >"SliderTrackFillPointerOver" [](start = 53, length = 28)

"SliderTrackFillPressed" for case "Pressed"?

LottieViewer/Scrubber.xaml.cs
```diff
@@ -134,33 +144,253 @@ public double Value
...
+
+        // Returns a single-pixel-wide Rectangle for displaying a marker above or below the track.
+        Rectangle CreateMarkerRectangle() => new Rectangle() { Fill = _markerBrush, Width = 1, HorizontalAlignment = HorizontalAlignment.Left };
+
+        void Markers_CollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            switch (e.Action)
+            {
+                case NotifyCollectionChangedAction.Add:
+                    // Add rectangles to display each marker. There are 2 rectangle - one
```
derobinsMSFT, line 287: >rectangle [](start = 74, length = 9)

"rectangles"

LottieViewer/Scrubber.xaml.cs
simeoncran: I don't know how to press the slider, and I'm fine with it not changing colors. This makes it simpler than trying to handle the pressed state explicitly.

---
In reply to: [463305999](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r463305999) [](ancestors = 463305999)

LottieViewer/Scrubber.xaml
simeoncran: It is supposed to be the same shape as the slider we draw, which is not theme aware.

---
In reply to: [463298421](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r463298421) [](ancestors = 463298421)

LottieViewer/Scrubber.xaml
```diff
@@ -1,262 +1,222 @@
...
-                    </ControlTemplate>
-                </Setter.Value>
-            </Setter>
-        </Style>
-    </UserControl.Resources>
+        <Slider x:Name="_slider" 
+                Maximum="1" 
+                StepFrequency="0.0001" 
+                SmallChange="0.001" 
+                AutomationProperties.Name="Progress slider">
```
karenbtlai, line 27: nit: rename to `ProgressSlider`

LottieViewer/Scrubber.xaml.cs
```diff
@@ -1,128 +1,138 @@
...
+
+            // Create the brushes and set up expression animations so their colors can
+            // be changed by writing to the property set.
+            var thumbFillBrush = CreateBoundColorBrush(_properties, "ThumbFillColor");
+            var thumbStrokeBrush = CreateBoundColorBrush(_properties, "ThumbStrokeColor");
+            _trackRectangleBrush = CreateBoundColorBrush(_properties, "TrackColor");
+            _decreaseRectangleBrush = CreateBoundColorBrush(_properties, "DecreaseRectangleColor");
+
+            // Create the track rectangle. This is the track that the thumb moves along.
+            _trackRectangle = c.CreateSpriteVisual();
```
karenbtlai, line 96: I'm curious why the `trackRectangle` and `decreaseRectangle` have to be initialized in code behind when it's already set in the template?

LottieViewer/Scrubber.xaml
simeoncran: This is a copy of the automation name that was given by another contributor. I don't know the rules for these names - is it supposed to be a single word?

---
In reply to: [465417002](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r465417002) [](ancestors = 465417002)

LottieViewer/Scrubber.xaml.cs
```diff
@@ -134,33 +144,253 @@ public double Value
...
+                        decreaseRectangle: GetResourceBrushColor("LottieBasicBrush"));
+                    break;
+                default:
+                    throw new InvalidOperationException();
+            }
+        }
+
+        void SetColors(Color markers, Color thumbFill, Color thumbStroke, Color track, Color decreaseRectangle)
+        {
+            _markerBrush.Color = markers;
```
karenbtlai, line 265: I'm curious as to why colours are set in code behind? The brushes are already set in the VisualStateManager in xaml, thus you can update colours through updating the VisualState.

LottieViewer/Scrubber.xaml.cs
simeoncran: They're not the same object. The template contains the XAML Rectangles, but with opacity=0 so they don't show up. We use Composition to draw the UI, but XAML to receive the input. Composition SpriteVisuals can be animated via expressions that respond to changes on other Composition object. The problem this solves is making the slider move while the Lottie is playing. When the Lottie is playing, there's no code running in the LottieViewer process, so we can't animate the XAML Rectangles.



---
In reply to: [465905081](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r465905081) [](ancestors = 465905081)

LottieViewer/Scrubber.xaml.cs
simeoncran: Different brush - this is updating the Composition objects, not the XAML objects.

---
In reply to: [465914466](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r465914466) [](ancestors = 465914466)

LottieViewer/Scrubber.xaml
simeoncran: Discussed privately - WinUI team makes these a single Pascal-cased word, so we'll do the same.s

---
In reply to: [465910831](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r465910831) [](ancestors = 465910831,465417002)

simeoncran: But that would go against MSDN official advice: [https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.automationproperties.nameproperty?view=winrt-19041](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.automationproperties.nameproperty?view=winrt-19041)

"The Name property should be the same as the label text on screen."

We wouldn't smoosh and Pascal-case label text.

We don't have a label for this slider, but the name should correspond with the label that we would use if there were a label.

---
In reply to: [465975160](https://github.com/windows-toolkit/Lottie-Windows/pull/318#discussion_r465975160) [](ancestors = 465975160,465910831,465417002)

PR #326, opened 2020-08-04
Fix translation of rectangles with RoundCorners.

simeoncran:
We were not handling the distinction between RoundCorners.Radius and Rectangle.Roundness. This change does a much better job.
Note that it gets the Trimming wrong for RoundCorners - that will come in a later fix. I'm getting this in now because it has so much refactoring that it will be hard to focus on the trimming algorithm alongside this.

These are the After Effects rules for how RoundCorners.Radius and Rectangle.Roundness interact:
RoundCorners.Radius is ignored if Rectangle.Roundness != 0.
RoundCorners.Radius is able to produce an ellipse if the radius is large enough.
Rectangle.Roundness creates round corners with equal X and Y radii, so it can never create an ellipse.

### DIFF
--- a/source/LottieToWinComp/CompositeOpacity.cs
+++ b/source/LottieToWinComp/CompositeOpacity.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         /// this <see cref="CompositeOpacity"/>.</returns>
         internal CompositeOpacity ComposedWith(in TrimmedAnimatable<Opacity> opacity)
         {
-            if (opacity.AlwaysEquals(Opacity.Opaque))
+            if (opacity.Always(Opacity.Opaque))
             {
                 // Nothing to do.
                 return this;
--- a/source/LottieToWinComp/ExpressionFactory.cs
+++ b/source/LottieToWinComp/ExpressionFactory.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         internal static readonly Vector2 MyPosition = MyVector2("Position");
         internal static readonly Vector2 MySize = MyVector2("Size");
         internal static readonly Matrix3x2 MyTransformMatrix = MyMatrix3x2("TransformMatrix");
+        static readonly Scalar MyRadius = MyScalar("Radius");
         static readonly Scalar MyRoundness = MyScalar("Roundness");
         static readonly Scalar MyTStart = MyScalar("TStart");
         static readonly Scalar MyTEnd = MyScalar("TEnd");
@@ -68,13 +69,34 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
         internal static Scalar RootScalar(string propertyName) => Scalar(RootProperty(propertyName));
 
-        internal static Vector2 ConstrainedCornerRadiusScalar(double roundness)
+        // Converts a RoundCorners.Radius value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // RoundCorners corner radii are constrained to half of the coresponding side.
+        internal static Vector2 RadiusToCornerRadius(double radius)
+            => Vector2(Min(radius, MySize.X / 2), Min(radius, MySize.Y / 2));
+
+        // Converts a RoundCorners.Radius value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // RoundCorners corner radii are constrained to half of the coresponding side.
+        internal static Vector2 RadiusToCornerRadius()
+            => Vector2(Min(MyRadius, MySize.X / 2), Min(MyRadius, MySize.Y / 2));
+
+        // Converts a RoundCorners.Radius value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // RoundCorners corner radii are constrained to half of the coresponding side.
+        internal static Vector2 RadiusToCornerRadius(Sn.Vector2 size)
+            => Vector2(Min(MyRadius, size.X / 2), Min(MyRadius, size.Y / 2));
+
+        // Converts a Rectangle.Roundness value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // Rectangle.Roundness corner radius is constrained to half of the smaller side.
+        internal static Vector2 RoundnessToCornerRadius(double roundness)
             => Vector2(Min(roundness, Min(MySize.X, MySize.Y) / 2), Min(roundness, Min(MySize.X, MySize.Y) / 2));
 
-        internal static Vector2 ConstrainedCornerRadiusScalar()
+        // Converts a Rectangle.Roundness value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // Rectangle.Roundness corner radius is constrained to half of the smaller side.
+        internal static Vector2 RoundessToCornerRadius()
             => Vector2(Min(MyRoundness, Min(MySize.X, MySize.Y) / 2), Min(MyRoundness, Min(MySize.X, MySize.Y) / 2));
 
-        internal static Vector2 ConstrainedCornerRadiusScalar(Sn.Vector2 size)
+        // Converts a Rectangle.Roundness value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // Rectangle.Roundness corner radius is constrained to half of the smaller side.
+        internal static Vector2 RoundnessToCornerRadius(Sn.Vector2 size)
             => Vector2(Min(MyRoundness, Math.Min(size.X, size.Y) / 2), Min(MyRoundness, Math.Min(size.X, size.Y) / 2));
 
         // The value of a Color property stored as a Vector4 on the theming property set.
--- a/source/LottieToWinComp/LottieToWinCompTranslator.Rectangles.cs
+++ b/source/LottieToWinComp/LottieToWinCompTranslator.Rectangles.cs
@@ -1,0 +1,497 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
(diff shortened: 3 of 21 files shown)


source/LottieToWinComp/ExpressionFactory.cs
```diff
@@ -68,13 +69,34 @@ internal static Color ColorMultipliedByPreMultipliedOpacities(Wui.Color color, S
...
+
+        // Converts a RoundCorners.Radius value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // RoundCorners corner radii are constrained to half of the coresponding side.
+        internal static Vector2 RadiusToCornerRadius(Sn.Vector2 size)
+            => Vector2(Min(MyRadius, size.X / 2), Min(MyRadius, size.Y / 2));
+
+        // Converts a Rectangle.Roundness value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // Rectangle.Roundness corner radius is constrained to half of the smaller side.
+        internal static Vector2 RoundnessToCornerRadius(double roundness)
             => Vector2(Min(roundness, Min(MySize.X, MySize.Y) / 2), Min(roundness, Min(MySize.X, MySize.Y) / 2));
```
raymond-wh-leung, line 90: > Min(MySize.X, MySize.Y) / 2 [](start = 37, length = 28)

Is that because Rectangle.Roundness won't have different X/Y radius?

source/LottieToWinComp/ExpressionFactory.cs
```diff
@@ -68,13 +69,34 @@ internal static Color ColorMultipliedByPreMultipliedOpacities(Wui.Color color, S
...
+        // Converts a Rectangle.Roundness value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // Rectangle.Roundness corner radius is constrained to half of the smaller side.
+        internal static Vector2 RoundessToCornerRadius()
             => Vector2(Min(MyRoundness, Min(MySize.X, MySize.Y) / 2), Min(MyRoundness, Min(MySize.X, MySize.Y) / 2));
 
-        internal static Vector2 ConstrainedCornerRadiusScalar(Sn.Vector2 size)
+        // Converts a Rectangle.Roundness value to CompositionRoundedRectangleGeometry.CornerRadius.
+        // Rectangle.Roundness corner radius is constrained to half of the smaller side.
+        internal static Vector2 RoundnessToCornerRadius(Sn.Vector2 size)
             => Vector2(Min(MyRoundness, Math.Min(size.X, size.Y) / 2), Min(MyRoundness, Math.Min(size.X, size.Y) / 2));
```
raymond-wh-leung, line 100: >Math. [](start = 40, length = 5)

nit: can this just bin Min() like the rest of the calls here?

source/LottieToWinComp/CompositeOpacity.cs
```diff
@@ -38,7 +38,7 @@ sealed class CompositeOpacity
         /// this <see cref="CompositeOpacity"/>.</returns>
         internal CompositeOpacity ComposedWith(in TrimmedAnimatable<Opacity> opacity)
         {
-            if (opacity.AlwaysEquals(Opacity.Opaque))
```
raymond-wh-leung, line 41: >AlwaysEquals [](start = 24, length = 12)

nit: AlwaysEquals() seems clearer about what it does.

source/LottieToWinComp/LottieToWinCompTranslator.Rectangles.cs
```diff
@@ -0,0 +1,497 @@
...
+                        cornerRadius,
+                        geometry,
+                        cornerRadiusIsRectangleRoundness ? "Roundness" : "Radius");
+
+                    if (isSizeAnimated)
+                    {
+                        // Both size and cornerRadius are animated.
+                        cornerRadiusExpression = cornerRadiusIsRectangleRoundness
+                                                        ? RoundessToCornerRadius()
+                                                        : RadiusToCornerRadius();
```
raymond-wh-leung, line 180: Would it work to use this expression for all the other cases? (i.e. would the expression read the InitialRadius / InitialiSize for the non-animating case?)

source/LottieToWinComp/ExpressionFactory.cs
simeoncran: Correct - roundness describes the radius of a circle, not an ellipse. Once the smaller side of the rectangle is a circle, the radius is not allowed to grow any bigger.
Let me know how I can make that clearer in the comments.

---
In reply to: [465894379](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r465894379) [](ancestors = 465894379)

source/LottieToWinComp/ExpressionFactory.cs
simeoncran: I have mixed feelings about it - it will create a bunch of Expression objects to represent Math.Min(size.X, size.Y) / 2. Those objects will get simplified down to the literal value of Math.Min(size.X, size.Y) / 2 by the time we actually create the expression string, so it will work just fine. But is it better to do that rather than making it explicit here that the value is a literal?

Thoughts?


---
In reply to: [465895461](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r465895461) [](ancestors = 465895461)

source/LottieToWinComp/CompositeOpacity.cs
simeoncran: How strongly do you feel about it? I shortened it because I felt that the Equals was redundant, and opacity.Always(Opaque) was just as descriptive.

---
In reply to: [465897126](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r465897126) [](ancestors = 465897126)

source/LottieToWinComp/LottieToWinCompTranslator.Rectangles.cs
simeoncran: Yes, I think that would be fine. It would add another dependency to the expressions in those other cases. For the other cases the expression uses a literal for the non-animated value (either radius or size) so DWM doesn't have to listen for changes to those non-animated values.
Are you asking because it seems like it would be better that way?

---
In reply to: [465904859](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r465904859) [](ancestors = 465904859)

source/LottieToWinComp/ExpressionFactory.cs
raymond-wh-leung: Makes sense, it is clearer after reading the PR description about the Rectangle.Roundness behavior.

---
In reply to: [466582891](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466582891) [](ancestors = 466582891,465894379)

source/LottieToWinComp/ExpressionFactory.cs
raymond-wh-leung: Is Math.Min() used here for constant values and Min() is used at the other places for reference parameters?

---
In reply to: [466587733](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466587733) [](ancestors = 466587733,465895461)

source/LottieToWinComp/CompositeOpacity.cs
raymond-wh-leung: Seems fine to me too. (Always() seems match closer to the other methods, like Ever() and EverNot() in Animatable). Had to take a peek at the function comments about what it does the first time I saw this.

---
In reply to: [466589696](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466589696) [](ancestors = 466589696,465897126)

source/LottieToWinComp/LottieToWinCompTranslator.Rectangles.cs
raymond-wh-leung: I saw special handlings for the different cases in various places so was wondering about the reason behind that. Providing the values upfront (if the value is constant) would indeed save some operations in dwm (not sure if the perf delta is anything significant though).

---
In reply to: [466594134](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466594134) [](ancestors = 466594134,465904859)

source/LottieToWinComp/ExpressionFactory.cs
simeoncran: Math.Min() is used for values known at compile time. Min(...) is used for expressions that will be evaluated at runtime by DWM. However we later simplify Min(...) to a compile time value if possible.

More context:
Min(...) is a method that returns an Expression.Scalar object. The Expression class is part of Lottie-Windows - it allows us to build Composition animation expression strings in a strongly typed way and to control policy in one place for how expressions are simplified and converted to strings. Math.Min is the .NET Min method that returns a double. So Min(MyRoundness, Math.Min(size.X, size.Y) /2) is a call to:   Expression.Scalar Min(Expression.Scalar, double).
If we used Min instead of Math.Min, it would be a call to: Expression.Scalar Min(Expression.Scalar, Expression.Scalar)



---
In reply to: [466606405](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466606405) [](ancestors = 466606405,466587733,465895461)

source/LottieToWinComp/CompositeOpacity.cs
simeoncran: I think now that this should be IsAlways(), IsEver(), IsEverNot(), etc. I'll do that on a later change though.


---
In reply to: [466611227](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466611227) [](ancestors = 466611227,466589696,465897126)

source/LottieToWinComp/ExpressionFactory.cs
raymond-wh-leung: Gotcha, thanks for clarifying. 

---
In reply to: [466658397](https://github.com/windows-toolkit/Lottie-Windows/pull/326#discussion_r466658397) [](ancestors = 466658397,466606405,466587733,465895461)

PR #328, opened 2020-08-05
CX classes stopped compiling because it wasn't public. Make it public.

simeoncran:
Apparently something changed in the way XAML gets info about CX classes - it used to compile but now it won't compile unless the classes are public. Public classes publish their metadata, which is presumably how XAML is now getting info about the classes. This change makes the CX classes always public.

### DIFF
--- a/source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
+++ b/source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
@@ -50,6 +50,14 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen
             return (cppText, hText, assetList);
         }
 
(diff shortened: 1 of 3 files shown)


source/UIDataCodeGen/CodeGen/CppwinrtInstantiatorGenerator.cs
```diff
@@ -50,6 +50,14 @@ sealed class CppwinrtInstantiatorGenerator : CppInstantiatorGeneratorBase
             return (cppText, hText, assetList);
         }
 
+        private protected override void WriteHeaderClassStart(HeaderBuilder builder, IAnimatedVisualSourceInfo info, string inherits)
+        {
+            builder.Preamble.WriteLine($"{(info.Public ? "public" : string.Empty)}class {SourceClassName}");
```
derobinsMSFT, line 55: >" [](start = 64, length = 1)

Add a space.

PR #338, opened 2020-08-24
Fix the namespace for IGeometrySource2D.

simeoncran:
This got broken when we added the WinUI3 switch to LottieGen.


source/UIDataCodeGen/CodeGen/CSharpInstantiatorGenerator.cs
```diff
@@ -85,8 +85,8 @@ protected override void WriteFileStart(CodeBuilder builder)
 
             if (Info.UsesCanvasGeometry)
             {
-                // Microsoft.Graphics is needed for IGeometrySource2D.
-                namespaces.Add("Microsoft.Graphics");
+                // Windows.Graphics is needed for IGeometrySource2D.
+                namespaces.Add("Windows.Graphics");
                 namespaces.Add("Microsoft.Graphics.Canvas.Geometry");
```
derobinsMSFT, line 90: >Microsoft [](start = 32, length = 9)

Does this also need to be "Windows"?

simeoncran: No, this is part of Win2D, which was never in the Windows namespace.

---
In reply to: [476195993](https://github.com/windows-toolkit/Lottie-Windows/pull/338#discussion_r476195993) [](ancestors = 476195993)

Issue #333, opened 2020-08-15
Load animation from memory


under3415:
Instead of loading animation from a storage file or a URI, is it possible to load it from memory? Either a string or a memory buffer of some sort?

Some scenarios where this would be useful:
- JSON stored in a database
- JSON files encrypted to protect IP
- JSON files compressed to save on disk space or reduce number of files on disk for deployment

nishitha-burman:
P2
There is already a PR for this
Dev Cost: Half a day

simeoncran:
Fixed with #341

PR #341, opened 2020-08-25
Allow JSON to be loaded from any stream, including memory

simeoncran:
There was a request (#333) to be able to load Lottie files from memory. This change allows that by enabling any IInputStream to be used.

As part of this change, I found some cruft that I could clean up, and I pulled out the classes for the Loader and ContentFactory and did some other straightforward refactoring.

### DIFF
--- a/source/Lottie/ContentFactory.cs
+++ b/source/Lottie/ContentFactory.cs
@@ -1,0 +1,73 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Diagnostics;
+using Microsoft.UI.Xaml.Controls;
+using Windows.UI.Composition;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie
+{
+    /// <summary>
+    /// Information from which a composition's content can be instantiated. Contains the WinCompData
+    /// translation of a composition and some metadata. This allows multiple instances of the translation
+    /// to be instantiated without requiring repeated translations.
+    /// </summary>
+    sealed class ContentFactory : IAnimatedVisualSource
+    {
+        internal static readonly ContentFactory FailedContent = new ContentFactory(null);
+        readonly LottieVisualDiagnostics _diagnostics;
+        WinCompData.Visual _wincompDataRootVisual;
+        double _width;
+        double _height;
+        TimeSpan _duration;
+
+        internal ContentFactory(LottieVisualDiagnostics diagnostics)
+        {
+            _diagnostics = diagnostics;
+        }
+
+        internal void SetDimensions(double width, double height, TimeSpan duration)
+        {
+            _width = width;
+            _height = height;
+            _duration = duration;
+        }
+
+        internal void SetRootVisual(WinCompData.Visual rootVisual)
+        {
+            _wincompDataRootVisual = rootVisual;
+        }
+
+        internal bool CanInstantiate => _wincompDataRootVisual != null;
+
+        public IAnimatedVisual TryCreateAnimatedVisual(Compositor compositor, out object diagnostics)
+        {
+            var diags = _diagnostics?.Clone();
+            diagnostics = diags;
+
+            if (!CanInstantiate)
+            {
+                return null;
+            }
+            else
+            {
+                var sw = Stopwatch.StartNew();
+                var result = new DisposableAnimatedVisual()
+                {
+                    RootVisual = Instantiator.CreateVisual(compositor, _wincompDataRootVisual),
+                    Size = new System.Numerics.Vector2((float)_width, (float)_height),
+                    Duration = _duration,
+                };
+
+                if (diags != null)
+                {
+                    diags.InstantiationTime = sw.Elapsed;
+                }
+
+                return result;
+            }
+        }
+    }
+}
--- a/source/Lottie/Loader.cs
+++ b/source/Lottie/Loader.cs
@@ -1,0 +1,315 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if DEBUG
+// Uncomment this to slow down async awaits for testing.
+//#define SlowAwaits
+#endif
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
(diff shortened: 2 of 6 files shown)


source/Lottie/ContentFactory.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
```
eliezerpMS, line 5: Random style question: What is the reasoning behind putting the System namespace before the Microsoft one?

source/Lottie/Loader.cs
```diff
@@ -0,0 +1,315 @@
...
+namespace Microsoft.Toolkit.Uwp.UI.Lottie
+{
+    /// <summary>
+    /// Handles loading a composition from a Lottie file. The result of the load
+    /// is a <see cref="ContentFactory"/> that can be used to instantiate a
+    /// Composition tree that will render the Lottie.
+    /// </summary>
+    abstract class Loader
+    {
+        Loader()
```
eliezerpMS, line 32: >Loader() [](start = 8, length = 8)

does it need a default constructor?

source/Lottie/Loader.cs
```diff
@@ -0,0 +1,315 @@
...
+
+                if (diagnostics != null)
+                {
+                    diagnostics.JsonParsingIssues = ToIssues(readerIssues);
+                }
+            }));
+
+            if (diagnostics != null)
+            {
+                diagnostics.ParseTime = sw.Elapsed;
```
eliezerpMS, line 86: >sw.Elapsed [](start = 40, length = 10)

Just curious, do you always want to do "sw.Restart()" after getting sw.Elapsed? If so then what about having 1 call that does both so that future users never accidently forget to call Restart()?

source/Lottie/Loader.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            // Translating large Lotties can take significant time. Do it on another thread.
+            WinCompData.Visual wincompDataRootVisual = null;
+            uint requiredUapVersion = 0;
+            var optimizationEnabled = options.HasFlag(LottieVisualOptions.Optimize);
+
+            TranslationResult translationResult;
+            await CheckedAwaitAsync(Task.Run(() =>
+            {
+                // translatePropertyBindings is turned on so that it can report
```
eliezerpMS, line 128: >translatePropertyBindings [](start = 19, length = 25)

nit: Capitalize?

source/Lottie/Loader.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+            readonly IInputStream _inputStream;
+
+            internal FromInputStream(IInputStream inputStream)
+            {
+                AssertNotNull(inputStream);
+                _inputStream = inputStream;
+            }
+
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
```
eliezerpMS, line 232: >// Async method lacks 'await' operators and will run synchronously [](start = 31, length = 66)

Maybe a comment here as to why you need to disable the warning? (something about how it is needed so it can be used with another method that is async I assume?)

source/Lottie/Loader.cs
```diff
@@ -0,0 +1,315 @@
...
+                    }
+                }
+
+                return (null, null);
+            }
+        }
+
+        // ----
+        // BEGIN: DEBUGGING HELPERS
+        // ----
```
eliezerpMS, line 291: Maybe use #pragma section?

source/Lottie/ContentFactory.cs
simeoncran: [https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#organize-using-directives](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#organize-using-directives)
Arbitrary as to whether System goes first "dotnet_sort_system_directives_first = true"
We inherited this style from the Community Toolkit.

---
In reply to: [477472562](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477472562) [](ancestors = 477472562)

source/Lottie/Loader.cs
simeoncran: No! Good catch.

---
In reply to: [477473230](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477473230) [](ancestors = 477473230)

source/Lottie/Loader.cs
simeoncran: Not sure what else I can put there to make it clearer. It has to be an async so that returning a value will actually return a Task. And it has to return a Task because the method it's overriding is a Task. Will think about what to write here...

---
In reply to: [477478355](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477478355) [](ancestors = 477478355)

source/Lottie/Loader.cs
simeoncran: I think you mean #region, but alas, StyleCop doesn't allow #regions - they're considered very 2010 apparently.

---
In reply to: [477479109](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477479109) [](ancestors = 477479109)

source/Lottie/Loader.cs
simeoncran: And also I think this should be tied to a  LottieVisualOptions option.

---
In reply to: [477476455](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477476455) [](ancestors = 477476455)

source/Lottie/Loader.cs
simeoncran: Not sure if I've made it better... see what you think.

---
In reply to: [477475929](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477475929) [](ancestors = 477475929)

eliezerpMS: Yep I think this is good

---
In reply to: [477562981](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477562981) [](ancestors = 477562981,477475929)

source/Lottie/Loader.cs
eliezerpMS: hmm strange that StyleCop cares about that...

---
In reply to: [477529929](https://github.com/windows-toolkit/Lottie-Windows/pull/341#discussion_r477529929) [](ancestors = 477529929,477479109)

PR #352, opened 2020-09-18
Start on cppwinrt codegen for IDynamicAnimatedVisualSource.

simeoncran:
This is not complete (IDynamicAnimatedVisualSource code will not compile). However it fixes a number of IDynamicAnimatedVisualSource issues.
Also gives unique names to the AnimatedVisual classes. This should make callstacks clearer in C++.


derobinsMSFT (approved):
:shipit:

PR #354, opened 2020-09-22
Get image layer working with cppwinrt.

simeoncran:
Has some issues if the images are not found, but it works otherwise.
Includes some untested cleaning up of the CX code.

### DIFF
--- a/source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
+++ b/source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
@@ -694,7 +694,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen.Cppwinrt
             builder.WriteLine($"compositeEffect.SetMode({_s.CanvasCompositeMode(compositeEffect.Mode)});");
             foreach (var source in compositeEffect.Sources)
             {
-                builder.WriteLine($"compositeEffect.AddSource(CompositionEffectSourceParameter(L\"{source.Name}));");
+                builder.WriteLine($"compositeEffect.AddSource(CompositionEffectSourceParameter(L\"{source.Name}\"));");
             }
 
             return "compositeEffect";
--- a/source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
+++ b/source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
@@ -175,7 +175,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.CodeGen.Cppwinrt
                 throw new NotImplementedException();
             }
 
-            builder.Class.Preamble.WriteLine($"class {_sourceClassName} final");
+            builder.Class.Preamble.WriteLine($"class {_sourceClassName}");
             builder.Class.Preamble.Indent();
             builder.Class.Preamble.WriteLine($": public {_sourceClassName}T<{_sourceClassName}>");
             builder.Class.Preamble.UnIndent();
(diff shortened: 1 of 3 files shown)


scott-moore-ms:
Would it be useful/helpful to have a checked-in reference file?  Then you can look at the actual code being generated and have context to reason about what the codegen tool is doing

source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
```diff
@@ -362,69 +390,57 @@ void WriteThemePropertyImpls(CodeBuilder builder)
...
+            var pub = builder.Class.Public;
+            var priv = builder.Class.Private;
+
+            priv.WriteLine($"const int c_loadedImageSurfaceCount = {SourceInfo.LoadedImageSurfaces.Distinct().Count()};");
+            priv.WriteLine("double m_imageSuccessfulLoadingProgress{};");
+            priv.WriteLine("int m_loadCompleteEventCount{};");
+            priv.WriteLine("bool m_isAnimatedVisualSourceDynamic{};");
+            priv.WriteLine("bool m_isImageLoadingCompleted{};");
+            priv.WriteLine("bool m_isTryCreateAnimatedVisualCalled{};");
+            priv.WriteLine("bool m_isImageLoadingStarted{};");
```
scott-moore-ms, line 402: There seems to be an amount of redundant state stored on this object which is simply derived from the completion count relative to the total count.  Was it an explicit choice to have progress/started/completed be state that needs to be updated and kept in sync, vs simply computed when requested?

simeoncran: Do you mean the isImageLoadingCompleted being the same as loadCompleteEventCount == number_of_images? Or are there others?
I don't remember all the decisions that went into the design - it was the result of people asking for features for handling various loading states, but I think they changed their minds and never used them. And I think we changed the policy a bit at one point and the code probably could have been simplified then.

---
In reply to: [493105017](https://github.com/windows-toolkit/Lottie-Windows/pull/354#discussion_r493105017) [](ancestors = 493105017)

simeoncran:
Yes, but I don't have a file that I can share publicly. I've sent you an example output in mail.

---
In reply to: [697048111](https://github.com/windows-toolkit/Lottie-Windows/pull/354#issuecomment-697048111) [](ancestors = 697048111)

source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
scott-moore-ms: Well, I was originally talking also about m_imageSuccessfulLoadingProgress, but looking at the generated code for that, it actually appears on the surface to be buggy?

        m_loadCompleteEventCount++;
        if (e.Status() == LoadedImageSourceLoadStatus::Success)
        {
            m_imageSuccessfulLoadingProgress = (double)m_loadCompleteEventCount / c_loadedImageSurfaceCount;
            m_PropertyChanged(*this, Windows::UI::Xaml::Data::PropertyChangedEventArgs(L"ImageSuccessfulLoadingProgress"));
        }

If I have 5 images, and 4 of them fail to load, but the 5th one loads successfully, that would result in a progress of 1.0 rather than the (I think?) intended 0.2?


---
In reply to: [493109988](https://github.com/windows-toolkit/Lottie-Windows/pull/354#discussion_r493109988) [](ancestors = 493109988,493105017)

scott-moore-ms (approved):
:shipit:

source/UIDataCodeGen/CodeGen/Cppwinrt/CppwinrtInstantiatorGenerator.cs
simeoncran: Yep, that looks wrong to me. I'll remove the image loading progress. [https://github.com/windows-toolkit/Lottie-Windows/issues/355](https://github.com/windows-toolkit/Lottie-Windows/issues/355)
Not doing it with this change because it affects all of the code generators rather than just C++.

---
In reply to: [493804067](https://github.com/windows-toolkit/Lottie-Windows/pull/354#discussion_r493804067) [](ancestors = 493804067,493109988,493105017)

PR #356, opened 2020-09-24
Fix 2 bugs - path translation optimizer, cppwinrt codegen.

simeoncran:
The path translation optimizer didn't handle the case of paths that consist of single beziers. As a result, some animations were being incorrectly optimized into offset animations.
CppWinrt codegen was instantiating the implementation of CompositeEffect on the stack when it should have been on the heap.

### DIFF
--- a/source/LottieToWinComp/Paths.cs
+++ b/source/LottieToWinComp/Paths.cs
@@ -344,19 +344,32 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 return false;
             }
 
+            // Don't compare the values directly - there could be some rounding errors that
+            // are acceptable. This value is just a guess about what is acceptable. We could
+            // do something a lot more sophisticated (e.g. take into consideration the size
+            // of the path) but this is probably good enough.
+            const double acceptableError = 0.005;
+
             offset = b[0].ControlPoint0 - a[0].ControlPoint0;
+
+            // The first bezier in a is an offset of b if all of the control points
+            // in a are offset from b by the same amount.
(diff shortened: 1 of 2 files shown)


source/LottieToWinComp/Paths.cs
```diff
@@ -344,20 +344,33 @@ static bool TryGetPathTranslation(Sequence<BezierSegment> a, Sequence<BezierSegm
...
+                !IsFuzzyEqual(b[0].ControlPoint2 - a[0].ControlPoint2, offset, acceptableError) ||
+                !IsFuzzyEqual(b[0].ControlPoint3 - a[0].ControlPoint3, offset, acceptableError))
+            {
+                offset = default;
+                return false;
+            }
+
+            // Compare all of the points in the sequence of beziers to see if they are all offset
+            // by the same amount.
             for (var i = 1; i < a.Count; i++)
```
raymond-wh-leung, line 367: >var i = 1 [](start = 17, length = 9)

Can i starts from 0 to take care of index 0? (instead of outside of the for loop?)

simeoncran: Yeah, I agree. I was avoiding the unnecessary compare of ControlPoint0 for the first Bezier, but the extra complexity isn't worth it.

---
In reply to: [494442893](https://github.com/windows-toolkit/Lottie-Windows/pull/356#discussion_r494442893) [](ancestors = 494442893)

raymond-wh-leung (approved):
:shipit:

derobinsMSFT (approved):
:shipit:

PR #357, opened 2020-09-25
Cppwinrt-consistent file naming

simeoncran:
This change makes the file names produced by LottieGen match the pattern used by cppwinrt.exe in VS projects. Specifically, the file names are based on the fully-qualified name of the class that they are being generated for. This means that, for example, the Foo.Bar.Grotz class will be generated in: Foo.Bar.Grotz.h and Foo.Bar.Grotz.cpp, and it will reference the cppwinrt.exe-generated files Foo.Bar.Grotz.g.h and Foo.Bar.Grotz.g.cpp.

However, if the namespace of the generated class matches the "default namespace" of the VS project, cppwinrt sets a parameter to strip the namespace off the file name. To be consistent with that behavior LottieGen now has a "-RootNamespace" parameter (named that way to be consistent with the naming of the .targets file parameter used by cppwinrt) which is the namespace prefix that will be stripped from filename. So, for example, if "-RootNamespace Foo", the generated files will be Bar.Grotz.h and Bar.Grotz.cpp.


derobinsMSFT:
>         // The -Public switch is ignored for cppwinrt.

Nit: may as well add a comment like this below.

---
Refers to: LottieGen/CommandLineOptions.cs:125 in 0a4eee7. [](commit_id = 0a4eee77f070462732c64cbd51e2d1dd7342cef4, deletion_comment = False)

derobinsMSFT:
>         // The -Public switch is ignored for cppwinrt.

That is, pointing out that RootNamespace is ignored for any language but cppwinrt.

---
In reply to: [699074881](https://github.com/windows-toolkit/Lottie-Windows/pull/357#issuecomment-699074881) [](ancestors = 699074881)

---
Refers to: LottieGen/CommandLineOptions.cs:125 in 0a4eee7. [](commit_id = 0a4eee77f070462732c64cbd51e2d1dd7342cef4, deletion_comment = False)

derobinsMSFT (approved):
:shipit:

PR #358, opened 2020-09-26
Fix rendering issue caused by hole in the optimizer.

simeoncran:
An optimization stage was not taking into account the centerpoint property when moving properties up the tree. This was seen as a precomp that ended up offset from where it should have been.


derobinsMSFT (approved):
:shipit:

PR #359, opened 2020-09-29
Fixes a hole in the GraphCompactor optimizer wrt visiblity and masks.

simeoncran:
We use CompositionVisualSurface to implement masks. CompositionVisualSurface allows a Visual tree to be used as a brush, however it ignores many of the properties on the root of the Visual tree it hosts.
The GraphCompactor attempts to push IsVisible animations higher up the tree, however before this fix it could push IsVisible into a Visual that was the root of a CompositionVisualSurface's tree, with the result that the IsVisible property would get ignored.
The fix disallows pushing of IsVisible into a Visual that is at the root of a CompositionVisualSurface's tree.

### DIFF
--- a/source/UIData/Tools/GraphCompactor.cs
+++ b/source/UIData/Tools/GraphCompactor.cs
@@ -662,14 +662,14 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
             // CompositionVisualSurface (CompositionVisualSurface ignores
             // IsVisible on its SourceVisual).
             var visualsWithSingleParents =
-                (from n in graph.CompositionObjectNodes
-                 let visual = n.Object as Visual
-                 let parent = n.Node.Parent as ContainerVisual
-                 where visual != null &&
-                       parent != null &&
-                       parent.Children.Count == 1 &&
-                       !IsVisualSurfaceSourceVisual(graph, parent)
-                 select (visual, parent)).ToArray();
+                from n in graph.CompositionObjectNodes
+                let visual = n.Object as Visual
+                where visual != null
+                let parent = n.Node.Parent as ContainerVisual
+                where parent != null &&
+                      parent.Children.Count == 1 &&
+                      !IsVisualSurfaceSourceVisual(graph, parent)
+                select (visual, parent);
 
             foreach (var (visual, parent) in visualsWithSingleParents)
             {
--- a/source/UIData/Tools/GraphCompactor.cs
+++ b/source/UIData/Tools/GraphCompactor.cs
@@ -656,17 +656,30 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
         // animation up to the top of the chain.
         void PushVisualVisibilityUp(ObjectGraph<Node> graph)
         {
-            var visualsWithSingleParents = graph.CompositionObjectNodes.Where(n =>
-                n.Object is Visual &&
-                n.Node.Parent != null &&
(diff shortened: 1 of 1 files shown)


source/UIData/Tools/GraphCompactor.cs
```diff
@@ -656,17 +656,30 @@ bool TryElideContainerVisual(ObjectGraph<Node> graph, ContainerVisual container)
...
+            // IsVisible on its SourceVisual).
+            var visualsWithSingleParents =
+                (from n in graph.CompositionObjectNodes
+                 let visual = n.Object as Visual
+                 let parent = n.Node.Parent as ContainerVisual
+                 where visual != null &&
+                       parent != null &&
+                       parent.Children.Count == 1 &&
+                       !IsVisualSurfaceSourceVisual(graph, parent)
+                 select (visual, parent)).ToArray();
```
eliezerpMS, line 672: >.ToArray() [](start = 41, length = 10)

Maybe a dumb question but do we need "ToArray" for the foreach below?

source/UIData/Tools/GraphCompactor.cs
```diff
@@ -1210,11 +1223,11 @@ void CoalesceOrthogonalContainerVisuals(ObjectGraph<Node> graph)
...
 
-        // True iff the given ContainerVisual is the SourceVisual of a CompositionVisualSurface.
-        // In this case the transforming properties (e.g. offset) will be ignored, so it is not
-        // safe to hoist any such properties from its child.
-        static bool IsVisualSurfaceRoot(ObjectGraph<Node> graph, ContainerVisual containerVisual)
-         => graph[containerVisual].InReferences.Any(vertex => vertex.Node.Object is CompositionVisualSurface);
+        // True iff the given Visual is the SourceVisual of a CompositionVisualSurface.
+        // In this case the transforming properties (e.g. offset) and visiblity will be ignored,
+        // so it is not safe to hoist any such properties from its child.
+        static bool IsVisualSurfaceSourceVisual(ObjectGraph<Node> graph, Visual visual)
```
eliezerpMS, line 1229: >IsVisualSurfaceSourceVisual [](start = 20, length = 27)

Nit: Do other visuals/composition objects have sources that can be surfaces? (I think effects do?) would it makes sense to have this be something more like "IsCompositionVisualSurfaceSource"?

source/UIData/Tools/GraphCompactor.cs
simeoncran: No, I don't think we do. But it saves having to think about it. The ToArray guarantees that the list is stable by the time we do the foreach. If we don't have the ToArray you'd need to be certain that the work in the foreach cannot cause the evaluation of the list to change, or that it changes in an acceptable way.
So I tend to have some redundant ToArrays around the code. I think it's safe to remove in this case, so I'll remove it.

---
In reply to: [497073981](https://github.com/windows-toolkit/Lottie-Windows/pull/359#discussion_r497073981) [](ancestors = 497073981)

source/UIData/Tools/GraphCompactor.cs
simeoncran: Adding the "Composition" at the front is reasonable... I didn't do it only because it made the name really long and didn't seem to help readability (in fact it makes it less readable in one regard because "Composition" could be read in various ways).
I don't think I understand the first part of the question though.
As you can see, I changed the name because on a fresh reading I found it confusing (because Root is a new concept and it could also be read in various ways).
I think the name it has now is a good compromise - it's pretty hard to misconstrue. What do you think?


---
In reply to: [497076077](https://github.com/windows-toolkit/Lottie-Windows/pull/359#discussion_r497076077) [](ancestors = 497076077)

eliezerpMS (approved):
:shipit:

PR #360, opened 2020-10-01
Simplifications around image loading, and C++ generated code cleanup.

simeoncran:


source/LottieToWinComp/Images.cs
```diff
@@ -42,10 +42,12 @@ public static LayerTranslator CreateImageLayerTranslator(ImageLayerContext conte
                 case ImageAsset.ImageAssetType.Embedded:
                     var embeddedImageAsset = (EmbeddedImageAsset)imageAsset;
                     surface = LoadedImageSurface.StartLoadFromStream(embeddedImageAsset.Bytes);
+                    surface.SetName(imageAsset.Id);
```
simeoncran, line 45: Set a description here as well so that comments will be generated explaining what the big generated array is for.

source/UIDataCodeGen/CodeGen/CodegenConfiguration.cs
```diff
@@ -71,7 +71,7 @@ sealed class CodegenConfiguration
         /// The root namespace for the project that will consume the generated code.
         /// This affects the names used to reference files generated by cppwinrt.exe.
         /// </summary>
-        public string RootNamespace { get; set; }
+        public string RootNamespace { get; set; } = string.Empty;
```
eliezerpMS, line 74: > = string.Empty; [](start = 49, length = 16)

Stupid question: Why is it necessary to initialize it to Empty? Shouldn't users assume that a string can be null or Empty?

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -226,6 +226,14 @@ protected IReadOnlyList<string> AutoGeneratedHeaderText
             CodeBuilder builder,
             IAnimatedVisualInfo info);
 
+        /// <summary>
+        /// Write a byte array field.
+        /// </summary>
+        /// <param name="builder">A <see cref="CodeBuilder"/> used to create the code.</param>
+        /// <param name="fieldName">The name of the field to be written.</param>
+        /// <param name="bytes">The bytes in the array.</param>
+        protected abstract void WriteByteArrayField(CodeBuilder builder, string fieldName, byte[] bytes);
```
eliezerpMS, line 235: >CodeBuilder builder, string fieldName, byte[] bytes [](start = 52, length = 51)

nit: separate lines for params

eliezerpMS (approved):
:shipit:

source/UIDataCodeGen/CodeGen/CodegenConfiguration.cs
simeoncran: It just saves the user from having to differentiate those cases. I hit a codepath that wasn't checking for null, so rather than making the code more complex there I just set it here.
A null namespace doesn't make sense. Ideally this would be a non-nullable string. Let me see if I can get that enabled (we couldn't use nullable reference types previously because we didn't have the latest version of C#, but it might work now).

---
In reply to: [498460566](https://github.com/windows-toolkit/Lottie-Windows/pull/360#discussion_r498460566) [](ancestors = 498460566)

simeoncran: Will enable nullable reference types in a separate PR... doing it now but it changes too much stuff to be part of this PR.

---
In reply to: [498475365](https://github.com/windows-toolkit/Lottie-Windows/pull/360#discussion_r498475365) [](ancestors = 498475365,498460566)

PR #363, opened 2020-10-08
Fully enable nullables.

simeoncran:


source/GenericData/GenericDataList.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable // Temporary while enabling nullable everywhere.
+#nullable enable
```
derobinsMSFT, line 5: Weren't nullable types already enabled throughout the codebase? Why are these still needed?

simeoncran: They suppress some error message from Intellisense. The problem is that our Lottie-Windows.sln has the same files being built multiple ways, so Intellisense gets confused and reports problems that aren't real. Sometimes, enabling nullable explicitly like this made things better.
It's possible removing these will be fine now that all of the files are fixed, but rather that wrestling with it more I just left them in there.
I suspect the problem is that .NET Native turns off nullable by default. We turn it on, but that isn't sufficient to enable nullable for code analysis passes on .NET Native projects. Turning nullable in a file turns it on for everyone, which makes code analysis happier.
Maybe there's a way to turn on nullable for code analysis on .NET Native but we can solve that later.

---
In reply to: [503035336](https://github.com/windows-toolkit/Lottie-Windows/pull/363#discussion_r503035336) [](ancestors = 503035336)

derobinsMSFT: Thanks for the explanation!

---
In reply to: [503479554](https://github.com/windows-toolkit/Lottie-Windows/pull/363#discussion_r503479554) [](ancestors = 503479554,503035336)

derobinsMSFT (approved):
:shipit:

PR #364, opened 2020-10-09
Go back to using latest .NET 3.1 SDK.

simeoncran:
We'd previously had to stick to 3.1.302 due to build issues.


derobinsMSFT (approved):
:shipit:

PR #365, opened 2020-10-13
Add ability to specify additional interfaces for the generated code.

simeoncran:
This adds a feature to LottieGen to specify interfaces that LottieGen will then claim that the generated class implements.

This feature allows you to access methods that we generated on the class that are not members of IAnimatedVisualSource. This provides a way for us to add new features to the generated code, that can be accessed without knowing the exact type of the generated class, and without us having to update the IAnimatedVisualSource interface.

For example, I am about to add some metadata methods:
  double GetMarkerAsProgress(string markerName);
  void SetColorBinding(string bindingName, Color color);

Cool, but how do you access those methods? If you know the type of the generated class (e.g. "MyCoolLottie") then you can just call the methods on a variable of that type. But what if you want to call one of these methods without knowing the type? Or maybe you want to treat all your generated code in the same way, so the types get in the way.
The solution is to tell LottieGen that the generated code supports an additional interface that you will define to contain the new methods. LottieGen will then say "sure, the generated code implements "MyCoolAdditionalInterface". As long as your interface's methods are the same as the ones that we generate, you'll now be able to use MyCoolAdditionalInterface to call those methods.


source/UIDataCodeGen/CodeGen/CSharp/CSharpInstantiatorGenerator.cs
```diff
@@ -396,7 +406,26 @@ string GetDefaultPropertyBindingValue(PropertyBinding prop)
...
+            {
+                builder.WriteLine($"        : DependencyObject");
+                builder.WriteLine("        , IDynamicAnimatedVisualSource");
+            }
+            else
+            {
+                builder.WriteLine("        : IDynamicAnimatedVisualSource");
+            }
+
+            foreach (var itfc in SourceInfo.AdditionalInterfaces)
```
derobinsMSFT, line 423: >itfc [](start = 25, length = 4)

Nit: "iface"? Might just be a personal preference, but it took me a minute to realize what "itfc" meant-- I think "iface" would be a bit clearer.

derobinsMSFT (approved):
:shipit:

Issue #366, opened 2020-10-14
LottieVisualSource - allow for SetSourceAsync(string json)


alex-harper:
Having the ability to control the JSON source directly would be incredibly useful - this is available on the airbnb lottie players. A use case would be that we can add basic support for [dotLottie](https://dotlottie.io/) to our player project, passing in the JSON once extracted from the archive.

simeoncran:
I don't want to expose a string overload only because the Stream overload of SetSourceAsync(...) is more general and can be easily used to load from a string. But it turns out that our Stream overload was being hidden by mistake. #368 addresses that, and with that fix you can load from a string like this:

using System.IO;
using System.Text;

await lottieVisualSource.SetSourceAsync(new MemoryStream(Encoding.UTF8.GetBytes(str)).AsInputStream());

I would expect that when you are loading from a dotLottie file you'll have a Stream anyway and the conversion to string will be unnecessary. But if you need to use a string, the code above should do it.

simeoncran:
Closing this because it is addressed now. You can try it out via nuget at https://dotnet.myget.org/F/uwpcommunitytoolkit/api/v3/index.json
with 7.0.0-build.56 or later.

PR #368, opened 2020-10-16
Expose overload of SetSourceAsync that takes a stream.

simeoncran:
We always had this overload, however due to a misunderstanding about the rules around overload attributes for WinRT, it was never visible in the WinMD.
This should address #366.


peterjtan (approved):
LGTM 👍

PR #370, opened 2020-10-17
Code cleanup: take advantage of C# 8 features.

simeoncran:
Also adds some more #nullable enable annotation so quieten code analysis.

None of these changes should have changed any behaviors.


derobinsMSFT:
> // Licensed to the .NET Foundation under one or more agreements.

If all of these error suppressions are gone, is this file needed anymore? Or does Code Analysis always expect to find this file even if it's "empty"?

---
Refers to: GlobalSuppressions.cs:1 in 1c4bdab. [](commit_id = 1c4bdab0a9aa129eeb1254306534c8273eadda43, deletion_comment = False)

source/YamlData/YamlWriter.cs
```diff
@@ -192,14 +192,10 @@ bool TryInlineObject(YamlObject obj, int maximumWidth, [MaybeNullWhen(false)] ou
             {
                 switch (obj.Kind)
```
derobinsMSFT, line 193: >switch (obj.Kind) [](start = 16, length = 17)

`return (obj.Kind) switch`?

derobinsMSFT (approved):
:shipit:

source/YamlData/YamlWriter.cs
simeoncran: I tried, and failed. I think it's hitting a limitation of the nullable analysis. Even though the TryInline methods all return a string with MaybeNullWhen(false), the nullable analysis complains if this is a switch expression.

---
In reply to: [508131986](https://github.com/windows-toolkit/Lottie-Windows/pull/370#discussion_r508131986) [](ancestors = 508131986)

PR #371, opened 2020-10-17
Fix codegen errors.

simeoncran:
Fixes needed after first tests on CX since adding the markers and property support.


derobinsMSFT (approved):
:shipit:

PR #372, opened 2020-10-20
Cleanup resulting from enabling nullables.

simeoncran:
Replace Asserts that are now handled by nullable analysis.
Add some more "#nullable enable" to quieten code analysis in VS.
Rename a method in CodeBuilder because it referred to Cpp but is now used by C# as well.


source/UIData/Tools/GraphCompactor.cs
```diff
@@ -1340,7 +1340,8 @@ void CoalesceOrthogonalVisuals(ObjectGraph<Node> graph)
 
         static void TransferShapeProperties(CompositionShape from, CompositionShape to)
         {
-            void TransferProperty<T>(Func<CompositionShape, T> get, Action<CompositionShape, T> set)
```
derobinsMSFT, line 1343: >TransferProperty [](start = 17, length = 16)

Curious about the changes in this file-- why was this method separated? The new methods appear identical other than what type `T` is allowed to be.

derobinsMSFT (approved):
:shipit:

source/UIData/Tools/GraphCompactor.cs
simeoncran: It's to allow T? into the type signature. As soon as you have a T?, you have to differentiate between value types and reference types, otherwise it's not clear to the compiler what the ? means. So I add the ?, and added a constraints for the class and struct versions. 
You can't overload methods based on constraints, so it also required giving the methods distinct names.

---
In reply to: [508724228](https://github.com/windows-toolkit/Lottie-Windows/pull/372#discussion_r508724228) [](ancestors = 508724228)

PR #374, opened 2020-10-22
Allow external images to be loaded, and add .lottie file support

simeoncran:
External images (i.e. images that are not embedded in the .JSON file) are a problem for LottieVisualSource because it doesn't know how to get them. For securit reasons, we don't want LottieVisualSource to have the ability to read files that are referenced from a Lottie .json file, and in many cases LottieVisualSource wouldn't be able to resolve the file because it doesn't have permissions.

This change adds an ImageAssetDelegate that the user can set on the AnimatedVisualSource in order to handle reading of external images.

This change also adds support for the .lottie format (see https://dotlottie.io). There is a parser for .lottie files, and support in LottieViewer for displaying Lotties from a .lottie file. Because .lottie files package images together with the .json Lottie file, it enables LottieViewer to display Lotties that have external images.

### DIFF
--- a/source/DotLottie/DotLottieFile.cs
+++ b/source/DotLottie/DotLottieFile.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.DotLottie
 {
     /// <summary>
     /// Provides access to the contents of a .lottie file.
+    /// See https://dotlottie.io.
     /// </summary>
     sealed class DotLottieFile : IDisposable
     {
--- a/source/DotLottie/DotLottieFile.cs
+++ b/source/DotLottie/DotLottieFile.cs
@@ -1,0 +1,398 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.DotLottie
+{
+    /// <summary>
+    /// Provides access to the contents of a .lottie file.
+    /// </summary>
+    sealed class DotLottieFile : IDisposable
+    {
+        readonly Manifest _manifest;
+
+        DotLottieFile(ZipArchive zipArchive, Manifest manifest)
+        {
+            ZipArchive = zipArchive;
+            _manifest = manifest;
+            Animations = manifest.Animations.Select(a =>
+                new DotLottieFileAnimation(
+                        this,
+                        a.Id,
+                        a.Loop)).ToArray();
+        }
+
+        /// <summary>
+        /// The name of the tool that generated the .lottie file.
+        /// </summary>
+        public string? Generator => _manifest.Generator;
+
+        /// <summary>
+        /// The .lottie file format version.
+        /// </summary>
+        public string Version => _manifest.Version;
+
+        /// <summary>
+        /// The author of the .lottie file.
+        /// </summary>
+        public string? Author => _manifest.Author;
+
+        /// <summary>
+        /// Returns the animations that are contained in the .lottie file.
+        /// </summary>
+        public IReadOnlyList<DotLottieFileAnimation> Animations { get; }
+
+        /// <summary>
+        /// Opens a file given its path in the .lottie file.
+        /// </summary>
+        /// <param name="path">The path to the file.</param>
+        /// <returns>A stream containing the file, or null if the file is not found.</returns>
+        public Stream? OpenFile(string path)
+        {
+            var zipEntry = GetEntryForPath(path);
+
+            return zipEntry is null ? null : zipEntry.Open();
+        }
+
+        /// <summary>
+        /// Opens a file given its path in the .lottie file.
+        /// </summary>
+        /// <param name="path">The path to the file.</param>
+        /// <returns>A stream containing the file, or null if the file is not found.</returns>
+        public MemoryStream? OpenFileAsMemoryStream(string path)
+        {
+            var zipEntry = GetEntryForPath(path);
+
+            if (zipEntry is null)
+            {
+                return null;
+            }
+
+            var buffer = new byte[zipEntry.Length];
+            using (var zipEntryStream = zipEntry.Open())
+            {
+                if (zipEntryStream.Read(buffer, 0, buffer.Length) != buffer.Length)
+                {
+                    return null;
+                }
+            }
+
+            return new MemoryStream(buffer, writable: false);
+        }
+
+        /// <summary>
+        /// Returns a <see cref="DotLottieFile"/> that reads from the given <see cref="ZipArchive"/>.
+        /// </summary>
+        /// <param name="zipArchive">The <see cref="ZipArchive"/>.</param>
+        /// <returns>A <see cref="DotLottieFile"/> or null on error.</returns>
+        public static async Task<DotLottieFile?> FromZipArchiveAsync(
+            ZipArchive zipArchive)
+        {
+            // The manifest contains information about the animation.
+            var manifestEntry = zipArchive.GetEntry("manifest.json");
+
+            if (manifestEntry is null)
+            {
+                // Not a valid .lottie file.
+                return null;
+            }
+
+            var manifestBytes = new byte[manifestEntry.Length];
+            using var manifestStream = manifestEntry.Open();
+            var bytesRead = await manifestStream.ReadAsync(
+                                    manifestBytes,
+                                    0,
+                                    manifestBytes.Length);
+            if (bytesRead != manifestBytes.Length)
+            {
+                // Failed to read the manifest
+                return null;
+            }
+
+            var manifest = Manifest.ParseManifest(manifestBytes);
+
+            return manifest is null ? null : new DotLottieFile(zipArchive, manifest);
+        }
+
+        public void Dispose()
+        {
+            ZipArchive.Dispose();
+        }
+
+        internal ZipArchive ZipArchive { get; }
+
+        ZipArchiveEntry GetEntryForPath(string path)
+        {
+            if (path.StartsWith("/"))
+            {
+                path = path.Substring(1);
+            }
+
+            return ZipArchive.GetEntry(path);
+        }
+
+        static void ConsumeToken(ref Utf8JsonReader reader)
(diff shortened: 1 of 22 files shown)


source/Lottie/DotLottieLoader.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// </summary>
+    sealed class DotLottieLoader : Loader
+    {
+        DotLottieFile? _dotLottieFile;
+
+        DotLottieLoader()
+        {
+        }
+
+        [return: NotNullIfNotNull("stream")]
```
simeoncran, line 30: >stream [](start = 35, length = 6)

"file"

source/Lottie/DotLottieLoader.cs
```diff
@@ -0,0 +1,104 @@
...
+        internal static async Task<AnimatedVisualFactory>? LoadAsync(
+            StorageFile file,
+            LottieVisualOptions options)
+        {
+            var stream = (await file.OpenReadAsync()).AsStreamForRead();
+            return await LoadAsync(file.Name, stream, options);
+        }
+
+        [return: NotNullIfNotNull("stream")]
```
simeoncran, line 39: >NotNullIfNotNull [](start = 17, length = 16)

Attribute doesn't belong here.

source/Lottie/StorageFileLoader.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            var loader = new StorageFileLoader(imageLoader, file);
+            return Loader.LoadAsync(
+                loader.GetJsonStreamAsync,
+                loader,
+                options);
+        }
+
+        // Starts loading from an ms-appx asset file. This loads embedded assets.
+        internal static async Task<AnimatedVisualFactory>? StartLoadingAsync(
```
simeoncran, line 57: LoadAsync.

source/DotLottie/DotLottieFile.cs
```diff
@@ -0,0 +1,398 @@
...
+            }
+        }
+
+        sealed class Animation
+        {
+            Animation(string id, double speed, string themeColor, bool loop)
+            {
+                Id = id;
+                Speed = speed;
+                ThemeColor = themeColor;
```
raymond-wh-leung, line 158: >ThemeColor [](start = 16, length = 10)

Curious what this ThemeColor property is used for, for an animation.

source/DotLottie/DotLottieFileAnimation.cs
```diff
@@ -0,0 +1,33 @@
...
+            bool loop)
+        {
+            _owner = owner;
+            Id = id;
+            Loop = loop;
+        }
+
+        public string Id { get; }
+
+        public bool Loop { get; }
```
raymond-wh-leung, line 27: Should this have the Speed and the ThemeColor properties?

simeoncran: Maybe eventually. I am in conversation with the lottiefiles people about those features because I don't understand the intention of them. For now I'm parsing it but ignoring it.

---
In reply to: [510498455](https://github.com/windows-toolkit/Lottie-Windows/pull/374#discussion_r510498455) [](ancestors = 510498455)

source/DotLottie/DotLottieFile.cs
simeoncran: I'm also curious. Waiting to hear from lottiefiles to determine what to do with it. I'm guessing it's the default background color that players should use.

---
In reply to: [510497000](https://github.com/windows-toolkit/Lottie-Windows/pull/374#discussion_r510497000) [](ancestors = 510497000)

raymond-wh-leung (dismissed):
:shipit:

source/Lottie/ImageAssetHandler.cs
```diff
@@ -0,0 +1,18 @@
...
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+using System;
+using Windows.UI.Composition;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie
+{
+    /// <summary>
```
derobinsMSFT, line 12: ><summary> [](start = 8, length = 9)

Documentation suggestions: Does one need to provide an ImageAssetHandler? What are the benefits/use cases of including one?

source/Lottie/InputStreamLoader.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Threading.Tasks;
+using Windows.Storage.Streams;
+using Windows.UI.Composition;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie
+{
+    /// <summary>
+    /// A loader that loads from an IInputStream.
```
derobinsMSFT, line 17: >IInputStream [](start = 36, length = 12)

Missing cref?

source/Lottie/Loader.cs
```diff
@@ -317,22 +239,6 @@ public TimeSpan GetElapsedAndRestart()
...
-            await Task.Delay(_checkedDelayMs);
-            await task;
-            await Task.Delay(_checkedDelayMs);
-#else
-#pragma warning disable VSTHRD003 // Avoid awaiting foreign Tasks
-            return task;
-#pragma warning restore VSTHRD003 // Avoid awaiting foreign Tasks
-#endif
-        }
+        public abstract void Dispose();
```
derobinsMSFT, line 242: >public abstract void Dispose(); [](start = 8, length = 31)

Nit: can this be a virtual that does nothing instead? Most of the overrides appear to be empty no-ops.

source/WinCompData/MetaData/UapVersionAttribute.cs
```diff
@@ -33,8 +33,8 @@ public UapVersionAttribute(int version)
             //  7 = 1809 / 10.0.17763 / RS5  / October 2018 Update
             //  8 = 1903 / 10.0.18362 / 19H1 / May 2019 Update
             //  9 = 1909 / 10.0.18363 / 19H2 / November 2019 Update
-            //  10 = 2004 / 10.0.19041 / 20H1 / ?????
-            //  11 = ???? / 10.0.????? / 20H2 / ?????
+            //  10 = 2004 / 10.0.19041 / 20H1 / May 2020 Update
+            //  11 = 20H2 / 10.0.19042 / 20H2 / October 2020 Update
```
scott-moore-ms, line 37: I want this table for my own reference 🤣

source/Lottie/Loader.cs
simeoncran: It could, but I like the idea of forcing subclasses to make a statement about what they need to dispose, even if it's nothing. Also, consider that if you extend from a Disposable class that implements the Dispose() methods, you also need to call base.Dispose() in your Dispose implementation, whereas if the method you're overriding is abstract then it's clear that there's no need to call the base class method.

---
In reply to: [511033242](https://github.com/windows-toolkit/Lottie-Windows/pull/374#discussion_r511033242) [](ancestors = 511033242)

derobinsMSFT: Ah, true, good point.

---
In reply to: [511038186](https://github.com/windows-toolkit/Lottie-Windows/pull/374#discussion_r511038186) [](ancestors = 511038186,511033242)

source/DotLottie/DotLottieFile.cs
```diff
@@ -0,0 +1,398 @@
...
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.DotLottie
+{
+    /// <summary>
+    /// Provides access to the contents of a .lottie file.
```
derobinsMSFT, line 18: >lottie [](start = 46, length = 6)

Nit: can you link to [https://dotlottie.io/structure](https://dotlottie.io/structure) in the description here? Just for some extra reference.

derobinsMSFT (approved):
:shipit:

source/Lottie/DotLottieLoader.cs
```diff
@@ -0,0 +1,110 @@
...
+        [return: NotNullIfNotNull("file")]
+        internal static async Task<AnimatedVisualFactory?> LoadAsync(
+            StorageFile file,
+            LottieVisualOptions options)
+        {
+            var stream = (await file.OpenReadAsync()).AsStreamForRead();
+            var factory = LoadAsync(file.Name, stream, options);
+            if (factory == null)
+            {
+                // This won't work... we failed, but we already returned a Task.
```
derobinsMSFT, line 39: >// This won't work... we failed, but we already returned a Task. [](start = 16, length = 64)

Not entirely sure what you mean here. We haven't tried to return a Task from this method at this point. Could you elaborate?

source/Lottie/AnimatedVisualFactory.cs
```diff
@@ -19,19 +20,22 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie
...
+        Loader? _imageLoader;
         WinCompData.Visual? _wincompDataRootVisual;
         WinCompData.CompositionPropertySet? _wincompDataThemingPropertySet;
         CompositionPropertySet? _themingPropertySet;
         double _width;
         double _height;
         TimeSpan _duration;
 
-        internal ContentFactory(LottieVisualDiagnostics? diagnostics)
+        internal AnimatedVisualFactory(Loader imageLoader, LottieVisualDiagnostics? diagnostics)
```
derobinsMSFT, line 36: >Loader [](start = 39, length = 6)

Should this be `Loader?`. Otherwise this won't be null.

simeoncran: No, otherwise that would allow null to be passed as the loader. The reason that the field _imageLoader is nullable is because once we've finished loading we null it out so it can be garbage collected.
However, seeing as you drew my attention to it, the name should just be "loader" seeing as we require a Loader (used to be something else)

---
In reply to: [511093450](https://github.com/windows-toolkit/Lottie-Windows/pull/374#discussion_r511093450) [](ancestors = 511093450)

source/Lottie/DotLottieLoader.cs
simeoncran: That was a note to me when I was struggling with how to indicate failure. I previously had this returning Task<AnimatedVisualFactory>? with a null Task indicating failure, but you can't return null from a method marked "async", so it would have required some refactoring to get nullability correct. 

---
In reply to: [511080916](https://github.com/windows-toolkit/Lottie-Windows/pull/374#discussion_r511080916) [](ancestors = 511080916)

derobinsMSFT (approved):
:shipit:

PR #375, opened 2020-10-23
Show the current UAP version in Lottie Viewer.

simeoncran:
Because Lotties can render differently on different UAP versions, it's important that users know what version of UAP they're using. That way, when they say "this doesn't look right", and I say "looks fine to me", we can check the UAP version easily to determine whether that accounts for the difference.


derobinsMSFT (approved):
:shipit:

PR #376, opened 2020-10-25
Support .lottie files in LottieGen.

simeoncran:
See https://dotlottie.io for file details.
LottieGen will now read .lottie files as well as .json files.


LottieGen/FileProcessor.cs
```diff
@@ -0,0 +1,261 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using Microsoft.Toolkit.Uwp.UI.Lottie.DotLottie;
+
+/// <summary>
+/// Processes a file containing an animation and produces various generated outputs.
```
simeoncran, line 12: Can contain more than one animation.

LottieGen/FileProcessor.cs
```diff
@@ -0,0 +1,261 @@
...
+            case ".lottie":
+                return ProcessDotLottieFile(
+                    options: _options,
+                    reporter: _reporter,
+                    filePath: _filePath,
+                    outputFolder: _outputFolder,
+                    timestamp: _timestamp);
+
+            case ".json":
+            default:
```
derobinsMSFT, line 68: It's not an error to have a file that doesn't end in `.json` or `.lottie`?

derobinsMSFT (approved):
:shipit:

LottieGen/FileProcessor.cs
simeoncran: Yeah, on purpose. Nothing requires JSON files to be .json. However .lottie files do need to be named with .lottie.

---
In reply to: [512253491](https://github.com/windows-toolkit/Lottie-Windows/pull/376#discussion_r512253491) [](ancestors = 512253491)

PR #377, opened 2020-10-27
Parse layer effects.

simeoncran:
Enable parsing of layer effects. This only fully supports drop shadow effects, however the data model is set up so that other effects can be supported in the future.


source/LottieData/DropShadowEffect.cs
```diff
@@ -0,0 +1,25 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData
+{
+    /// <summary>
+    /// An effect applied to a layer.
```
simeoncran, line 11: "drop shadow"

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        DropShadowEffect ReadDropShadoEffect(in LottieJsonObjectElement obj)
+        {
+            obj.IgnorePropertyIntentionally("ix");
+            obj.IgnorePropertyIntentionally("mn");
+            obj.IgnorePropertyIntentionally("n");
+
+            var effectName = obj.StringPropertyOrNull("nm") ?? string.Empty;
+
+            var isEnabled = obj.BoolPropertyOrNull("en") ?? true;
```
derobinsMSFT, line 78: >true [](start = 60, length = 4)

`false`?

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            // TODO - what if there are no parameters? Need to throw or somethign.
```
derobinsMSFT: >somethign [](start = 72, length = 9)

"something"

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+            var effectType = obj.DoublePropertyOrNull("ty");
+
+            switch (effectType)
+            {
+                // DropShadows are type 25. This is the only type we currently support.
+                case 25:
+                    return ReadDropShadoEffect(obj);
+
+                default:
+                    obj.IgnorePropertyThatIsNotYetSupported(
```
derobinsMSFT, line 51: >IgnorePropertyThatIsNotYetSupported [](start = 24, length = 35)

Why not swap out the overloads of this method with one method with a `params` array? Or, alternatively, add a 6-parameter overload so you don't have to call this method twice here.

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+                    {
+                        result.AddItemIfNotNull(ReadEffect(effectObject.Value, layerName));
+                    }
+                }
+            }
+
+            return result.ToArray();
+        }
+
+        // See: https://github.com/airbnb/lottie-web/blob/master/docs/json/effects/layer.json.
```
derobinsMSFT, line 39: >layer.json [](start = 83, length = 10)

This file doesn't have all of the properties listed below-- in particular, it's missing "n". There's also another value in this file ("v") that's not referenced below.

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+                        "ix");  // index.
+                    obj.IgnorePropertyThatIsNotYetSupported(
+                        "ef"); // effect parameters
+
+                    _issues.LayerEffectsIsNotSupported(layerName);
+                    return null;
+            }
+        }
+
+        DropShadowEffect ReadDropShadoEffect(in LottieJsonObjectElement obj)
```
derobinsMSFT: >Shado [](start = 33, length = 5)

"Shadow"

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+                var p = parameters!.Value[i].AsObject();
+                var value = p!.Value.ObjectPropertyOrNull("v");
+
+                switch (i)
+                {
+                    case 0:
+                        color = ReadAnimatableColor(value);
+                        break;
+
+                    // Opacity.
```
derobinsMSFT, line 101: >. [](start = 30, length = 1)

Nit: no period? (Or periods below.)

source/LottieReader/Effects.cs
```diff
@@ -0,0 +1,138 @@
...
+
+                default:
+                    obj.IgnorePropertyThatIsNotYetSupported(
+                        "nm",   // name.
+                        "mn",   // match name.
+                        "en",   // enabled.
+                        "n",
+                        "ix");  // index.
+                    obj.IgnorePropertyThatIsNotYetSupported(
+                        "ef"); // effect parameters
```
derobinsMSFT, line 58: >parameters [](start = 41, length = 10)

NIt: period?

source/LottieReader/Effects.cs
simeoncran: This was intentional - if the "en" property isn't in the file for some reason then I want the effect to be enabled. There's no much point having an effect in the file if you don't use it.
I don't know whether we'll ever see an effect that doesn't have this property, but it's easy to be resilient to it missing.

---
In reply to: [513010186](https://github.com/windows-toolkit/Lottie-Windows/pull/377#discussion_r513010186) [](ancestors = 513010186)

source/LottieReader/Serialization/ArrayBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// <typeparam name="T">The type of item in the array.</typeparam>
+    ref struct ArrayBuilder<T>
+        where T : class
+    {
+        int _count;
+        T[]? _array;
+
+        public void AddItemIfNotNull(T? item)
+        {
+            if (!(item is null))
```
derobinsMSFT, line 25: >!(item is null) [](start = 16, length = 15)

`item != null`?

simeoncran: Yeah, that would be fine. I was being careful != can be overridden by classes but because T is generic that won't be a problem.

---
In reply to: [513028611](https://github.com/windows-toolkit/Lottie-Windows/pull/377#discussion_r513028611) [](ancestors = 513028611)

source/LottieReader/Effects.cs
derobinsMSFT: >// TODO - what if there are no parameters? Need to throw or somethign. [](start = 12, length = 70)

Not having parameters at all is functionally the same as these parameters not having values. Looks like that might be the plan below-- if so, might as well throw for this case too.

simeoncran: It does throw. You seem to be looking at an older revision of the file.

---
In reply to: [513032366](https://github.com/windows-toolkit/Lottie-Windows/pull/377#discussion_r513032366) [](ancestors = 513032366)

source/LottieReader/Effects.cs
simeoncran: Rather than copying around the same comment about the the schema being out of date, I've removed the comment. There is a comment about finding the schema in the main LottieCompositionReader class.
I don't want to add documentation here about how to figure out the schema, because it's true of all of the reader code, not just this.

---
In reply to: [513049690](https://github.com/windows-toolkit/Lottie-Windows/pull/377#discussion_r513049690) [](ancestors = 513049690,513031288,513016234)

derobinsMSFT (dismissed):
:shipit:

derobinsMSFT (approved):
:shipit:

PR #378, opened 2020-10-27
Cleanup: remove redundant type tags.

simeoncran:
These type tags evolved a little unevenly. The goal is that the tags represent each of the direct subclasses, but some of the tags were referring to subclasses of subclasses.


derobinsMSFT (approved):
:shipit:

derobinsMSFT (approved):
:shipit:

PR #379, opened 2020-11-04
Fix crash on some cases of path groups.

simeoncran:
As a result of looking again at the algorithm that we use to combine animated paths I decided it needed a lot more documentation and to be rewritten to make sure that we handle all cases.

### DIFF
--- a/source/LottieToWinComp/PathGeometryGroup.cs
+++ b/source/LottieToWinComp/PathGeometryGroup.cs
@@ -15,13 +15,10 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
     /// </summary>
     sealed class PathGeometryGroup : IEquatable<PathGeometryGroup>
     {
-        PathGeometryGroup(PathGeometry[] data, Easing easing)
+        PathGeometryGroup(PathGeometry[] data)
         {
             Data = data;
-            Easing = easing;
         }
-
-        public Easing Easing { get; }
 
         /// <summary>
         /// The geometries in the group.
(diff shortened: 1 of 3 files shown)


source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,314 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
-                    success = false;
+                    indices[i] = -1;
                 }
+            }
 
-                for (var i = 0; i < geometries.Length; i++)
+            // Get enumerators for the key frames in each of the Animatable<PathGeometry>s.
+            var enumerators = pathData.Select(EnumerateKeyFrames).ToArray();
+
+            // Start enumerating.
```
derobinsMSFT, line 196: >// Start enumerating. [](start = 12, length = 21)

Why are we skipping the first keyframe in each path? Is it because we don't need the starting values if the path is animated?

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,314 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
+                // always end up with a non-null preferred easing here.
+                Debug.Assert(preferredEasing != null, "Invariant");
+
+                yield return new PathGeometryGroupKeyFrames(currentFrame, geometries, preferredEasing!, easingIsCorrect);
+
+                // Advance the enumerators that are on the current frame unless they are completed.
+                for (var i = 0; i < enumerators.Length; i++)
+                {
+                    var enumerator = enumerators[i];
+                    if (!enumerator.Current.isCompleted &&
```
simeoncran, line 257: >!enumerator.Current.isCompleted && [](start = 24, length = 34)

I don't think this is necessary - the enumerators will happily output forever.

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,314 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
                 }
             }
 
-            // Make sure that every frame has a geometry from each path.
-            // For any missing path, fill the hole with the path from the
-            // previous frame.
-            var frames = groupsByFrame.OrderBy(kvp => kvp.Key).Select(kvp => (frame: kvp.Key, geometries: kvp.Value)).ToArray();
-            var previousGeometries = frames[0].geometries;
-            var success = true;
+            result =
```
derobinsMSFT, line 160: >result = [](start = 12, length = 8)

Add comment explaining what this assignment is doing.

source/LottieToWinComp/PathGeometryGroup.cs
simeoncran: We're not skipping anything. We are simply moving to the first item in the enumeration. Enumerators have to have MoveNext() called on them to get to the first item. If it didn't work that way there would be no way to handle empty enumerations.

---
In reply to: [518930487](https://github.com/windows-toolkit/Lottie-Windows/pull/379#discussion_r518930487) [](ancestors = 518930487)

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,319 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
-                    // any cases where this happens to determine if we should be trying
-                    // harder.
-                    Debug.Assert(array[i] is null, "Path key frames very close together");
+                        var currentFrameNumber = records[frameIndex].Frame;
+                        KeyFrame<PathGeometry>? previousKeyFrame = null;
+                        KeyFrame<PathGeometry>? nextKeyFrame = null;
+                        if (frameIndex > 0)
+                        {
+                            previousKeyFrame = records[frameIndex - 1].Geometries[pathIndex];
+                        }
```
derobinsMSFT, line 127: Maybe add another comment here explaining why we only look back one frame for the "previous" key frame but might look ahead by multiple frames for the "next" one-- just for extra clarity.

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,319 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
+                            closestKeyFrame = previousKeyFrame!;
+                        }
+                        else
+                        {
+                            closestKeyFrame = nextKeyFrame!;
+                        }
+
+                        // Fill in the null with the value from the closest key frame.
+                        records[frameIndex].Geometries[pathIndex] =
+                            new KeyFrame<PathGeometry>(currentFrameNumber, closestKeyFrame.Value, HoldEasing.Instance);
```
derobinsMSFT, line 155: >HoldEasing.Instance [](start = 98, length = 19)

Should this be whatever the preferred easing of the record is instead?

simeoncran: No, the easing doesn't matter here - the KeyFrame is just a convenient way to store a frame number an PathGeometry. When we create the Animatable<PathGeometryGroup> we'll use the PreferredEasing only.

---
In reply to: [520098584](https://github.com/windows-toolkit/Lottie-Windows/pull/379#discussion_r520098584) [](ancestors = 520098584)

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,321 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
         public bool Equals(PathGeometryGroup? other) => other != null && Enumerable.SequenceEqual(Data, other.Data);
 
-        // Takes a group of possibly-animated paths and returns an animatable
-        // of PathGeometryGroups. Returns true if it succeeds without issues.
-        // Even if false is returned a best-effort animatable is returned.
+        /// <summary>
+        /// Takes a group of possibly-animated paths and returns an animatable
+        /// of PathGeometryGroups. Returns true if it succeeds without issues.
+        /// Even if false is returned a best-effort animatable is returned.</summary>
+        /// <returns><c>true</c> iff the grouping worked perfectly.</returns>
```
patrickeganMSFT, line 34: Do you think it's worth it to add something about returning false in there <return> section? Not sure how likely it would be that someone would read only the return section and not the summary, but worth considering.

derobinsMSFT (dismissed):
:shipit:

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,321 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
+                    propertyIndex: null);
+
+            return success;
+        }
 
-            // Start from the second frame. The initial frame (0) will always have a value (.InitialValue).
-            foreach (var (frame, geometries) in frames.Skip(1))
+        /// <summary>
+        /// From a list of animatable paths, produce a record for each distinct key frame number.
+        /// Each record consists of the frame number a list of geometries for that frame number
```
benwestprog, line 177: >number [](start = 46, length = 6)

number,

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,321 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
-                // Get the easing for this frame.
-                var easings = geometries.Where(g => g != null).Select(g => g.Easing).Distinct().ToArray();
-                if (easings.Length > 1)
+                if (!pathData[i].IsAnimated)
                 {
-                    // There are conflicting easings. We can't currently handle that.
-                    success = false;
+                    indices[i] = -1;
                 }
+            }
```
benwestprog, line 195: Is this used anywhere?

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,321 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
 
-        // Takes a group of possibly-animated paths and returns an animatable
-        // of PathGeometryGroups. Returns true if it succeeds without issues.
-        // Even if false is returned a best-effort animatable is returned.
+        /// <summary>
+        /// Takes a group of possibly-animated paths and returns an animatable
+        /// of PathGeometryGroups. Returns true if it succeeds without issues.
+        /// Even if false is returned a best-effort animatable is returned.</summary>
+        /// <returns><c>true</c> iff the grouping worked perfectly.</returns>
         internal static bool TryGroupPaths(
```
benwestprog, line 35: >TryGroupPaths [](start = 29, length = 13)

The naming here threw me until I read the comments, expected null with a false return to match other common "Try" methods. Not sure what the local idiom is, but if this isn't consistent consider a more suggestive name, maybe leave out the "Try"? Or make the bool an out param with a helpful name?

source/LottieToWinComp/PathGeometryGroup.cs
```diff
@@ -15,139 +15,321 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
+            for (var pathIndex = 0; pathIndex < paths.Count; pathIndex++)
+            {
+                for (var frameIndex = 0; frameIndex < records.Length; frameIndex++)
                 {
-                    // See if there's a key frame at the frame number already.
-                    if (!groupsByFrame.TryGetValue(kf.Frame, out var array))
+                    // Get the key frame for the current frame number.
+                    var keyFrame = records[frameIndex].Geometries[pathIndex];
+                    if (keyFrame is null)
                     {
```
benwestprog, line 108: Consider moving this block into its own method, to make this one more readable. It's over three screens in my CodeFlow view.

benwestprog (dismissed):
:shipit:

source/LottieToWinComp/PathGeometryGroup.cs
simeoncran: Called from line 93

---
In reply to: [520956036](https://github.com/windows-toolkit/Lottie-Windows/pull/379#discussion_r520956036) [](ancestors = 520956036)

source/LottieToWinComp/PathGeometryGroup.cs
simeoncran: Great suggestion.

---
In reply to: [520959164](https://github.com/windows-toolkit/Lottie-Windows/pull/379#discussion_r520959164) [](ancestors = 520959164)

source/LottieToWinComp/PathGeometryGroup.cs
simeoncran: The renaming of the method will fix this.

---
In reply to: [520757936](https://github.com/windows-toolkit/Lottie-Windows/pull/379#discussion_r520757936) [](ancestors = 520757936)

source/LottieToWinComp/PathGeometryGroup.cs
simeoncran: Oh, you mean indices... Ooops. Old code.

---
In reply to: [522427828](https://github.com/windows-toolkit/Lottie-Windows/pull/379#discussion_r522427828) [](ancestors = 522427828,520956036)

benwestprog (approved):
:shipit:

derobinsMSFT (approved):
:shipit:

PR #382, opened 2020-11-10
Replace -winui3mode with -winuiversion.

simeoncran:
This will give us more flexibility (e.g. outputing IAnimatedIcon if WinUI 2.5 is being targeted).


LottieGen/CommandLineOptions.cs
```diff
@@ -74,11 +75,7 @@ sealed class CommandLineOptions
     // a previous version of the tool.
     internal bool TestMode { get; private set; }
 
-    // Experimental feature to control code generation for WinUI3. Eventually there
-    // will be separate versions of LottieGen for WinUI3 and system APIs so there will
-    // be no need for this switch. For now, if you're using LottieGen to generate
-    // code for WinUI3, set this switch and the codegen will need less hand fixing.
-    internal bool WinUI3Mode { get; private set; }
+    internal Version WinUIVersion => _winUIVersion ?? new Version(2, 4);
```
derobinsMSFT, line 78: >internal Version WinUIVersion [](start = 4, length = 29)

Add a comment explaining why 2.4 is the default version.

LottieGen/LottieJsonFileProcessor.cs
```diff
@@ -637,9 +637,7 @@ static bool TryEnsureDirectoryExists(string directoryPath)
 
         var result = new CodegenConfiguration(
             className: _className,
-            interfaceType: string.IsNullOrWhiteSpace(_options.Interface)
-                            ? "Microsoft.UI.Xaml.Controls.IAnimatedVisual"
-                            : _options.Interface,
+            interfaceType: _options.InterfaceBaseName!,
```
derobinsMSFT, line 640: >! [](start = 53, length = 1)

Necessary? This will never be null.

LottieGen/Program.cs
```diff
@@ -252,8 +252,10 @@ at runtime.
          -GenerateDependencyObject
                        Generates code that extends DependencyObject. This is useful
                        to allow XAML binding to properties in the Lottie source.
-         -Interface    Specifies the name of the interface to implement in the generated
-                       code. Defaults to Microsoft.UI.Xaml.Controls.IAnimatedVisual.
+         -InterfaceBaseName
+                       Specifies the name of the interface to implement in the generated
+                       code. Overrides the default interface base name. The default
+                       is determined by the -WinUIVersion value.
```
derobinsMSFT, line 258: The default is currently always `Microsoft.UI.Xaml.Controls.IAnimatedVisual` regardless of the WinUI version.

simeoncran: That's only true today. Any day now there is likely to be a new IRichAnimatedVisual for the XAML AnimatedIcon element. The comment is accurate anyway - it's just always determined to be IAnimatedVisual for now.

---
In reply to: [520825970](https://github.com/windows-toolkit/Lottie-Windows/pull/382#discussion_r520825970) [](ancestors = 520825970)

LottieGen/LottieJsonFileProcessor.cs
simeoncran: Good catch. It was possible briefly during development but no longer is. I wish the compiler would complain about such things.

---
In reply to: [520823554](https://github.com/windows-toolkit/Lottie-Windows/pull/382#discussion_r520823554) [](ancestors = 520823554)

LottieGen/CommandLineOptions.cs
simeoncran: It's arbitrary. It just happens to be the current version so I'm making sure that existing users don't get broken by needing to specify a version. What comment would help here?

---
In reply to: [520820972](https://github.com/windows-toolkit/Lottie-Windows/pull/382#discussion_r520820972) [](ancestors = 520820972)

derobinsMSFT: Probably just a note that this is an arbitrary choice and/or the current WinUI version.

---
In reply to: [520894269](https://github.com/windows-toolkit/Lottie-Windows/pull/382#discussion_r520894269) [](ancestors = 520894269,520820972)

derobinsMSFT (approved):
:shipit:

PR #390, opened 2020-11-20
Add support for LayerVisual and DropShadow Composition types. Cleanup.

simeoncran:
Getting ready to actually support DropShadow - we need some Composition types added.
As a result of adding the Composition types, I noticed many small cleanup tasks and included them.


source/LottieToWinComp/PreComps.cs
```diff
@@ -66,6 +68,27 @@ static class PreComps
                 result.Children.Add(rootNode);
             }
 
+            var dropShadowEffects =
+                context.Layer.Effects.Where(eff => eff.Type == LottieData.Effect.EffectType.DropShadow).ToArray();
+
+            if (dropShadowEffects.Length > 0 && dropShadowEffects.All(eff => eff.Type == LottieData.Effect.EffectType.DropShadow))
+            {
+#if DropShadows // Drop shadows are not yet supported.
```
derobinsMSFT, line 76: >DropShadows [](start = 4, length = 11)

Nit: might as well move this to before initializing `dropShadowEffects`-- that code is also not yet needed.

source/UIData/Tools/GraphCompactor.cs
```diff
@@ -77,14 +77,25 @@ void Compact(ObjectGraph<Node> graph)
...
                         break;
+
+                    case CompositionObjectType.ShapeVisual:
+                        foreach (var child in ((IContainShapes)node.Object).Shapes)
+                        {
+                            graph[child].Parent = node.Object;
+                        }
+
+                        // ShapeVisual is also a ContainerVisual.
+                        goto case CompositionObjectType.ContainerVisual;
```
derobinsMSFT, line 94: >goto case CompositionObjectType.ContainerVisual; [](start = 24, length = 48)

Necessary? Can you just let this case "fall through"?

source/WinCompData/StepEasingFunction.cs
```diff
@@ -20,19 +20,19 @@ internal StepEasingFunction(int steps)
         }
 
         // Default is 1.
-        public int StepCount { get; set; }
+        public int? StepCount { get; set; }
```
scott-moore-ms, line 23: Curious about the nullable changes here - these have a default value, but someone might explicitly set it to null?  Or is the semantic "if set, the default is xxxx (1 in this case)"?

scott-moore-ms:
> namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData

For my own education - why does this file not require a #nullable enable?   (Or alternatively, why /do/ others require it?)

---
Refers to: source/WinCompData/StepEasingFunction.cs:5 in 680604f. [](commit_id = 680604f9982495183fa6e90512fd763cf129db9a, deletion_comment = False)

source/WinCompData/CompositionEllipseGeometry.cs
```diff
@@ -16,6 +16,7 @@ internal CompositionEllipseGeometry()
         {
         }
 
+        // Default: <0, 0>
```
scott-moore-ms, line 19: >// Default: <0, 0> [](start = 8, length = 18)

Vector2.Zero as you did elsewhere?
(Is using angle brackets on vectors a thing?  I don't think I've seen that before :))

source/UIDataCodeGen/CodeGen/Stringifier.cs
```diff
@@ -86,6 +87,17 @@ public string ColorSpace(CompositionColorSpace value)
...
 
+        public string DropShadowSourcePolicy(CompositionDropShadowSourcePolicy value)
+        {
+            const string typeName = nameof(CompositionDropShadowSourcePolicy);
+            return value switch
+            {
+                CompositionDropShadowSourcePolicy.Default => $"{typeName}{ScopeResolve}{nameof(CompositionDropShadowSourcePolicy.Default)}",
+                CompositionDropShadowSourcePolicy.InheritFromVisualContent => $"{typeName}{ScopeResolve}{nameof(CompositionDropShadowSourcePolicy.InheritFromVisualContent)}",
+                _ => throw new InvalidOperationException(),
+            };
```
scott-moore-ms, line 98: What wizardry is this?!  

Is this a shorthand equivalent to this?
        public string DropShadowSourcePolicy(CompositionDropShadowSourcePolicy value)
        {
            const string typeName = nameof(CompositionDropShadowSourcePolicy);
            switch (value)
            {
                case CompositionDropShadowSourcePolicy.Default:
                    return $"{typeName}{ScopeResolve}{nameof(CompositionDropShadowSourcePolicy.Default)}";

                case CompositionDropShadowSourcePolicy.InheritFromVisualContent:
                    return $"{typeName}{ScopeResolve}{nameof(CompositionDropShadowSourcePolicy.InheritFromVisualContent)}";

                default:
                    throw new InvalidOperationException();
            };
        }

derobinsMSFT: Yes; this is a [switch expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression), added in C# 8.0.

---
In reply to: [527875015](https://github.com/windows-toolkit/Lottie-Windows/pull/390#discussion_r527875015) [](ancestors = 527875015)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
```diff
@@ -231,13 +231,13 @@ static NodeName NameExpressionAnimation(ExpressionAnimation obj)
         static NodeName NameStepEasingFunction(StepEasingFunction obj)
         {
             // Recognize 2 common patterns: HoldThenStep and StepThenHold
-            if (obj.StepCount == 1)
+            if (!obj.StepCount.HasValue || obj.StepCount.Value == 1)
             {
-                if (obj.IsFinalStepSingleFrame && !obj.IsInitialStepSingleFrame)
+                if (obj.IsFinalStepSingleFrame == true && obj.IsInitialStepSingleFrame != true)
```
scott-moore-ms, line 236: Curious that you chose to expand these out given your overall preference towards conciseness :)
FWIW with the long names, I actually found it easier to mentally parse and track in the previous form

source/UIData/Tools/GraphCompactor.cs
simeoncran: Modern languages don't allow fall through because it's so easy to not notice that a break is missing.

---
In reply to: [527340124](https://github.com/windows-toolkit/Lottie-Windows/pull/390#discussion_r527340124) [](ancestors = 527340124)

simeoncran:
> namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData

Even though nullables are enabled in the .csproj, VS was running Intellisense with nullables disabled on UWP projects. The result was a bunch of errors showing up in the VS error list even though there were no errors building. Adding the #nullable enable to files that VS complained about fixes the problem. 
Note that we have an unusual configuration because we are building the UWP code with nullables enabled. UWP does not build with nullables by default.
The workaround is only needed on files that use nullable reference types. "int?" is a nullable value type, which is an older .NET feature that is fully supported by VS.

---
In reply to: [731309921](https://github.com/windows-toolkit/Lottie-Windows/pull/390#issuecomment-731309921) [](ancestors = 731309921)

---
Refers to: source/WinCompData/StepEasingFunction.cs:5 in 680604f. [](commit_id = 680604f9982495183fa6e90512fd763cf129db9a, deletion_comment = False)

source/WinCompData/StepEasingFunction.cs
simeoncran: Nullable allows use to differentiate between the value being set to its default value and the value not being set. Before, we'd have code in various places in the back end (Instantiator and the code generators) that would check whether the value was set to its default so that it could avoid setting the value on the WUC object. Ensuring we got that right in all places was too hard, so we switched to using nullables to indicate that the value doesn't need setting on the WUC object.
The PropertyValueOptimizer is the one place that knows about default values. It walks over the properties and sets them to null if they have their default value.

Not all of the properties are using nullables. I've been fixing them as I came across them. That's why this one changed today.

---
In reply to: [527863441](https://github.com/windows-toolkit/Lottie-Windows/pull/390#discussion_r527863441) [](ancestors = 527863441)

source/UIDataCodeGen/CodeGen/NodeNamer.cs
simeoncran: It's because IsFinalStepSingleFrame is a bool? instead of a bool.
The comparison with true is equivalent to "if (obj.IsFinalStepSingleFrame.HasValue and obj.IsFinalStepSingleFrame.Value == true)".
I agree that it's easier to read without the ==, it's just not possible with nullables.

---
In reply to: [527878735](https://github.com/windows-toolkit/Lottie-Windows/pull/390#discussion_r527878735) [](ancestors = 527878735)

scott-moore-ms (approved):
:shipit:

PR #392, opened 2020-11-23
Implement DropShadow effect.

simeoncran:
This is currently turned off, however we do everything except adding the shadow to the LayerVisual so that the code is being exercised.
I have not output any issues yet for the cases we don't handle because we don't know how long before LayerVisual is fixed, and the set of issues might need changing when LayerVisual is finally working.

### DIFF
--- a/source/LottieReader/Serialization/Animatables.cs
+++ b/source/LottieReader/Serialization/Animatables.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
             CreateAnimatableParser((in LottieJsonElement element) => element.AsDouble() ?? 0);
 
         static readonly AnimatableParser<Opacity> s_animatableOpacityParser = CreateAnimatableParser(ParseOpacity);
-        static readonly AnimatableParser<Opacity> s_animatableOpacityParserByte = CreateAnimatableParser(ParseOpacityByte);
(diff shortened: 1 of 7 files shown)


source/LottieReader/Serialization/Animatables.cs
```diff
@@ -36,6 +36,7 @@ sealed partial class LottieCompositionReader
             CreateAnimatableParser((in LottieJsonElement element) => element.AsDouble() ?? 0);
 
         static readonly AnimatableParser<Opacity> s_animatableOpacityParser = CreateAnimatableParser(ParseOpacity);
+        static readonly AnimatableParser<Opacity> s_animatableOpacityParserByte = CreateAnimatableParser(ParseOpacityByte);
```
simeoncran, line 39: Maybe this should called s_animatableOpacityByteParser

source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -308,9 +308,6 @@ static float Clamp(float value, float min, float max)
 
         static Sn.Vector2 ClampedVector2(float x, float y) => Vector2(Clamp(x, 0, 1), Clamp(y, 0, 1));
```
derobinsMSFT, line 309: >static Sn.Vector2 [](start = 8, length = 17)

A couple other methods in this class also have equivalents in `ConvertTo`.

derobinsMSFT (dismissed):
:shipit:

derobinsMSFT (approved):
:shipit:

PR #394, opened 2020-11-25
Validate marker names.

simeoncran:
Marker names don't have to be unique, but it's usually a bug if they are, so issue a warning if non-unique markers are found.
LottieGen didn't have validation turned on for some reason - now it does.
Small tweak in the DGML output: give LayerVisual its own category. This allows it to be displayed with a unique color.


eliezerpMS (approved):
:shipit:

patrickeganMSFT (approved):
:shipit:

PR #397, opened 2020-11-30
Add GuassianBlurEffect to WinCompData.

simeoncran:
This adds the effect and the code to instantiate and codegen it, but the translator does not yet use it.

### DIFF
--- a/source/UIData/Tools/Optimizer.cs
+++ b/source/UIData/Tools/Optimizer.cs
@@ -990,33 +990,58 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
                 return result;
             }
 
-            var effectBase = obj.GetEffect();
+            IEnumerable<CompositionEffectSourceParameter> sources;
+            GraphicsEffectBase newEffect;
 
+            var effectBase = obj.GetEffect();
             switch (effectBase.Type)
             {
                 case GraphicsEffectType.CompositeEffect:
-                    var compositeEffect = (CompositeEffect)effectBase;
-
-                    var newCompositeEffect = new CompositeEffect
                     {
-                        Mode = compositeEffect.Mode,
-                    };
+                        var effect = (CompositeEffect)effectBase;
 
-                    var effectFactory = _c.CreateEffectFactory(newCompositeEffect);
-                    var compositeEffectBrush = effectFactory.CreateBrush();
+                        var newCompositeEffect = new CompositeEffect
+                        {
+                            Mode = effect.Mode,
+                        };
 
-                    result = CacheAndInitializeCompositionObject(obj, compositeEffectBrush);
+                        foreach (var source in effect.Sources)
+                        {
+                            newCompositeEffect.Sources.Add(source);
+                        }
 
-                    foreach (var source in compositeEffect.Sources)
+                        newEffect = newCompositeEffect;
+                        sources = effect.Sources;
+                    }
+
+                    break;
+                case GraphicsEffectType.GaussianBlurEffect:
                     {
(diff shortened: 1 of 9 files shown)


source/UIData/Tools/Optimizer.cs
```diff
@@ -990,35 +990,60 @@ CompositionEffectBrush GetCompositionEffectBrush(CompositionEffectBrush obj)
                 return result;
             }
 
-            var effectBase = obj.GetEffect();
+            IEnumerable<CompositionEffectSourceParameter> sources;
+            GraphicsEffectBase newEffect;
 
+            var effectBase = obj.GetEffect();
             switch (effectBase.Type)
```
benwestprog, line 997: >switch (effectBase.Type) [](start = 12, length = 24)

From here down the code is very similiar to Instantiator.GetCompositionEffectBrush...naively wondering if you can merge.

simeoncran: I've wanted to combine that for forever, but I can't see how to do it without templates. Both the Optimizer and the Instantiator take a WinCompData graph and produce another graph. The Optimizer produces a copy of the WinCompData graph, while the Instantiator produces a WUC graph. While you could parameterize it to take a factory that knows how to instantiate the appropriate types of objects, there are subtle differences between the initialization needed for the 2 graphs that (I think) make this impractical, e.g. WinCompData graphs contain objects with nullable properties, while WUC graphs do not. The Instantiator is careful not to unnecessarily set properties because a property set is an expensive COM call in WUC, but in WinCompData it's cheap to copy the properties over without checking for null first.

It's a difficult problem that would welcome other thoughts!

---
In reply to: [533000623](https://github.com/windows-toolkit/Lottie-Windows/pull/397#discussion_r533000623) [](ancestors = 533000623)

source/WinCompData/WinCompData.projitems
```diff
@@ -73,6 +73,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)LinearEasingFunction.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)MetaData\PropertySetValueType.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)MetaData\UapVersionAttribute.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Mgce\GaussianBlurEffect.cs" />
```
simeoncran, line 76: Fix alphabetization

source/UIDataCodeGen/CodeGen/InstantiatorGeneratorBase.cs
```diff
@@ -9,6 +9,7 @@
 using System.Diagnostics;
 using System.Linq;
 using System.Numerics;
+using System.Runtime.Versioning;
```
simeoncran, line 12: Looks like an accident.

source/UIDataCodeGen/CodeGen/CSharp/CSharpInstantiatorGenerator.cs
```diff
@@ -851,17 +852,36 @@ protected override void WriteCanvasGeometryTransformedGeometryFactory(CodeBuilde
...
             }
 
-            return compositeEffectString;
+            return effectVariable;
+        }
+
+        /// <inheritdoc/>
+        protected override string WriteGaussianBlurEffectFactory(CodeBuilder builder, GaussianBlurEffect effect)
+        {
+            var effectVariable = "guassianBlurEffect";
```
derobinsMSFT, line 871: >guassian [](start = 34, length = 8)

"gaussian"

derobinsMSFT (approved):
:shipit:

PR #399, opened 2020-12-02
Support a Gaussian blur effect field that sometimes shows up.

simeoncran:
And polish the yaml serializer a bit.

### DIFF
--- a/source/LottieReader/Serialization/Effects.cs
+++ b/source/LottieReader/Serialization/Effects.cs
@@ -103,9 +103,9 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Serialization
                         break;
 
                     case 3:
-                        // Optional parameter. Parse it is an animatable because it has the
+                        // Optional parameter. Parse it as an animatable because it has the
                         // animatable format, however it would be very strange if the value
-                        // was actually animated so convert it to non-animated bool;
+                        // was actually animated so convert it to non-animated bool.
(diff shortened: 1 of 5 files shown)


source/YamlData/YamlScalar.cs
```diff
@@ -28,7 +29,7 @@ internal YamlScalar(object value, string presentation)
             var escapedValue = value;
             if (escapedValue is null)
```
simeoncran, line 30: value is not nullable. Why is escapedValue nullable?

source/LottieReader/Serialization/Effects.cs
```diff
@@ -101,12 +102,19 @@ GaussianBlurEffect ReadGaussianBlurEffect(in LottieJsonObjectElement obj, string
                         repeatEdgePixels = ReadAnimatableBool(value) ?? throw ParseFailure();
                         break;
 
+                    case 3:
+                        // Optional parameter. Parse it is an animatable because it has the
```
derobinsMSFT, line 106: >is [](start = 56, length = 2)

"as"

source/LottieReader/Serialization/Effects.cs
```diff
@@ -101,12 +102,19 @@ GaussianBlurEffect ReadGaussianBlurEffect(in LottieJsonObjectElement obj, string
...
+                        // animatable format, however it would be very strange if the value
+                        // was actually animated so convert it to non-animated bool;
```
derobinsMSFT, line 108: >; [](start = 83, length = 1)

Nit: period.

derobinsMSFT (approved):
:shipit:

PR #403, opened 2020-12-04
Eliminate the default value checks from ConvertTo class.

simeoncran:
The checking for default values should all be in the PropertyValueOptimizer.
Also, fix some bugs in the CX output.


eliezerpMS (approved):
:shipit:

derobinsMSFT (approved):
:shipit:

PR #404, opened 2020-12-05
 Add: version checking for Blur effect, path information for error message.

simeoncran:
Blur effects will only work on UAP version 8 and above. Previously if you tried to run LottieGen to generate code that could run on version 7 and the Lottie file used blur, we'd throw an exception. Now we report the issue and turn off the blur.

Designers have asked for more context for the LT0035 error message, so I've added information about the layer and ancestor precomp layers. This information could be added to other issue messages in future.


source/LottieToWinComp/PreComps.cs
```diff
@@ -211,13 +211,21 @@ static LayerVisual ApplyDropShadow(PreCompLayerContext context, Visual visual, D
...
+            ContainerVisual source,
             GaussianBlurEffect gaussianBlurEffect)
         {
             Debug.Assert(gaussianBlurEffect.IsEnabled, "Precondition");
 
+            var factory = context.ObjectFactory;
+
+            if (!factory.IsUapApiAvailable(nameof(CompositionVisualSurface), versionDependentFeatureDescription: "Gaussian blur"))
+            {
+                // The effect can't be displayed on the targeted version.
```
scott-moore-ms, line 225: >// The effect can't be displayed on the targeted version. [](start = 16, length = 57)

Should this report some sort of warning?

scott-moore-ms (approved):
:shipit:

derobinsMSFT (approved):
:shipit:

source/LottieToWinComp/PreComps.cs
simeoncran: It does. IsUapApiAvailable complains.

---
In reply to: [537666412](https://github.com/windows-toolkit/Lottie-Windows/pull/404#discussion_r537666412) [](ancestors = 537666412)

PR #411, opened 2021-01-17
Expose animation metadata in generated code

aschultz:
Adds getters for frame count, duration, and frame rate.


simeoncran (approved):
Love it!

PR #412, opened 2021-01-23
Add WinUI 2.6 animated icon support.

simeoncran:
Add support for WinUI 2.6 animated icon. Add a IAnimatedVisualSource2 interface.

    This also removes the ability to set a custom interface in place of the WinUI
    interface, because nobody seems to need it now, and it adds a bunch of
    complexity.


eliezerpMS (approved):
:shipit:

PR #413, opened 2021-01-23
Create a project just for the animatables.

simeoncran:
The animatables are types that are used by both LottieData and the upcoming
intermediate representation, so we need them to be in a separate project so
the code can be shared.

This also adds some small features to some animatables for use by IR, and
changes some IAnimatableVector3 types to IAnimatableVector2 (e.g. gradient
stops should never by vector3.

I've also removed the "property index" from animatables. That is a Lottie concept
that is only needed if you are supporting javascript expressions. We will never
support javascript expressions, so they're not needed. Removing them makes the
code a bit cleaner.


source/Animatables/AnimatableVector2.cs
```diff
@@ -0,0 +1,51 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.Animatables
+{
+    /// <summary>
+    /// An animatable Vector2 value expressed as a single animatable Vector3 value.
```
eliezerpMS, line 12: >Vector3 [](start = 69, length = 7)

Should this be "Vector 2"?

source/Animatables/AnimatableXY.cs
```diff
@@ -0,0 +1,64 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.Animatables
+{
+    /// <summary>
+    /// An animatable Vector3 value expressed as 3 animatable floating point values.
```
eliezerpMS, line 10: >3 [](start = 49, length = 1)

Should this be "2"?

eliezerpMS (approved):
:shipit:

eliezerpMS (approved):
:shipit:

PR #415, opened 2021-01-27
Adjust the set of things that are built for different configurations.

simeoncran:
We were unnecessarily building the DotLottie and Animatables project for the non-AnyCPU configurations.
This should speed up the build a little.


eliezerpMS (approved):
:shipit:

PR #416, opened 2021-01-27
Fix bug in the Select method for non-animatable animatables.

simeoncran:
Also rationalized the select methods a bit to make them simpler.


eliezerpMS (approved):
:shipit:

PR #419, opened 2021-02-10
Turn off multi-core builds in attempt to diagnose failures to build in master

simeoncran:
This shouldn't be necessary, but master is on the floor right now and I need a PR to trigger another build.


simeoncran:
Seems to be an infrastructure issue.

PR #420, opened 2021-02-10
Pin a version of Cake to work around Cake change that broke the build

simeoncran:
An update to Cake broke the build. For now, pin the version.


eliezerpMS (approved):
:shipit:

PR #421, opened 2021-02-10
Show another decimal place on the LottieViewer tooltip.

simeoncran:
Previously, you'd press the arrow keys 10 times before the tooltip value would change. That was confusing! Now the tooltip will always change in response to the arrow keys.


eliezerpMS (approved):
:shipit:

PR #423, opened 2021-02-12
Make progress values for markers slightly after their exact time

simeoncran:
Marker values refer to the start of a frame of interest. Due to floating point imprecision we could end up with a marker that refers  to a progress value that is effectively just before the frame of interest. This can result in showing content from a previous frame.
This change attempts to fix it by nudging the marker progress values by 1/100th of a frame. This should guarantee that the marker is pointing to a time slightly after the start of the frame, even after the error caused by floating point imprecision.


raymond-wh-leung (approved):
:shipit:

simeoncran:
Combining this with https://github.com/windows-toolkit/Lottie-Windows/pull/424 because it is all for the same issue.

PR #424, opened 2021-02-12
Fixes for AnimatedIcon

simeoncran:
2 fixes for issues that became apparent as we started using markers more for AnimatedIcon.

1. Marker values refer to the start of a frame of interest. Due to floating point imprecision we could end up with a marker that refers to a progress value that is effectively just before the frame of interest. This can result in showing content from a previous frame.
This change attempts to fix it by nudging the marker progress values by 1/100th of a frame. This should guarantee that the marker is pointing to a time slightly after the start of the frame, even after the error caused by floating point imprecision.
2. There was a bug in the visibility optimizer that would cause some parts of the animation to remain visible when they should no longer be.

### DIFF
--- a/source/UIData/Tools/VisibilityAtProgress.cs
+++ b/source/UIData/Tools/VisibilityAtProgress.cs
@@ -1,0 +1,86 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
+{
+    readonly struct VisibilityAtProgress
+    {
+        internal VisibilityAtProgress(bool isVisible, float progress)
+            => (IsVisible, Progress) = (isVisible, progress);
+
+        internal bool IsVisible { get; }
+
+        internal float Progress { get; }
+
+        internal void Deconstruct(out bool isVisible, out float progress)
+        {
+            isVisible = IsVisible;
+            progress = Progress;
+        }
+
+        // Composes 2 list sequences of VisibilityAtProgress by ANDing them together.
+        // The resulting sequence is initially invisible, then visible when both a and
+        // b are visible, and invisible if either a or b are invisible.
+        internal static IEnumerable<VisibilityAtProgress> Compose(
+            VisibilityAtProgress[] a,
+            VisibilityAtProgress[] b)
+        {
+            // Get the visibilities in order, with any redundant visibilities removed.
+            var items = SanitizeSequence(a).Concat(SanitizeSequence(b)).OrderBy(v => v.Progress).ThenBy(v => !v.IsVisible).ToArray();
+
+            // The output is visible any time both a and b are visible at the same time.
+            var visibilityCounter = 0;
+
+            var initialInvisibilityOutput = false;
+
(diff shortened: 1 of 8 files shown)


source/UIData/Tools/VisibilityDescription.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
+{
+    readonly struct VisibilityDescription
+    {
+        internal VisibilityDescription(TimeSpan duration, VisibilityAtProgress[] sequence)
+            => (Duration, Sequence) = (duration, sequence);
+
+        /// <summary>
+        /// The time over which the <see cref="VisibilityDescription"/> is valide.
```
simeoncran, line 18: typo

source/UIData/Tools/VisibilityAtProgress.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        internal float Progress { get; }
+
+        internal void Deconstruct(out bool isVisible, out float progress)
+        {
+            isVisible = IsVisible;
+            progress = Progress;
+        }
+
+        // Composes 2 list sequences of VisibilityAtProgress by ANDing them together.
```
simeoncran, line 27: >list [](start = 22, length = 4)

typo

source/UIData/Tools/VisibilityAtProgress.cs
```diff
@@ -0,0 +1,86 @@
...
+        // The resulting sequence is initially invisible, then visible when both a and
```
simeoncran, line 28: >is initially invisible [](start = 34, length = 22)

Misleading. It is made explicitly invisible.

source/UIData/Tools/VisibilityAtProgress.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            var initialInvisibilityOutput = false;
+
+            foreach (var item in items)
+            {
+                visibilityCounter += item.IsVisible ? 1 : -1;
+                if (visibilityCounter == 2)
+                {
+                    // Both a and b are now visible.
+                    if (!initialInvisibilityOutput)
```
simeoncran, line 48: Use a less ambiguous name.

source/UIData/Tools/VisibilityAtProgress.cs
```diff
@@ -0,0 +1,86 @@
...
+                    {
+                        initialInvisibilityOutput = true;
+                        if (item.Progress != 0)
```
simeoncran, line 51: Add a comment.

source/UIData/Tools/VisibilityAtProgress.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        // Orders the items in the sequence by Progress, and removes any redundant items.
+        static IEnumerable<VisibilityAtProgress> SanitizeSequence(IEnumerable<VisibilityAtProgress> sequence)
+        {
+            // Sequences start implicitly invisible.
+            var previousVisibility = false;
+
+            foreach (var item in sequence.OrderBy(v => v.Progress).ThenBy(v => !v.IsVisible))
+            {
+                // Ignore any repeats of the same visibility state.
```
simeoncran, line 75: For completeness, consider how to sanitize a visible and invisible at the same progress. Right now we output the visible followed by the invisible, but probably we should treat it the same as being invisible.

eliezerpMS (approved):
:shipit:

PR #425, opened 2021-02-22
Show frame numbers in generated code.

simeoncran:
This makes it much easier to figure out what is causing issues in animations.
Also remove the "readonly" on properties in the generated C# because they aren't supported by the language versions used in UWP.


peterjtan (approved):
:shipit:

PR #426, opened 2021-02-23
LottieViewer UI improvements for AnimatedIcon

simeoncran:
These are all changes that I realized had value as people started using LottieViewer for AnimatedIcon.

1. The nudge factors have been adjusted up to 0.05. This is the value used to compensate for loss of floating point precision that can cause a progress value to refer to the end of a previous frame, resulting in flashes. The value was previously 0.01 and there still seemed to be some occasional issues. 0.05 seems safe, so using that everywhere now.
2. Removed all references to "Progress" in the LottieViewer UI. Seeing as LottieViewer is for designers, progress makes no sense as they are dealing in frames. So now the marker hyperlinks are in frames, and the tooltip on the scrubber only shows frames. Developers that need to convert to progress can the the math themselves (it's just frame_count/number_of_frames plus a nudge.
3. Make the arrow keys on the scrubber move the scrubber by a single frame.
4. Give the scrubber focus more easily. Clicking most places on the page will give the scrubber focus, so the scrubber key presses should work more intuitively.
5.  Fix the colors on hyperlinks and the playback rate slider so that the color doesn't change with the system theme. Previously the accent color would cause the slider and the hyperlinks to change color, which could cause problems if the color was not compatible with the gray background of LottieViewer.


LottieViewer/FloatFormatter.cs
```diff
@@ -0,0 +1,25 @@
...
+#nullable enable
+
+using System;
+using Windows.UI.Xaml.Data;
+
+namespace LottieViewer
+{
+    public sealed class FloatFormatter : IValueConverter
+    {
+        object IValueConverter.Convert(object value, Type targetType, object parameter, string language)
```
peterjtan, line 14: Just checking my knowledge here. Is this implementing the interface method? What's the difference between this and `object Convert(...)`? #Resolved

peterjtan (approved):

LottieViewer/FloatFormatter.cs
simeoncran: This is an "explicit interface implementation". It allows the class to implement the interface without the methods being public methods on the class. In general it's good to have as little public surface area on a class as possible as it helps encapsulation. The other reason for using explicit implementations is when you have interface methods that conflict with other methods on the class. If they're explicitly implemented then they won't conflict. A common example is the implementation of IEnumerable.GetEnumerator and IEnumerable<T>.GetEnumerator.

---
In reply to: [581266511](https://github.com/windows-toolkit/Lottie-Windows/pull/426#discussion_r581266511) [](ancestors = 581266511)

source/UIDataCodeGen/CodeGen/MarkerInfo.cs
```diff
@@ -64,28 +62,7 @@ internal static IEnumerable<MarkerInfo> GetMarkerInfos(IEnumerable<Marker> marke
...
-            {
-                var adjustedFrame = marker.Frame;
-
-                if (adjustedFrame.Number > 0)
-                {
-                    adjustedFrame += new Duration(0.01, marker.Duration);
-                }
-
-                yield return new Marker(marker.Name, adjustedFrame, marker.Duration);
+                yield return new MarkerInfo(m, name, startConstant, endConstant, nudgeFrameProportion: 0.05);
```
simeoncran, line 65: I think this should be parameter rather than hard coded here.

source/UIData/Tools/Optimizer.cs
```diff
@@ -38,20 +38,19 @@ public static Visual Optimize(Visual root, bool ignoreCommentProperties)
...
-
-            // The graph compactor may create new CompositionObjects (e.g. BooleanKeyFrameAnimations).
-            // Re-run the canonicalizer so that the new objects are canonicalized.
-            result = CanonicalizeGraph(result, ignoreCommentProperties);
+            //// Re-run the property value optimizer because the compactor may have set new properties.
+            //result = PropertyValueOptimizer.OptimizePropertyValues(result);
 
+            //// The graph compactor may create new CompositionObjects (e.g. BooleanKeyFrameAnimations).
+            //// Re-run the canonicalizer so that the new objects are canonicalized.
+            //result = CanonicalizeGraph(result, ignoreCommentProperties);
```
simeoncran, line 53: The optimizer has been accidentally disabled!!! Fix.

source/LottieMetadata/LottieCompositionMetadata.cs
```diff
@@ -28,7 +28,7 @@ sealed class LottieCompositionMetadata
             Markers = markers.Select(
                 m => new Marker(
                     m.name,
-                    new Frame(Duration, m.frame - inPoint),
+                    Duration.GetFrame(m.frame - inPoint),
```
simeoncran, line 31: >GetFrame [](start = 29, length = 8)

Naming is confusing. Is this really "GetFrameByFrameNumber" ?

source/LottieMetadata/Duration.cs
```diff
@@ -5,6 +5,7 @@
 #nullable enable
 
 using System;
+using System.Reflection.Metadata.Ecma335;
```
simeoncran, line 8: Accidental. Remove!

source/LottieMetadata/Duration.cs
```diff
@@ -39,6 +40,38 @@ public Duration(double frames, Duration other)
 
         public double FPS { get; }
 
+        /// <summary>
+        /// Gets the <see cref="Frame"/> with the given <paramref name="frameNumber"/>.
+        /// </summary>
+        /// <param name="frameNumber">The frame number.</param>
+        /// <returns>The frame in this <see cref="Duration"/>.</returns>
+        public Frame GetFrame(double frameNumber)
```
simeoncran, line 48: Naming isn't clear. GetFrameFromFrameNumber

LottieViewer/VisibilityConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                boolValue = count > 0;
+            }
+            else if (value is ICollection collection)
+            {
+                boolValue = collection.Count > 0;
+            }
+            else
+            {
+                // Used !null to convert to a boolean.
```
simeoncran, line 33: >Used [](start = 19, length = 4)

Typo.

LottieViewer/Scrubber.xaml.cs
```diff
@@ -216,6 +221,77 @@ protected override Size ArrangeOverride(Size finalSize)
...
+                }
+                else
+                {
+                    Value = _diagnostics.FrameToNudgedProgress(previousFrame);
+                }
+            }
+        }
+
+        // Called when a keypress indicates that the scrubber should move its position to the right.
+        void OnScrubberRightKey(VirtualKeyModifiers modifiers)
```
simeoncran, line 257: >OnScrubberRightKey [](start = 13, length = 18)

Can we combine the right and left impl? Common code just needs to take a +1 or -1.

LottieViewer/MainPage.xaml.cs
```diff
@@ -21,10 +21,10 @@
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Controls.Primitives;
-using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Documents;
```
simeoncran, line 24: These new usings - are they really used?

LottieViewer/ViewModel/LottieVisualDiagnosticsViewModel.cs
```diff
@@ -151,6 +174,18 @@ public string SizeText
             }
         }
 
+        // Get the frame with the given number.
+        public Frame GetFrame(double frameNumber)
```
simeoncran, line 178: GetFrameFromNumber

LottieViewer/ViewModel/LottieVisualDiagnosticsViewModel.cs
```diff
@@ -151,6 +174,18 @@ public string SizeText
...
+            => LottieVisualDiagnostics is null ? default : new Frame(Duration, frameNumber);
+
+        // Returns the nudged progress that is equivalent to the given frame number.
+        public double FrameToNudgedProgress(double frameNumber)
+            => LottieVisualDiagnostics is null ? default : GetFrame(frameNumber).GetNudgedProgress(NudgeFrameProportion);
+
+        // Converts the given nudged progress value to a frame number.
+        public Frame NudgedProgressToFrame(double progress)
```
simeoncran, line 186: GetFrameFromNumdgedProgress

LottieViewer/ViewModel/LottieVisualDiagnosticsViewModel.cs
```diff
@@ -151,6 +174,18 @@ public string SizeText
             }
         }
 
+        // Get the frame with the given number.
+        public Frame GetFrame(double frameNumber)
+            => LottieVisualDiagnostics is null ? default : new Frame(Duration, frameNumber);
+
+        // Returns the nudged progress that is equivalent to the given frame number.
+        public double FrameToNudgedProgress(double frameNumber)
```
simeoncran, line 182: GetNudgedProgressFromFrameNumber

peterjtan (approved):
:shipit:

peterjtan (approved):
:shipit:

PR #427, opened 2021-02-23
Convert to modern null checking now that it works in .NET Native.

simeoncran:
This is mostly a find-and-replace of the == null and != null checks.
This makes the code consistent, modern, and safe from equality overloading errors.
I have not replaced them in the generated code.
I have made the C# generator output the "is not null" pattern.


raymond-wh-leung (approved):
:shipit:

Issue #406, opened 2021-01-05
AdditionalInterface param can add bad IDL import statements


aschultz:
When using the AdditionalInterfaces parameter to add an interface to the generated IDL files, it also causes an import statement to be added. For example:

```
// LottieGen -AdditionalInterface LottieReactNative.ILottieCodegenSource -L cppwinrt ...

import "LottieReactNative.ILottieCodegenSource.idl";

namespace AnimatedVisuals
{
    runtimeclass Watermelon
        : [default] Microsoft.UI.Xaml.Controls.IAnimatedVisualSource
        , LottieReactNative.ILottieCodegenSource
    {
        Watermelon();
    };
}
```

This causes problems if the interface is coming from a separate library (in this example, a C++/WinRT project named LottieReactNative that I've added as a project reference) because there is no IDL file for the current project to find and import.

Simply removing the import allows the build to proceed correctly. I'm not sure if the imports are actually necessary in the case where the referenced interface *is* part of the current project (I haven't seen the need to do this yet in my uses of C++/WinRT), but if it's required in some scenarios I suggest making this a flag for the tool so we can turn it off.

I'm not an expert on MIDL, so let me know if there's a different way I should be setting this up.

simeoncran:
Hey @aschultz , I'm not an expert on C++/WinRT, but if we put our 2 heads together I'm sure we can work this out. :-)
How does MIDL know what "LottieReactNative.ILottieCodegenSource" is in your case? Is it coming from a .winmd file from the other library? Can you see what arguments are being passed to the midl compiler?

aschultz:
The LottieReactNative library project that contains `ILottieCodegenSource` is compiled into DLL + winmd. The consuming app has a ProjectReference to the library project, which seems to be enough for CppWinRT to find and make use of the binaries.

Since the LottieGen-generated code is in the consuming app, and it's referencing an interface in the library, I think that obviates the need for the `import <>.idl` statement.

From my limited experience, imports only seem to be needed when you have IDL files referencing each other within the same project, not when you're referencing types from a separate library.

PR #433, opened 2021-03-08
Fix for #406

simeoncran:
When using the "-AdditionalInterface" parameter in LottieGen (which causes the generated code to declare that it implements interfaces that it otherwise doesn't know about), the interface needs to be known to the generated IDL file.
There are 2 ways that the IDL may know about the interface:
1. The interface may exist in a .winmd file that is referenced by the project.
2. The interface may be defined in the project using an IDL file.

LottieGen doesn't know which case the user requires. Previously we were assuming case 2 which causes problems if the interface is not in an IDL file.
Rather than requiring the user to tell LottieGen which case is being used, we know wrap the import in a condition that depends on the existence of the file. This should handle both cases seamlessly.


raymond-wh-leung (approved):
:shipit:

PR #434, opened 2021-03-30
Fix for security issue in System.Text.Encodings.Web v4.7.1.

simeoncran:
Error reported:
Microsoft.Toolkit.Uwp.UI.Lottie v7 uses System.Text.Encodings.Web v4.7.1 which has a high severity security vulnerability and as such doesn't pass component governance

This upgrades all of the runtime NuGet packages to the latest version.


scott-moore-ms (approved):
Don't know why I don't have the CodeFlow widget :-/

eliezerpMS (approved):
:shipit:

PR #437, opened 2021-04-05
Update Lottie to use 5.0.1 of System.Text.Encodings.Web

eliezerpMS:
Certain versions of System.Text.Encodings.Web are flagged as having possible security vulnerabilities. Version 5.0.1 is the newest version that is marked as safe.


raymond-wh-leung (approved):
:shipit:

patrickeganMSFT (approved):
:shipit:

Issue #443, opened 2021-04-22
LottieGen.MSBuild props file has invalid path to DLL


aschultz:
Directory structure of the nuget package looks like:

- bin\
  - LottieGen.MsBuild.dll
- build\
  - native\
    - LottieGen.MsBuild.props
  - win\
    - LottieGen.MsBuild.props

LottieGen.MsBuild.props:
```
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <UsingTask
    TaskName="LottieGen.Task.LottieGen"
    AssemblyFile="$(MsBuildThisFileDirectory)..\bin\LottieGen.MsBuild.dll"/>
</Project>
```

The path should read `$(MsBuildThisFileDirectory)..\..\bin\LottieGen.MsBuild.dll`

PR #444, opened 2021-04-23
Fixes for MSBuild task

aschultz:
- Parameter validation no longer fails on valid params
- Fixed props file path to LottieGen dll (#443)
- Fixed path generation for LottieGen exe
- Converted parameter types which are not supported by MSBuild to acceptable types
- Added escaping to command-line args to avoid mangling file paths


eliezerpMS (approved):
:shipit:

PR #448, opened 2021-06-07
Fix issue with nullable

eliezerpMS:
An issue was found with a recent PR build where we were returning "((Mgce.GaussianBlurEffect)effect).Source" for list that was not of nullable types.


eliezerpMS:
This PR was just to test out a quick change that was not reproing locally. Someone else is now on point to investigate this so I am closing this PR.

PR #450, opened 2021-06-10
Gradient fill for rectangle(rounded rectangle) shape fixed

aborziak-ms:
Because of the difference in how gradient fill for rectangles works in **WinComp API** and Lottie, previously it was displayed incorrectly.

**WinComp API** uses top left corner of rectangle as point [0, 0] while Lottie uses middle-middle instead. To account for this difference we need to add **(Rectangle.Size/2)** to Start and End points of gradients. We can't add a constant value since **Rectangle.Size** can also be animated, and if it is animated, we need to use expression animation.

In this change, I added InternalOffset class for rectangle shapes, which propagates information about offset to the gradients and then this offset is applied to Start and End points.

Examples below show how it works now:

### DIFF
--- a/source/LottieToWinComp/Brushes.cs
+++ b/source/LottieToWinComp/Brushes.cs
@@ -168,7 +168,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             ApplyCommonStrokeProperties(
                 context,
                 shapeStroke,
-                TranslateLinearGradient(context, shapeStroke, contextOpacity),
+                TranslateLinearGradient(context, shapeStroke, contextOpacity, null),
                 sprite);
         }
 
@@ -181,7 +181,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             ApplyCommonStrokeProperties(
                 context,
                 shapeStroke,
-                TranslateRadialGradient(context, shapeStroke, contextOpacity),
+                TranslateRadialGradient(context, shapeStroke, contextOpacity, null),
                 sprite);
         }
 
@@ -254,7 +254,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             sprite.StrokeBrush = brush;
         }
 
-        public static CompositionBrush? TranslateShapeFill(LayerContext context, ShapeFill? shapeFill, CompositeOpacity opacity)
+        public static CompositionBrush? TranslateShapeFill(
+            LayerContext context,
+            ShapeFill? shapeFill,
+            CompositeOpacity opacity,
+            Rectangles.InternalOffset? internalOffset)
         {
             if (shapeFill is null)
             {
@@ -264,8 +268,8 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             return shapeFill.FillKind switch
             {
                 ShapeFill.ShapeFillKind.SolidColor => TranslateSolidColorFill(context, (SolidColorFill)shapeFill, opacity),
-                ShapeFill.ShapeFillKind.LinearGradient => TranslateLinearGradient(context, (LinearGradientFill)shapeFill, opacity),
-                ShapeFill.ShapeFillKind.RadialGradient => TranslateRadialGradient(context, (RadialGradientFill)shapeFill, opacity),
+                ShapeFill.ShapeFillKind.LinearGradient => TranslateLinearGradient(context, (LinearGradientFill)shapeFill, opacity, internalOffset),
+                ShapeFill.ShapeFillKind.RadialGradient => TranslateRadialGradient(context, (RadialGradientFill)shapeFill, opacity, internalOffset),
                 _ => throw new InvalidOperationException(),
             };
         }
@@ -406,10 +410,40 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             return result;
         }
 
+        static void TranslateVector2AnimatableWithInternalOffset(
+            LayerContext context,
+            CompositionObject obj,
+            string propertyName,
+            TrimmedAnimatable<Vector2> value,
+            Rectangles.InternalOffset offset)
+        {
+            // anomate source property first
+            string sourcePropertyName = propertyName + "Source";
+            obj.Properties.InsertVector2(sourcePropertyName, ConvertTo.Vector2(value.InitialValue));
+            Animate.Vector2(context, value, obj, sourcePropertyName);
+
+            // create expression that offsets source property by internal offset
+            WinCompData.Expressions.Vector2 expression = offset.IsAnimated ?
+                ExpressionFactory.InternalOffsetExressionAdded(sourcePropertyName, offset.OffsetExpression!) :
+                ExpressionFactory.InternalOffsetValueAdded(sourcePropertyName, (Sn.Vector2)offset.OffsetValue!);
+
+            var expressionAnimation = context.ObjectFactory.CreateExpressionAnimation(expression);
+            expressionAnimation.SetReferenceParameter("my", obj);
+            if (offset.IsAnimated)
+            {
+                // expression can use geometry
+                expressionAnimation.SetReferenceParameter("geometry", offset.Geometry);
+            }
+
+            // animate original property with expression that applies  internal offset to it
+            Animate.WithExpression(obj, expressionAnimation, propertyName);
+        }
+
         static CompositionLinearGradientBrush? TranslateLinearGradient(
             LayerContext context,
             IGradient linearGradient,
-            CompositeOpacity opacity)
+            CompositeOpacity opacity,
+            Rectangles.InternalOffset? internalOffset)
         {
             var result = context.ObjectFactory.CreateLinearGradientBrush();
 
@@ -419,7 +453,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             var startPoint = Optimizer.TrimAnimatable(context, linearGradient.StartPoint);
             var endPoint = Optimizer.TrimAnimatable(context, linearGradient.EndPoint);
 
-            if (startPoint.IsAnimated)
+            if (internalOffset is not null)
+            {
+                TranslateVector2AnimatableWithInternalOffset(context, result, nameof(result.StartPoint), startPoint, internalOffset);
+            }
+            else if (startPoint.IsAnimated)
             {
                 Animate.Vector2(context, startPoint, result, nameof(result.StartPoint));
             }
@@ -428,7 +466,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 result.StartPoint = ConvertTo.Vector2(startPoint.InitialValue);
             }
 
-            if (endPoint.IsAnimated)
+            if (internalOffset is not null)
+            {
+                TranslateVector2AnimatableWithInternalOffset(context, result, nameof(result.EndPoint), endPoint, internalOffset);
+            }
+            else if (endPoint.IsAnimated)
             {
                 Animate.Vector2(context, endPoint, result, nameof(result.EndPoint));
             }
@@ -453,13 +495,14 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
         static CompositionGradientBrush? TranslateRadialGradient(
             LayerContext context,
             IRadialGradient gradient,
-            CompositeOpacity opacity)
+            CompositeOpacity opacity,
+            Rectangles.InternalOffset? internalOffset)
         {
             if (!context.ObjectFactory.IsUapApiAvailable(nameof(CompositionRadialGradientBrush), versionDependentFeatureDescription: "Radial gradient fill"))
             {
                 // CompositionRadialGradientBrush didn't exist until UAP v8. If the target OS doesn't support
                 // UAP v8 then fall back to linear gradients as a compromise.
-                return TranslateLinearGradient(context, gradient, opacity);
+                return TranslateLinearGradient(context, gradient, opacity, internalOffset);
             }
 
             var result = context.ObjectFactory.CreateRadialGradientBrush();
@@ -470,7 +513,11 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             var startPoint = Optimizer.TrimAnimatable(context, gradient.StartPoint);
             var endPoint = Optimizer.TrimAnimatable(context, gradient.EndPoint);
 
-            if (startPoint.IsAnimated)
+            if (internalOffset is not null)
+            {
+                TranslateVector2AnimatableWithInternalOffset(context, result, nameof(result.EllipseCenter), startPoint, internalOffset);
+            }
+            else if (startPoint.IsAnimated)
             {
                 Animate.Vector2(context, startPoint, result, nameof(result.EllipseCenter));
             }
--- a/source/LottieToWinComp/Brushes.cs
+++ b/source/LottieToWinComp/Brushes.cs
@@ -168,7 +168,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             ApplyCommonStrokeProperties(
                 context,
                 shapeStroke,
-                TranslateLinearGradient(context, shapeStroke, contextOpacity, null),
+                TranslateLinearGradient(context, shapeStroke, contextOpacity),
                 sprite);
         }
 
@@ -181,7 +181,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             ApplyCommonStrokeProperties(
                 context,
                 shapeStroke,
-                TranslateRadialGradient(context, shapeStroke, contextOpacity, null),
+                TranslateRadialGradient(context, shapeStroke, contextOpacity),
                 sprite);
         }
 
@@ -258,7 +258,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             LayerContext context,
             ShapeFill? shapeFill,
             CompositeOpacity opacity,
-            Rectangles.InternalOffset? internalOffset)
+            Rectangles.OriginOffset? originOffset)
         {
             if (shapeFill is null)
             {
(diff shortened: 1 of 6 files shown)


source/LottieToWinComp/Brushes.cs
```diff
@@ -406,10 +410,40 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
...
         }
 
+        static void TranslateVector2AnimatableWithInternalOffset(
+            LayerContext context,
+            CompositionObject obj,
+            string propertyName,
+            TrimmedAnimatable<Vector2> value,
+            Rectangles.InternalOffset offset)
+        {
+            // anomate source property first
```
eliezerpMS, line 420: >anomate [](http://example.com/codeflow?start=15&length=7)

Annotate?

source/LottieToWinComp/Brushes.cs
```diff
@@ -406,10 +410,40 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
...
+            string sourcePropertyName = propertyName + "Source";
+            obj.Properties.InsertVector2(sourcePropertyName, ConvertTo.Vector2(value.InitialValue));
+            Animate.Vector2(context, value, obj, sourcePropertyName);
+
+            // create expression that offsets source property by internal offset
```
eliezerpMS, line 425: >/ create expression that offsets source property by internal offset [](http://example.com/codeflow?start=13&length=67)

I would scub the comments to make them follow the format that the rest of the code has, which is to have then start with a capital letter and a period at the end (where appropriate)

source/LottieToWinComp/Brushes.cs
aborziak-ms: oh, it should be *an**i**mate* :)

eliezerpMS: Ah, that makes sense

source/LottieToWinComp/Rectangles.cs
```diff
@@ -18,6 +18,41 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
+        // top left corner and in Lottie they start in the middle
+        // To account for this we need to offset all the internal points
+        // for (Rectangle.Size / 2)
+        // This class represents this offset (static or animated)
+        public class InternalOffset
+        {
+            public RectangleOrRoundedRectangleGeometry Geometry { get; }
+
+            // Use expression if size is animated
+#nullable enable
```
eliezerpMS, line 32: Just curious, what was the reason for having this just scoped to here?

aborziak-ms: Ah, yes, it is interesting for me too. Without this Visual Studio complained about `public Expressions.Vector2?` and didn't want to build. I don't know what is special about Expressions.Vector2

source/LottieToWinComp/Brushes.cs
```diff
@@ -419,7 +453,11 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
             var startPoint = Optimizer.TrimAnimatable(context, linearGradient.StartPoint);
             var endPoint = Optimizer.TrimAnimatable(context, linearGradient.EndPoint);
 
-            if (startPoint.IsAnimated)
+            if (internalOffset is not null)
+            {
+                TranslateVector2AnimatableWithInternalOffset(context, result, nameof(result.StartPoint), startPoint, internalOffset);
```
aborziak-ms, line 458: I also need to optimize this case, if offset and start/end point are not animated we can use a constant instead of expression

source/LottieToWinComp/Brushes.cs
```diff
@@ -428,7 +466,11 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
                 result.StartPoint = ConvertTo.Vector2(startPoint.InitialValue);
             }
```
eliezerpMS, line 467: This segment is starting to get a bit complicated. I also think for readability how about making this all just 1 function? I think that the two if statements here can be combined (as in we just have the first case with no offset). Then we can call this helper method for both start and end point.

source/LottieToWinComp/Brushes.cs
```diff
@@ -264,8 +268,8 @@ static CompositionColorBrush CreateAnimatedColorBrush(LayerContext context, in T
             return shapeFill.FillKind switch
             {
                 ShapeFill.ShapeFillKind.SolidColor => TranslateSolidColorFill(context, (SolidColorFill)shapeFill, opacity),
-                ShapeFill.ShapeFillKind.LinearGradient => TranslateLinearGradient(context, (LinearGradientFill)shapeFill, opacity),
-                ShapeFill.ShapeFillKind.RadialGradient => TranslateRadialGradient(context, (RadialGradientFill)shapeFill, opacity),
+                ShapeFill.ShapeFillKind.LinearGradient => TranslateLinearGradient(context, (LinearGradientFill)shapeFill, opacity, internalOffset),
+                ShapeFill.ShapeFillKind.RadialGradient => TranslateRadialGradient(context, (RadialGradientFill)shapeFill, opacity, internalOffset),
```
eliezerpMS, line 272: nit: this line is getting a bit long. Mind breaking them up into mulitple lines?

source/LottieToWinComp/Brushes.cs
aborziak-ms: Yeah, initially that was my plan too. 

Then I noticed that in first two cases we are applying animation to the property without assigning anything, and in third case we are assigning a value. And also we have an optimization so that if value is not assigned then no additional code will be generated (to assign it to default) So I tried to preserve this behavior

But yeah, I will try to find a workaround, I don't like this part too

source/LottieToWinComp/Brushes.cs
```diff
@@ -406,10 +410,40 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
...
 
+        static void TranslateVector2AnimatableWithInternalOffset(
+            LayerContext context,
+            CompositionObject obj,
+            string propertyName,
+            TrimmedAnimatable<Vector2> value,
+            Rectangles.InternalOffset offset)
+        {
+            // anomate source property first
+            string sourcePropertyName = propertyName + "Source";
```
eliezerpMS, line 421: >propertyName + "Source" [](http://example.com/codeflow?start=40&length=23)

Just curious why add "Soure" for the name?

aborziak-ms: If we have an animation for StartPoint (for example) then I'm applying this animation to StartPointSource property, and then applying expression animation to StartPoint property itself, to make it equal to StartPointSource + geometry.Size/2. 

`{Property}Source` is basically a helper-property that stores original animation, and then used inside expression

source/LottieToWinComp/Brushes.cs
```diff
@@ -406,10 +410,40 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
...
+            obj.Properties.InsertVector2(sourcePropertyName, ConvertTo.Vector2(value.InitialValue));
+            Animate.Vector2(context, value, obj, sourcePropertyName);
+
+            // create expression that offsets source property by internal offset
+            WinCompData.Expressions.Vector2 expression = offset.IsAnimated ?
```
eliezerpMS, line 426: >offset.IsAnimated [](http://example.com/codeflow?start=57&length=17)

If offset.IsAnimated is true then is offset.OffsetExpression ever null?

aborziak-ms: `IsAnimated` is basically a toggle between `OffsetExpression` and `OffsetValue`, if it **is animated** then `OffsetExpression` is not null, if it **is not animated** then `OffsetValue` is not null. It is guaranteed by having two constructors, one for expression and one for value.

source/LottieToWinComp/Rectangles.cs
```diff
@@ -318,10 +353,15 @@ public static CompositionShape TranslateRectangleContent(ShapeContext context, R
             var height = size.InitialValue.Y;
             var trimOffsetDegrees = (width / (2 * (width + height))) * 360;
 
+            InternalOffset internalOffset = size.IsAnimated ?
```
eliezerpMS, line 356: >InternalOffset internalOffset = size.IsAnimated ? [](http://example.com/codeflow?start=12&length=49)

I would put a comment here calling why you need to do this and in the other location where you create InternalOffset.

source/LottieToWinComp/Brushes.cs
eliezerpMS: I guess I am wondering if OffsetExpression and OffsetValue need to be nullables?

source/LottieToWinComp/Brushes.cs
eliezerpMS: Looking at the other usages of "Properties.Insert..." in the codebase it looks like the names are kept simple. Would it cause an issue to just leave the name as the value of "propertyName"?

source/LottieToWinComp/Rectangles.cs
```diff
@@ -18,6 +18,41 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
     /// </summary>
     static class Rectangles
     {
+        // Rectangles are implemented differently in WinComp API
+        // and Lottie. In WinComp API coordinates inside rectangle start in
+        // top left corner and in Lottie they start in the middle
+        // To account for this we need to offset all the internal points
+        // for (Rectangle.Size / 2)
+        // This class represents this offset (static or animated)
+        public class InternalOffset
```
eliezerpMS, line 27: >InternalOffset [](http://example.com/codeflow?start=21&length=14)

For naming how about something like "OriginOffset"? I guess I wondering if people might be confused by the term "internal".....

source/LottieToWinComp/Rectangles.cs
```diff
@@ -18,6 +18,41 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
...
+        {
+            public RectangleOrRoundedRectangleGeometry Geometry { get; }
+
+            // Use expression if size is animated
+#nullable enable
+            public Expressions.Vector2? OffsetExpression { get; }
+#nullable disable
+
+            // Use constant value if size if static
```
eliezerpMS, line 36: >if [](http://example.com/codeflow?start=42&length=2)

is

source/LottieToWinComp/Brushes.cs
aborziak-ms: I guess we can't do that, since propertyName is the name of reserved property, (e.g. nameof(gradient.StartPoint)) and here I'm just creating auxiliary property (with "Source" at the end) that holds initial animation, so that we can use it inside expression.

**Shortly:** We need to animate "{propertyName}" so that it is equal to **intended animation + offset**, so basically the steps are following:
1. Animate "{propertyName}Source" instead (with original animation)
2. Create expression "{propertyName}Source" + "geometry.Size/2"
3. Animate "{propertyName}" using this expression

source/LottieToWinComp/Shapes.cs
```diff
@@ -34,17 +34,18 @@ public static LayerTranslator CreateShapeLayerTranslator(ShapeLayerContext conte
             ShapeContext context,
             CompositionSpriteShape shape,
             bool reverseDirection) =>
-            TranslateAndApplyShapeContextWithTrimOffset(context, shape, reverseDirection, 0);
+            TranslateAndApplyShapeContextWithTrimOffset(context, shape, reverseDirection, null, 0);
```
eliezerpMS, line 37: >null [](http://example.com/codeflow?start=90&length=4)

Other shape types don't have this offset issue?

source/LottieToWinComp/Brushes.cs
eliezerpMS: Ok. That sounds fine.

source/LottieToWinComp/Rectangles.cs
eliezerpMS: Just curious, what was the error and which project complained?

eliezerpMS (changes requested):
:clock1:

source/LottieToWinComp/Rectangles.cs
```diff
@@ -318,10 +353,15 @@ public static CompositionShape TranslateRectangleContent(ShapeContext context, R
...
 
+            InternalOffset internalOffset = size.IsAnimated ?
+                new InternalOffset(geometry, ExpressionFactory.GeometryHalfSize) :
+                new InternalOffset(geometry, ConvertTo.Vector2(size.InitialValue / 2));
+
             Shapes.TranslateAndApplyShapeContextWithTrimOffset(
                 context,
                 compositionRectangle,
                 rectangle.DrawingDirection == DrawingDirection.Reverse,
+                internalOffset,
```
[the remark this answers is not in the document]
aborziak-ms: Hm, I just noticed that RectangleOrRoundedRectangleGeometry is not CompositionSpriteShape, it is just a helper class. 
We can move offset info to CompositionRectangleGeometry (and CompositionRoundedRectangleGeometry) but not sure if we should add some temporary fields there.

source/LottieToWinComp/Rectangles.cs
aborziak-ms: image

source/LottieToWinComp/Shapes.cs
aborziak-ms: Yes, only Rectangle and RoundedRectangle shapes have Offset value.

source/LottieToWinComp/Rectangles.cs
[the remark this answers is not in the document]
eliezerpMS: Yeah it might be a bit odd to "pollute" Composition object representations with something like that. I guess the next candidate could be to use the ShapeContext.

source/LottieToWinComp/Brushes.cs
aborziak-ms: If we make them non-nullable then we need to initialize them with some default values? I just wanted to use null-ness of them as a flag, `OffsetValue is null` => Animated, `OffsetExpresion is null` => NotAnimated

source/LottieToWinComp/Brushes.cs
```diff
@@ -419,22 +480,16 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
             var startPoint = Optimizer.TrimAnimatable(context, linearGradient.StartPoint);
             var endPoint = Optimizer.TrimAnimatable(context, linearGradient.EndPoint);
 
-            if (startPoint.IsAnimated)
-            {
-                Animate.Vector2(context, startPoint, result, nameof(result.StartPoint));
-            }
-            else
+            var startPointValue = AnimateVector2WithOriginOffsetOrGetValue(context, result, nameof(result.StartPoint), startPoint, originOffset);
```
eliezerpMS: >var startPointValue = AnimateVector2WithOriginOffsetOrGetValue(context, result, nameof(result.StartPoint), startPoint, originOffset); [](http://example.com/codeflow?start=12&length=133)

nit: make this into two line like:
var startPointValue = AnimateVector2WithOriginOffsetOrGetValue(
    context,
    result,
    nameof(result.StartPoint),
    startPoint,
    originOffset);

aborziak-ms:
Ok, so I moved OriginOffset to ShapeLayerContainer, now it looks kind of better in Brushes.cs, Shapes.cs. But now we have this weird line (Rectangles.cs:333) `context.LayerContext.OriginOffset = null;`  because we need to cleanup the offset before exiting this scope.

source/LottieToWinComp/Brushes.cs
```diff
@@ -406,10 +413,64 @@ static Color DefaultValueOf(Animatable<Color> animatableColor)
             return result;
         }
 
+        // Animate Vector2 property of object with TrimmedAnimatable while applying OriginOffset to it.
+        // Returns non-null Sn.Vector2 value if no animation is needed (and no animation was applied).
```
eliezerpMS, line 416: >Returns non-null Sn.Vector2 value if no animation is needed (and no animation was applied). [](http://example.com/codeflow?start=11&length=91)

I guess a philosophical conversation to have offline: How do you feel about the <summary> syntax for C# functions? It has the "Return" value xml where we could specify things like this. I don't remember why we don't use that in this codebase though...

source/LottieToWinComp/Brushes.cs
eliezerpMS: Thanks! I like how it look now much better!

source/LottieToWinComp/Brushes.cs
eliezerpMS: Yeah I guess I was thinking that you already know the flag value from the contructor and it does not change. so you could just have "public bool IsAnimated { get; }". Mostly I am making this suggesting because I feel that the code that uses OriginOffset does not really make use of the nullable nature of OffsetValue and OffsetExpression.

source/LottieToWinComp/Brushes.cs
aborziak-ms: Now without originOffset parameter line is shorter. It is about the length of other lines in the file, now it is ok without newlines?

source/LottieToWinComp/Rectangles.cs
eliezerpMS: Looks like some other files also have this warning. I guess we can discuss offline about using "#nullable enable" or not or if we want enable it project wide.

eliezerpMS:
In TranslateAndApplyShapeContextWithTrimOffset in Shapes.cs before you call "Brushes.TranslateAndApplyStroke(context, context.Stroke, shape, context.Opacity);" Do you need to call do the clearing out of OriginOffset here instead? It looks like TranslateAndApplyStroke can hit the linear gradient code but I assume it does not need the offset stuff factored in. If you do put the clearing here then also put a comment about how the offset is only needed for shape fills.

---
In reply to: [860911730](https://github.com/windows-toolkit/Lottie-Windows/pull/450#issuecomment-860911730) [](http://example.com/codeflow?ancestors=860911730,860911730)

source/LottieToWinComp/Brushes.cs
eliezerpMS: Yep this is fine.

aborziak-ms:
> In TranslateAndApplyShapeContextWithTrimOffset in Shapes.cs before you call "Brushes.TranslateAndApplyStroke(context, context.Stroke, shape, context.Opacity);" Do you need to call do the clearing out of OriginOffset here instead? It looks like TranslateAndApplyStroke can hit the linear gradient code but I assume it does not need the offset stuff factored in. If you do put the clearing here then also put a comment about how the offset is only needed for shape fills.

Agree, done

source/LottieToWinComp/Brushes.cs
aborziak-ms: I see, yeah, agree. Done

source/LottieToWinComp/Brushes.cs
aborziak-ms: You mean this?
image

eliezerpMS: Yep. Not for this PR but it would be interesting to discuss with you (maybe during a sync meeting) about where if at this format should be used.

eliezerpMS (approved):
:shipit:

PR #451, opened 2021-06-22
Add support for round corners if there is no animations

aborziak-ms:
Implemented After Effects algorithm for **"Round Corners"** effect. (There is a comment in Shapes.cs:634 that mentioned this _// TODO - can round corners be implemented by composing cubic Beziers?_)

For now I added support only for one-path + no-animations scenario, in following PR will also add support for multiple paths and animated path/radius. Just to keep this PR small.
 
.
.
.

To make a round corner we are stepping `radius` pixels in both directions from the corner, and joining resulting points together with a bezier curve (`MakeCubicBezier` takes 3 points and creates a bezier curve). But also there is a tricky case, when one end of the segment is already curved, this case described on picture 2.

Algorithm loops over segments and creates new array of segments, there are several cases that can happen when we are processing current segment:
0 If current segment has both ends curved we do not need to change anything, just add current segment to the result
1 If current segment has both ends that are not curved:

On this step we are creating new segments `(prevControlPoint, point0)` and `(point0, point1)`

2 If current segment has first end curved:

New point will be created on distance `radius` from the second end of the segment, on the line that connects second end of the segment and controlPoint1 of original bezier curve

.
.
.

Here how it looks in lottie (it is pixel-perfect with how it looks in After Effects) :

(red - original shape, green - with round corners)

### DIFF
--- a/source/LottieToWinComp/Paths.cs
+++ b/source/LottieToWinComp/Paths.cs
@@ -482,9 +482,9 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 else if (canRoundBegin && canRoundEnd)
                 {
                     // Both ends are not curved, so we can make them rounded.
-                    var length = (segment.ControlPoint0 - segment.ControlPoint3).Length();
-
-                    var radiusVector = (segment.ControlPoint3 - segment.ControlPoint0) / length * radius;
+                    var cp0cp3 = segment.ControlPoint3 - segment.ControlPoint0;
+                    var length = cp0cp3.Length();
+                    var radiusVector = cp0cp3.Normalized() * radius;
 
                     Vector2 point0 = segment.ControlPoint0 + radiusVector;
                     Vector2 point1 = segment.ControlPoint3 - radiusVector;
@@ -541,7 +541,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
                 if (i == -1)
                 {
-                    // If i = -1 then is was special pre-pass to get the valid value of prevControlPoint,
+                    // If i = -1 then it was special pre-pass to get the valid value of prevControlPoint,
                     // all generated segments should be ignored.
                     resultSegments.Clear();
                 }
--- a/source/LottieToWinComp/Paths.cs
+++ b/source/LottieToWinComp/Paths.cs
@@ -435,6 +435,20 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             return new BezierSegment(cp0, cp0 + ((cp1 - cp0) * 0.55), cp2 + ((cp1 - cp2) * 0.55), cp2);
         }
 
+        // The way this function works is it detects if two segments form a corner (if they do not have smooth connection)
+        // Then it duplicates this point (shared by two segments) and moves newlt generated points in different directions
+        // for "radius" pixels along the segment.
+        // After that we are joining both new points with a bezier curve to make the corner look rounded.
+        //
+        // There are 3 possible cases:
+        // 1. When the segment is curved from both sides, then we can just keep it as it is
+        // 2. When the segment is not curved from both sides, then we can make two rounded corners, from both ends.
+        // 3. When the segment curved from one side (begin or end) we are making only one rounded corner.
+        //
+        // In order to make a rounded corner we also need two points on segments next to the current segment.
+        // In this algortihm we are processing segments one by one, and passing one point from one segment to another,
+        // so that currently processed segment can create rounded corner using this point, and pass new point
+        // to the next segment, so that it can create next rounded corner and so on.
         static PathGeometry MakeRoundCorners(PathGeometry pathGeometry, double radius)
         {
             // There is no corners if we have less than two segments.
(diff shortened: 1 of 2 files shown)


source/LottieToWinComp/Paths.cs
```diff
@@ -410,5 +426,128 @@ sealed class StateCache
...
+                return pathGeometry;
+            }
+
+            var count = pathGeometry.BezierSegments.Count;
+
+            // We treat array of segments as a circular array, so that first and last elements are adjacent.
+            Func<int, int> getCircularIndex = (i) => ((i % count) + count) % count;
+
+            // Initial value does not matter, it is guranteed that it will be reassigned before use.
+            Vector2 prevControlPoint = new Vector2(0, 0);
```
eliezerpMS, line 452: >Vector2 [](http://example.com/codeflow?start=43&length=7)

Vector2.Zero?

source/LottieToWinComp/Paths.cs
```diff
@@ -410,5 +426,128 @@ sealed class StateCache
...
+                    var cp0cp3 = segment.ControlPoint3 - segment.ControlPoint0;
+                    var length = cp0cp3.Length();
+                    var radiusVector = cp0cp3.Normalized() * radius;
+
+                    Vector2 point0 = segment.ControlPoint0 + radiusVector;
+                    Vector2 point1 = segment.ControlPoint3 - radiusVector;
+
+                    // If doubled radius is greater than length, then both points collapse into
+                    // one point right in the middle of the segment.
+                    if (length <= 2 * radius)
```
eliezerpMS, line 494: >if (length <= 2 * radius) [](http://example.com/codeflow?start=20&length=25)

I assume this is what After Effects does?

aborziak-ms: Yes, it basically means that if we step `radius` pixels from both ends and `length < 2*radius`, points will change their relative order, so there is a limit, how far points can step along the segment (not more than length/2)

If we have a segment (a, b) and corner radius 2px:
Initial segment:
`a-------------b`
Step 2 pixels (ok):
`--a---------b--`
Step 8 pixels (not ok): 
`------b-a------`
so instead we are doing this (a = b = p):
`-------p-------`

source/LottieToWinComp/Paths.cs
```diff
@@ -410,5 +426,128 @@ sealed class StateCache
             public Dictionary<(PathGeometry, ShapeFill.PathFillType, bool), CompositionPath> CompositionPaths { get; }
                 = new Dictionary<(PathGeometry, ShapeFill.PathFillType, bool), CompositionPath>();
         }
+
+        static BezierSegment MakeCubicBezier(Vector2 cp0, Vector2 cp1, Vector2 cp2)
+        {
+            // This is similar to converting QudraticBezier to CubicBezier, but instead of
+            // coefficiet 2/3 we are using 0.55 . This coefficient was found experimentally by comparing
```
eliezerpMS, line 433: >2/3 we are using 0.55 [](http://example.com/codeflow?start=26&length=21)

hmm... this is a bit surprising... I would be curious if there is something they are doing for their rounded corners that makes this necessary...

source/LottieToWinComp/Paths.cs
```diff
@@ -410,5 +426,128 @@ sealed class StateCache
...
+            // images from AfterEffects and LottieViewer.
+            return new BezierSegment(cp0, cp0 + ((cp1 - cp0) * 0.55), cp2 + ((cp1 - cp2) * 0.55), cp2);
+        }
+
+        static PathGeometry MakeRoundCorners(PathGeometry pathGeometry, double radius)
```
eliezerpMS, line 438: >MakeRoundCorners [](http://example.com/codeflow?start=28&length=16)

Would you mind summarizing the three cases in a comment for this method? (Especially laying the different parts of your algorithm below like how you are doing in the PR description?)

eliezerpMS (dismissed):
:shipit:

source/LottieToWinComp/Paths.cs
aborziak-ms: I tried CubicBezier(cp0, cp1, cp1, cp2) but it was far off. Then I tried to convert qudratic bezier built on points (cp0, cp1, cp2) to cubic bezier (it is basically, the case when we have coefficient = 2/3 in this formula). It also was a bit different than in AE. Then I tried 0.5, and again, a little difference :) But when I changed it to 0.55 it became 99.9% pixel perfect. I can't spot the difference. But yeah, the constant if weird.

source/LottieToWinComp/Paths.cs
aborziak-ms: Done

source/LottieToWinComp/Paths.cs
```diff
@@ -410,5 +426,152 @@ sealed class StateCache
...
+        static BezierSegment MakeCubicBezier(Vector2 cp0, Vector2 cp1, Vector2 cp2)
+        {
+            // This is similar to converting QudraticBezier to CubicBezier, but instead of
+            // coefficiet 2/3 we are using 0.55 . This coefficient was found experimentally by comparing
+            // images from AfterEffects and LottieViewer.
+            return new BezierSegment(cp0, cp0 + ((cp1 - cp0) * 0.55), cp2 + ((cp1 - cp2) * 0.55), cp2);
+        }
+
+        // The way this function works is it detects if two segments form a corner (if they do not have smooth connection)
+        // Then it duplicates this point (shared by two segments) and moves newlt generated points in different directions
```
eliezerpMS, line 439: >newlt [](http://example.com/codeflow?start=76&length=5)

newly

eliezerpMS (dismissed):
:shipit:

source/LottieToWinComp/Paths.cs
aborziak-ms: Done

eliezerpMS (approved):
:shipit:

PR #452, opened 2021-06-25
Added support for animated round corners modifier

aborziak-ms:
In previous PR I added basic support for AE's "Round Corner" modifier but without animations. In this PR I added approximation* of animated version of this modifier. 

*We can not animate it 100% perfect as in After Effects, because in after effects they are applying this modifier to the path after interpolation happened, but in our case we can't do that since Composition layer do not support this modifier directly, so instead we are applying this modifier to each keyframe and then Composition interpolates between these kayframes. But in most cases, it looks indistinguishable from After Effects.

- Radius animated:

- Path animated:

### DIFF
--- a/source/Issues/LT0016.md
+++ b/source/Issues/LT0016.md
@@ -6,7 +6,7 @@
 The Lottie file specifies round corners on a path. This is not fully supported by Lottie-Windows. 
 It works correctly if the radius and shape path are not animated. If both radius and shape path 
 are animated, the round corner modifier will not be applied at all, otherwise if only one of them 
-is animated Lottie-Windows will try to generate an aimation but it is not guaranteed to look exactly 
+is animated Lottie-Windows will try to generate an animation but it is not guaranteed to look exactly 
 the same as in After Effects.
 
 ## Remarks
--- a/source/Issues/LT0016.md
+++ b/source/Issues/LT0016.md
@@ -1,9 +1,13 @@
-﻿[comment]: # (name:PathWithRoundCornersIsNotSupported)
-[comment]: # (text:Path with round corners is not supported.)
+﻿[comment]: # (name:PathWithRoundCornersIsNotFullySupported)
+[comment]: # (text:Path with round corners is not fully supported.)
 
 # Lottie-Windows Warning LT0016
 
-The Lottie file specifies round corners on a path. This is not currently supported by Lottie-Windows.
+The Lottie file specifies round corners on a path. This is not fully supported by Lottie-Windows. 
+It works correctly if the radius and shape path are not animated. If both radius and shape path 
+are animated, the round corner modifier will not be applied at all, otherwise if only one of them 
+is animated Lottie-Windows will try to generate an aimation but it is not guaranteed to look exactly 
+the same as in After Effects.
 
 ## Remarks
 If support for this feature is important for your scenario please provide feedback
--- a/source/Issues/LT0016.md
+++ b/source/Issues/LT0016.md
@@ -1,13 +1,9 @@
-﻿[comment]: # (name:PathWithRoundCornersIsNotFullySupported)
-[comment]: # (text:Path with round corners is not fully supported.)
+﻿[comment]: # (name:PathWithRoundCornersIsNotSupported)
+[comment]: # (text:Path with round corners is not supported.)
 
 # Lottie-Windows Warning LT0016
 
-The Lottie file specifies round corners on a path. This is not fully supported by Lottie-Windows. 
-It works correctly if the radius and shape path are not animated. If both radius and shape path 
-are animated, the round corner modifier will not be applied at all, otherwise if only one of them 
-is animated Lottie-Windows will try to generate an animation but it is not guaranteed to look exactly 
-the same as in After Effects.
+The Lottie file specifies round corners on a path. This is not currently supported by Lottie-Windows.
 
 ## Remarks
 If support for this feature is important for your scenario please provide feedback
--- a/source/LottieToWinComp/Paths.cs
+++ b/source/LottieToWinComp/Paths.cs
@@ -132,17 +132,57 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
         private static TrimmedAnimatable<PathGeometry> TranslateRoundCorners(ShapeContext context, TrimmedAnimatable<PathGeometry> pathGeometry)
         {
-            if (context.RoundCorners.Radius.IsAlways(0) || context.RoundCorners.Radius.IsAnimated)
+            var radius = context.RoundCorners.Radius;
+
+            if (radius.IsAlways(0))
             {
                 return pathGeometry;
             }
 
-            if (!pathGeometry.IsAnimated)
+            bool animated = radius.IsAnimated || pathGeometry.IsAnimated;
+
+            // Unfortunately, it is not 100% perfect when animated and does not produce animation identical to After Effects.
+            // After Effects applies radius to the shape after the interpolation, but we are applying radius to the keyframes and
+            // then Composition layer interpolates paths that are already rounded.
+            if (animated)
             {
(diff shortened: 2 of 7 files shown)


source/Issues/LT0016.md
```diff
@@ -1,9 +1,13 @@
...
+﻿[comment]: # (name:PathWithRoundCornersIsNotFullySupported)
+[comment]: # (text:Path with round corners is not fully supported.)
 
 # Lottie-Windows Warning LT0016
 
-The Lottie file specifies round corners on a path. This is not currently supported by Lottie-Windows.
+The Lottie file specifies round corners on a path. This is not fully supported by Lottie-Windows. 
+It works correctly if the radius and shape path are not animated. If both radius and shape path 
+are animated, the round corner modifier will not be applied at all, otherwise if only one of them 
+is animated Lottie-Windows will try to generate an aimation but it is not guaranteed to look exactly
```
scott-moore-ms, line 9: >aimation [](http://example.com/codeflow?start=51&length=8)

animation

source/LottieToWinComp/Paths.cs
```diff
@@ -449,7 +489,10 @@ static BezierSegment MakeCubicBezier(Vector2 cp0, Vector2 cp1, Vector2 cp2)
         // In this algortihm we are processing segments one by one, and passing one point from one segment to another,
         // so that currently processed segment can create rounded corner using this point, and pass new point
         // to the next segment, so that it can create next rounded corner and so on.
-        static PathGeometry MakeRoundCorners(PathGeometry pathGeometry, double radius)
+        //
+        // Use optimizeNumberOfPoints = false if you need to keep the number of points constant, it can bee needed
```
scott-moore-ms, line 493: >bee [](http://example.com/codeflow?start=104&length=3)

be

source/LottieToWinComp/Paths.cs
```diff
@@ -517,7 +560,18 @@ static PathGeometry MakeRoundCorners(PathGeometry pathGeometry, double radius)
...
                     if (length <= 2 * radius)
                     {
-                        point0 = point1 = (segment.ControlPoint0 + segment.ControlPoint3) * 0.5;
+                        if (optimizeNumberOfPoints)
+                        {
+                            point0 = point1 = (segment.ControlPoint0 + segment.ControlPoint3) * 0.5;
+                        }
+                        else
+                        {
+                            const double epsilon = 1e-6;
```
scott-moore-ms, line 569: >const double epsilon = 1e-6; [](http://example.com/codeflow?start=28&length=28)

Single/Double.Epsilon?

aborziak-ms: Probably `Float32.NextLargerThan` should be better (It is also used in LottieToWinComp/Animate.cs:762), since `0.5 + Float.Epsilon` will still be equal to `0.5` I guess(?). 

Done

eliezerpMS:
Would it be possible to add a gif to this PR that shows a case where AE and Lottie look different? (This would help with triaging issues we might come across in the future and maybe we can consider adding this to documentation somewhere...)

source/LottieToWinComp/TranslationIssues.cs
```diff
@@ -64,7 +64,7 @@ internal TranslationIssues(bool throwOnIssue)
         // LT0015 has been deprecated.
         // Was: Opacity and color animated at the same time is not supported.
 
-        internal void PathWithRoundCornersIsNotSupported() => Report("LT0016", "Path with round corners is not supported.");
+        internal void PathWithRoundCornersIsNotFullySupported() => Report("LT0016", "Path with round corners is not fully supported.");
```
eliezerpMS, line 67: >LT0016 [](http://example.com/codeflow?start=75&length=6)

To me this feel like it should be a different issue and we should just make this one deprecated.(considering that before they were not supported at all) This way if a user tells us they got issue LT0016 we don't also need the version number of LottieGen to know what the issue is. I am also open with breaking set here if people feel strongly about it...

source/LottieToWinComp/Paths.cs
```diff
@@ -449,7 +489,10 @@ static BezierSegment MakeCubicBezier(Vector2 cp0, Vector2 cp1, Vector2 cp2)
         // In this algortihm we are processing segments one by one, and passing one point from one segment to another,
         // so that currently processed segment can create rounded corner using this point, and pass new point
         // to the next segment, so that it can create next rounded corner and so on.
-        static PathGeometry MakeRoundCorners(PathGeometry pathGeometry, double radius)
+        //
+        // Use optimizeNumberOfPoints = false if you need to keep the number of points constant, it can be needed
+        // for animated path.
```
eliezerpMS, line 494: Suggestion: This might be a good opportunity to use the <summary> function commenting format, then when you mouse over it in VS people can see your comment and when they start typing for the parameter optimizeNumberOfPoints they can see your comment about it...

source/LottieToWinComp/Paths.cs
```diff
@@ -517,7 +560,19 @@ static PathGeometry MakeRoundCorners(PathGeometry pathGeometry, double radius)
...
+                        {
+                            point0 = point1 = (segment.ControlPoint0 + segment.ControlPoint3) * 0.5;
+                        }
+                        else
+                        {
+                            float halfPlusEpsilon = Float32.NextLargerThan(0.5f);
+                            float halfMinusEpsilon = 1 - halfPlusEpsilon;
+
+                            // Generate two points instead of one, but place them close to each other.
+                            point0 = (segment.ControlPoint0 * halfPlusEpsilon) + (segment.ControlPoint3 * halfMinusEpsilon);
```
eliezerpMS, line 573: >point0 = (segment.ControlPoint0 * halfPlusEpsilon) + (segment.ControlPoint3 * halfMinusEpsilon); [](http://example.com/codeflow?start=28&length=96)

Mind putting a comment about where these two points would be and why. Is this supposed to effectively make two points like in the optimizaton case but with +epsilon and -epsilon?

source/LottieToWinComp/Paths.cs
```diff
@@ -132,17 +132,57 @@ public static CompositionShape TranslatePathContent(ShapeContext context, Path p
...
             {
                 return pathGeometry;
             }
 
-            if (!pathGeometry.IsAnimated)
+            bool animated = radius.IsAnimated || pathGeometry.IsAnimated;
+
+            // Unfortunately, it is not 100% perfect when animated and does not produce animation identical to After Effects.
+            // After Effects applies radius to the shape after the interpolation, but we are applying radius to the keyframes and
+            // then Composition layer interpolates paths that are already rounded.
```
eliezerpMS, line 146: Mind putting a comment here describing what the slight difference is and in what case it occurs. That shoud make it easier for future people to know what is a bug and what is just this special case.

eliezerpMS, line 146: By "after the interpolation" is this interpolation the interpolation of the shape points? If so then why would that cause issues if (radius.IsAnimated && !pathGeometry.IsAnimated)?

eliezerpMS (approved):
:shipit:

source/LottieToWinComp/Paths.cs
aborziak-ms: If radius is animated and pathGeometry is not animated => then pathGeometry becomes animated, but it is mimicking the "radius animation". WinComp support animated paths only, so we need to transform this "radius animation" to "path animation".

aborziak-ms:
> Would it be possible to add a gif to this PR that shows a case where AE and Lottie look different? (This would help with triaging issues we might come across in the future and maybe we can consider adding this to documentation somewhere...)

Yes, here is this case. Red shape - original shape. Green shape - after "round corner" effect was applied. In After Effects green shape will not go out of red shape (on top edge). But in our implementation green shape basically "lost" the connection to the original shape because we are interpolating the control points of the green shape itself, not the red shape. (Ideally on each frame we should interpolate red shape and only then apply round corners, but we can't do that.)

round_corners_animated_problem

eliezerpMS (approved):
:shipit:

PR #453, opened 2021-07-16
Added lottie merge optimization

aborziak-ms:
This PR adds optimization that reduces number of WinComp objects and animations that lottie generates for AE animations that have many similar/duplicate layers.

This optimization is turned on by default in LottieViewer and in LottieGen. In LottieGen it can be turned off with a command line parameter `-DisableLottieMergeOptimizer`


source/LottieData/Optimization/Experimental/LayerGroup.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        public LayerGroup(Layer mainLayer, Layer matteLayer, bool frozen)
+        {
+            MainLayer = mainLayer;
+            MatteLayer = matteLayer;
+            Frozen = frozen;
+        }
+
+        public static List<LayerGroup> LayersToLayerGroups(List<Layer> layers, Func<Layer, Layer?, bool> isFrozenFunc)
```
eliezerpMS, line 45: >Func<Layer, Layer?, bool> isFrozenFunc [](http://example.com/codeflow?start=79&length=38)

For readability can you just put the logic in this method and then have a bool for greater or less than?

source/LottieData/Optimization/Experimental/LayerGroup.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// Matte layer. It is null for most groups.
+        /// But if it is not null then it should be placed before main layer in layer collection.
+        /// </summary>
+        public Layer? MatteLayer { get; }
+
+        /// <summary>
+        /// Indicates if layer group should not be merged with other layer groups anymore.
+        /// </summary>
+        public bool Frozen { get; }
```
eliezerpMS, line 29: >Frozen [](http://example.com/codeflow?start=20&length=6)

Maybe call this something more indicative of its context such as "ExcludeFromMerging" or something like that?

source/LottieData/Optimization/Experimental/LayersGraph.cs
```diff
@@ -0,0 +1,248 @@
...
+
+                    if (!mergeMatterRes.Success)
+                    {
+                        continue;
+                    }
+
+                    a.MergedGroup = b.MergedGroup = new LayerGroup(
+                        mergeRes.Value!,
+                        mergeMatterRes.Value!,
+                        true
```
eliezerpMS, line 207: >true [](http://example.com/codeflow?start=24&length=4)

nit: use syntax like:
new LayerGroup(
                        mainLayer: mergeRes.Value!,
                        matteLayer: mergeMatterRes.Value!,
                        frozen: true);

This is especially useful when passing in something like "true" so the reader has more context.

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
```diff
@@ -0,0 +1,994 @@
...
+            return OriginalComposition.Assets.GetAssetById(id) ?? NewAssets.Find((Asset a) => a.Id == id);
+        }
+
+        public LottieCompositionOptimizer(LottieComposition composition)
+        {
+            OriginalComposition = composition;
+            NewAssets = new List<Asset>();
+        }
+
+        public LottieComposition GetOptimized()
```
eliezerpMS, line 39: >GetOptimized [](http://example.com/codeflow?start=33&length=12)

Definitely put a comment summarizing the algorithm.

raymond-wh-leung:
Dumb questions, what is an example of the optimized animations? (i.e. what are the optimized animations look like beofre / after the change?)

aborziak-ms:
> Dumb question, what is an example of the optimized animations? (i.e. what are the optimized animations look like beofre / after the change?)

They look the same visually (at least they should, if there is no bugs, but I didn't spot any differences yet). The only difference that now one WinComp object can be responsible for rendering multiple segments of the animation. For example: if earlier one Visual node was responsible for the animation on segment [5; 15] (frames), now one Visual node can be responsible for animating [5; 15], [40; 90], [120; 130] ranges.

raymond-wh-leung:
Ok got it. (just to clarify was wondering how the animation defintion looks like, not how it looks visually. :) never mind if you don't have that handy) Fewer animations (with more keyframes) should be good, would help reduce some operations on dwm side.. Just an Fyi, I've saved a copy of how the explorer's animation defintions look like on dwm side at: \\wexfs\users\rayleung\Temp\dwm_animations_explorer\explore_animations.txt. (the format of the list is like: type of animation (kfa or expression), target object/property of the animation, and some details of each animation etc.)

---
In reply to: [881776374](https://github.com/CommunityToolkit/Lottie-Windows/pull/453#issuecomment-881776374) [](http://example.com/codeflow?ancestors=881776374,881776374)

source/LottieData/Optimization/Experimental/LayerGroup.cs
aborziak-ms: In theory this class should not know the outer logic that stands behind isFrozen (canBeMerged). It only transforms layers to groups and groups to layers.

source/LottieData/Layer.cs
```diff
@@ -123,6 +123,28 @@ private protected Layer(in LayerArgs args)
             public MatteType LayerMatteType { get; set; }
         }
 
+        public LayerArgs CopyArgs()
```
eliezerpMS, line 126: >CopyArgs [](http://example.com/codeflow?start=25&length=8)

Could we just use clone here?

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -286,6 +287,9 @@ LottieComposition ParseLottieComposition(ref Reader reader)
                                                     fonts: fonts,
                                                     layers: new LayerCollection(layers),
                                                     markers: markers);
+
+                                result = LottieCompositionOptimizer.GetOptimized(result);
```
eliezerpMS, line 291: >GetOptimized [](http://example.com/codeflow?start=68&length=12)

If this is under a directory "Experimental" should it be on by default? I am ok if we do want it to be on by default but just like with the other optimizations it should have a command line parameter that can turn it off.

source/LottieData/Optimization/Experimental/LayerGroup.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    class LayerGroup
+    {
+        /// <summary>
+        /// Main layer of the group.
+        /// </summary>
+        public Layer MainLayer { get; }
+
+        /// <summary>
+        /// Matte layer. It is null for most groups.
```
eliezerpMS, line 25: >It is null for most groups [](http://example.com/codeflow?start=25&length=26)

Awesome! I like this convention of documenting what it means for a nullable to be null!

source/LottieData/Optimization/Experimental/LayersGraph.cs
```diff
@@ -0,0 +1,229 @@
...
+                    {
+                        node.AddParent(other);
+                    }
+                }
+
+                nodes.Add(node);
+            }
+        }
+
+        struct MergableNodePair
```
eliezerpMS, line 87: >MergableNodePair [](http://example.com/codeflow?start=15&length=16)

nit: just alias a tuple for this instead?

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
```diff
@@ -0,0 +1,112 @@
...
+using static Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Layer;
+using static Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.ShapeLayerContent;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
+{
+#if PUBLIC_LottieData
+    public
+#endif
+
+    sealed class LottieCompositionOptimizer
```
eliezerpMS, line 22: >LottieCompositionOptimizer [](http://example.com/codeflow?start=17&length=26)

nit: put a comment here about what this optimizer does (maybe something about layers?) then we can know to associate it with the layer graph and group in the future.

source/LottieData/Optimization/Experimental/LayerGroup.cs
eliezerpMS: I see now that it has the extra data you are including in the function that you pass in so that makes sense why it is not in here.

source/LottieData/Optimization/Experimental/LayersGraph.cs
```diff
@@ -0,0 +1,229 @@
...
+            foreach (var group in layerGroups)
+            {
+                var node = new GraphNode(group);
+                var range = Range.GetForLayer(group.MainLayer);
+
+                var reversedNodes = nodes.Select(v => v).ToList();
+                reversedNodes.Reverse();
+                foreach (var other in reversedNodes)
+                {
+                    if (!node.IsChildOf(other) && Range.GetForLayer(other.Group.MainLayer).Intersect(range))
```
eliezerpMS, line 77: >Range.GetForLayer(other.Group.MainLayer) [](http://example.com/codeflow?start=50&length=40)

This can just be range?

source/LottieData/Optimization/Experimental/LayersGraph.cs
```diff
@@ -0,0 +1,229 @@
...
+                MergedWithNode = null;
+                MergedGroup = null;
+            }
+
+            public void AddParent(GraphNode parent)
+            {
+                Parents.Add(parent);
+            }
+
+            public bool IsChildOf(GraphNode node, HashSet<GraphNode>? visited = null)
```
eliezerpMS, line 38: >visited [](http://example.com/codeflow?start=70&length=7)

Do callers ever use this?

source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -0,0 +1,837 @@
...
+            }
+
+            var mainLayerMergeRes = MergeLayers(a.MainLayer, b.MainLayer);
+
+            if (!mainLayerMergeRes.Success)
+            {
+                return Result<LayerGroup>.Failed;
+            }
+
+            if (a.MatteLayer is null && b.MatteLayer is null)
```
eliezerpMS, line 651: >if (a.MatteLayer is null && b.MatteLayer is null) [](http://example.com/codeflow?start=12&length=49)

Put a comment here again highlight the limitation with mattelayers.

source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -0,0 +1,837 @@
...
+        /// <returns>Result of merging two layers.</returns>
+        public Result<Layer> MergeLayers(Layer a, Layer b, bool ignoreParent = false)
+        {
+            // Layer a must go be before layer b
+            if (a.InPoint > b.InPoint)
+            {
+                return MergeLayers(b, a, ignoreParent);
+            }
+
+            if (a.Type != b.Type ||
```
eliezerpMS, line 688: >(a.Type != b.Type || [](http://example.com/codeflow?start=15&length=20)

Can we put most of this into a static in Layer.cs?

eliezerpMS, line 688: >a.Type != b.Type || [](http://example.com/codeflow?start=16&length=19)

What about having some of this in a "public bool Equals(Layer? other)" method? 

Also for the parts that are not tested yet then you can just bail in the equals implementation.

source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -0,0 +1,837 @@
...
+                case LayerType.Shape:
+                    return Result<Layer>.From(MergeShapeLayers((ShapeLayer)a, (ShapeLayer)b));
+                case LayerType.Null:
+                    return Result<Layer>.From(MergeNullLayers((NullLayer)a, (NullLayer)b));
+            }
+
+            return Result<Layer>.Failed;
+        }
+
+        Result<ShapeLayer> MergeShapeLayers(ShapeLayer a, ShapeLayer b)
```
eliezerpMS, line 718: >MergeShapeLayers [](http://example.com/codeflow?start=27&length=16)

It looks like most of these merge methods just have the type types to merge as parameters. How do you feel with having them in there respective classes?

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
```diff
@@ -0,0 +1,112 @@
...
+            // This heuristic is needed because in theory any two layers can be merged, but some of them can be merged with higher score.
+            // And if we merge pairs with worse score first - it can result in worse optimization overall.
+            // So we are trying to merge pairs of layers with the higher score first, while trying to merge pairs that are nerby
+            // because if we merge two layers that are far away then we can't insert any other layer between them (for merging).
+            // Example: we have three layers [0; 10] [30; 40] [70; 80]
+            // It is better to merge [0; 10] and [30; 40] first (result is [0; 40]), and then merge [70; 80] with the result.
+            // If we merge [0; 10] and [70; 80] first, we will get [0; 80] as the result, and we will not be able to merge it with [30; 40]
+            // because they intersect.
+            // TODO: there is surely a better heuristic, we need more experiments to find it.
+            for (double minimalScore = 0.95; minimalScore > 0.1; minimalScore *= 0.8)
```
eliezerpMS, line 51: >for (double minimalScore = 0.95; minimalScore > 0.1; minimalScore *= 0.8) [](http://example.com/codeflow?start=12&length=73)

How were these numbers determined? What sample input was used to determine them?

source/LottieReader/Serialization/LottieCompositionReader.cs
eliezerpMS: Also since this will incur a runtime cost maybe it should be off by default for non-lottie gen cases?

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
eliezerpMS: I am also wondering, how do you feel about having these are parameters that someone could pass in? 

Then you could empower the user to find the best fit. Although if we do go that route it might be interesting to somehow output the results of the optimization.

source/LottieData/Optimization/Experimental/ShiftUtils.cs
```diff
@@ -0,0 +1,227 @@
...
+using static Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.ShapeLayerContent;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
+{
+    /// <summary>
+    /// Set of methods to shift time range of different Lottie objects. Used by optimizer.
+    /// </summary>
+    static class ShiftUtils
+    {
+        static Animatable<T> ShiftAnimatable<T>(Animatable<T> a, double shift)
```
eliezerpMS, line 16: >shift [](http://example.com/codeflow?start=72&length=5)

Nit: rename this to have something to do with frames or put a comment on the method about the parameters.

eliezerpMS: same with the other methods below.

source/LottieData/Optimization/Experimental/ShiftUtils.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        static IAnimatableVector3 ShiftIAnimatableVector3(IAnimatableVector3 a, double shift)
+        {
+            if (a is AnimatableVector3)
+            {
+                var v = ShiftAnimatable((AnimatableVector3)a, shift);
+                return v.IsAnimated ? new AnimatableVector3(v.KeyFrames) : new AnimatableVector3(v.InitialValue);
+            }
+
+            Debug.Assert(a is AnimatableXYZ, "There are only AnimatableXYZ and AnimatableVector3 implementations");
```
eliezerpMS, line 42: >AnimatableXYZ [](http://example.com/codeflow?start=61&length=13)

nit: for this and any of the other messages change them to be:

$"There are only {typeof(AnimatableXYZ)} and {typeof(AnimatableVector3)} implementations"

source/LottieData/Optimization/Experimental/ShiftUtils.cs
```diff
@@ -0,0 +1,227 @@
...
+
+            foreach (var kf in a.KeyFrames)
+            {
+                keyFrames.Add(new KeyFrame<T>(kf.Frame + shift, kf.Value, kf.SpatialBezier, kf.Easing));
+            }
+
+            return new Animatable<T>(keyFrames);
+        }
+
+        static IAnimatableVector3 ShiftIAnimatableVector3(IAnimatableVector3 a, double shift)
```
eliezerpMS, line 34: >ShiftIAnimatableVector3 [](http://example.com/codeflow?start=34&length=23)

Where it makes sense how do you feel about putting these in their respective classes?

source/LottieData/Optimization/Experimental/ShiftHelper.cs
```diff
@@ -0,0 +1,263 @@
...
+    /// <summary>
+    /// Set of methods to shift time range of different Lottie objects. Used by optimizer.
+    /// </summary>
+#if PUBLIC_LottieData
+    public
+#endif
+
+    class ShiftHelper
+    {
+        Animatable<T> ShiftAnimatable<T>(Animatable<T> a, double shift)
```
eliezerpMS, line 19: >shift [](http://example.com/codeflow?start=65&length=5)

Nit: rename this to have something to do with frames or put a comment on the method about the parameters.

eliezerpMS: same with the other methods below.

source/LottieData/Optimization/Experimental/ShiftHelper.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            foreach (var kf in a.KeyFrames)
+            {
+                keyFrames.Add(new KeyFrame<T>(kf.Frame + shift, kf.Value, kf.SpatialBezier, kf.Easing));
+            }
+
+            return new Animatable<T>(keyFrames);
+        }
+
+        IAnimatableVector3 ShiftIAnimatableVector3(IAnimatableVector3 a, double shift)
```
eliezerpMS, line 37: >ShiftIAnimatableVector3 [](http://example.com/codeflow?start=27&length=23)

Where it makes sense how do you feel about putting these in their respective classes?

source/LottieData/Optimization/Experimental/ShiftHelper.cs
```diff
@@ -0,0 +1,263 @@
...
+        {
+            if (a is AnimatableVector3)
+            {
+                var v = ShiftAnimatable((AnimatableVector3)a, shift);
+                return v.IsAnimated ? new AnimatableVector3(v.KeyFrames) : new AnimatableVector3(v.InitialValue);
+            }
+
+            Debug.Assert(a is AnimatableXYZ, "There are only AnimatableXYZ and AnimatableVector3 implementations");
```
eliezerpMS, line 45: >"There are only AnimatableXYZ and AnimatableVector3 implementations" [](http://example.com/codeflow?start=45&length=68)

nit: for this and any of the other messages change them to be:

$"There are only {typeof(AnimatableXYZ)} and {typeof(AnimatableVector3)} implementations"

source/LottieData/Optimization/Experimental/ShiftUtils.cs
```diff
@@ -0,0 +1,227 @@
...
+using Microsoft.Toolkit.Uwp.UI.Lottie.Animatables;
+using static Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Layer;
+using static Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.ShapeLayerContent;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
+{
+    /// <summary>
+    /// Set of methods to shift time range of different Lottie objects. Used by optimizer.
+    /// </summary>
+    static class ShiftUtils
```
eliezerpMS, line 14: >ShiftUtils [](http://example.com/codeflow?start=17&length=10)

Hmm mabye you meant to delete this class?

source/LottieData/Optimization/Experimental/ShiftHelper.cs
```diff
@@ -0,0 +1,263 @@
...
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
+{
+    /// <summary>
+    /// Set of methods to shift time range of different Lottie objects. Used by optimizer.
+    /// </summary>
+#if PUBLIC_LottieData
+    public
+#endif
+
+    class ShiftHelper
```
eliezerpMS, line 17: >ShiftHelper [](http://example.com/codeflow?start=10&length=11)

Does this have state that it is keeping or can it be static?

source/LottieData/Optimization/Experimental/Result.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
+{
+#if PUBLIC_LottieData
+    public
+#endif
+
+    struct Result<T>
```
eliezerpMS, line 11: >Result [](http://example.com/codeflow?start=11&length=6)

Nit: Maybe name this to something like "merge result"?

source/LottieData/Optimization/Experimental/Range.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
+{
+    /// <summary>
+    /// Represents an arbitrary time range.
+    /// </summary>
+#if PUBLIC_LottieData
+    public
+#endif
+
+    struct Range
```
eliezerpMS, line 13: >Range [](http://example.com/codeflow?start=11&length=5)

FrameTimeRange?

source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -0,0 +1,837 @@
...
+            AssetsOriginal.AddRange(composition.Assets);
+        }
+
+        private Asset? GetAssetById(string id)
+        {
+            return AssetsOriginal.Find((Asset a) => a.Id == id) ?? AssetsGenerated.Find((Asset a) => a.Id == id);
+        }
+
+        /// <summary>
+        /// Optimizer pass. Builds a graph of layer groups and merge all layer groups that are still can be merged and
```
eliezerpMS, line 37: >are [](http://example.com/codeflow?start=91&length=3)

nit: remove

source/LottieData/Optimization/Experimental/ShiftHelper.cs
aborziak-ms: Yeah, I was also thinking about it. But the problem is that these methods are very specific to this optimization I think. They use the fact that parent PreComp layer will hide everything that is outside of provided ranges. 

If we can come up with some nice unification for these methods to work in any case (with ranges provided, or without) then yeah, probably it can make sense

source/LottieReader/Serialization/LottieCompositionReader.cs
eliezerpMS: Since we have the "while (MakeOptimizerPass(layerGroups))" it might be interesting to somehow output the time that codegen is taking in this optimization pass.

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
aborziak-ms: These numbers were chosen without comparing them with others (yet). I'm trying to come up with an more obvious and easy heuristic, that has similar results. "Merge layers with highest score first" or something like this.

source/LottieData/Optimization/Experimental/MergeHelper.cs
aborziak-ms: Hm, but them Equals method will work weirdly? Because here for some fields I'm using custom logic (ignoreParent for example, or InPoint OutPoint comparisons)

source/LottieData/Layer.cs
aborziak-ms: I need to extract LayerArgs structure from Layer class somehow and some fields are even called differently atm. And also Layer does not inherit LayerArgs. Or what do you mean by clone?

source/LottieData/Optimization/Experimental/LayersGraph.cs
aborziak-ms: you mean `new Range(layer)` ?

source/LottieData/Optimization/Experimental/ShiftUtils.cs
aborziak-ms: oh, I don't know how I missed it. yeah, it should be deleted

source/LottieData/Optimization/Experimental/Result.cs
aborziak-ms: Agree, previously I was also using it in Shift- methods, that is why it is called just Result<>

source/LottieData/Optimization/Experimental/LayersGraph.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        struct MergableNodePair
+        {
+            public GraphNode First { get; set; }
+
+            public GraphNode Second { get; set; }
+        }
+
+        List<MergableNodePair> GetMergableNodes()
```
eliezerpMS, line 94: >GetMergableNodes [](http://example.com/codeflow?start=31&length=16)

nit: rename to "FindUnrelatedNodePairs" or something like that? (It coudl still fail to merge later right?)

source/LottieData/Optimization/Experimental/LayersGraph.cs
eliezerpMS: never mind I was confusing this for something else.

source/LottieData/Optimization/Experimental/MergeHelper.cs
eliezerpMS: Hmm you could call it something other than Equals. My concern is that if someone in the future adds something to layer it will be less obvious that they need to edit this.

source/LottieData/Layer.cs
eliezerpMS: That makes sense.

source/LottieData/Optimization/Experimental/MergeHelper.cs
aborziak-ms: Ah ok. yeah, then it can make even more sense to move the "merge" and "shift" logic to the Lottie classes as well as these "equals"-like methods

source/LottieData/Optimization/Experimental/LayersIndexMapper.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        public int GetMapping(int oldIndex)
+        {
+            return indexMapping[oldIndex];
+        }
+
+        public void RemapLayer(Layer layer)
+        {
+            // TODO: create layer copy instad of assigning new index(!)
```
eliezerpMS, line 28: >instad [](http://example.com/codeflow?start=39&length=6)

instead

source/LottieData/Optimization/Experimental/LayersIndexMapper.cs
```diff
@@ -0,0 +1,47 @@
...
+    class LayersIndexMapper
+    {
+        Dictionary<int, int> indexMapping = new Dictionary<int, int>();
+
+        public void SetMapping(int oldIndex, int newIndex)
+        {
+            indexMapping[oldIndex] = newIndex;
+        }
+
+        public int GetMapping(int oldIndex)
```
eliezerpMS, line 21: >GetMapping [](http://example.com/codeflow?start=19&length=10)

make private or remove?

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
eliezerpMS, line 22: >LottieCompositionOptimizer [](http://example.com/codeflow?start=17&length=26)

I am wondering if there is a good place to mention what sort of AE file is best suited for this (Basically our markers setup) so that a year or two later when people run this and say they get different results on different files we can better refresh our memory.

eliezerpMS, line 22: >LottieCompositionOptimizer [](http://example.com/codeflow?start=17&length=26)

Nit: I am wondering if this maybe should have a different name like "LayerTreeMerger" or something more specific to what it does?

source/LottieData/Optimization/Experimental/ShiftHelper.cs
aborziak-ms: Agree, I just moved all the `Shift*` methods to the Lottie classes. And just noticed that Animatable already has this functionality `WithTimeOffset`, so I decided to use the same name convention.

I also agree that we should somehow move `Merge*` methods inside classes. But then we will probably need to pass some context variable along.

source/LottieGen/CommandLineOptions.cs
```diff
@@ -183,6 +190,7 @@ enum Keyword
             TargetUapVersion,
             TestMode,
             WinUIVersion,
+            DisableLottieMergeOptimizer,
```
eliezerpMS, line 193: >DisableLottieMergeOptimizer [](http://example.com/codeflow?start=12&length=27)

Alphabetize

source/LottieGen/CommandLineOptions.cs
```diff
@@ -252,7 +260,8 @@ void ParseCommandLineStrings(string[] args)
                 .AddPrefixedKeyword(Keyword.Strict)
                 .AddPrefixedKeyword(Keyword.TargetUapVersion)
                 .AddPrefixedKeyword(Keyword.TestMode)
-                .AddPrefixedKeyword(Keyword.WinUIVersion);
+                .AddPrefixedKeyword(Keyword.WinUIVersion)
+                .AddPrefixedKeyword(Keyword.DisableLottieMergeOptimizer);
```
eliezerpMS, line 264: >DisableLottieMergeOptimizer [](http://example.com/codeflow?start=44&length=27)

Alphabetize

source/LottieGen/LottieJsonFileProcessor.cs
```diff
@@ -129,6 +131,16 @@ bool Run(Stream jsonStream)
                 return false;
             }
 
+            if (!_options.DisableLottieMergeOptimizer)
+            {
+                Stopwatch stopWatch = new Stopwatch();
```
eliezerpMS, line 136: > Stopwatch stopWatch = new Stopwatch(); [](http://example.com/codeflow?start=15&length=39)

Could be for this change or future one but maybe consider integrating this with the Profiler.

source/LottieGen/CommandLineOptions.cs
aborziak-ms: Done

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -286,6 +287,9 @@ LottieComposition ParseLottieComposition(ref Reader reader)
                                                     fonts: fonts,
                                                     layers: new LayerCollection(layers),
                                                     markers: markers);
+
+                                result = LottieMergeOptimizer.Optimize(result);
```
eliezerpMS, line 291: >result = LottieMergeOptimizer.Optimize(result); [](http://example.com/codeflow?start=32&length=47)

We probably want this to factor in _options.DisableLottieMergeOptimizer?

aborziak-ms: But this parameter is lottieGen-only command line arg? We do not have it for Lottie Viewer, right?

source/LottieData/Optimization/Experimental/LottieCompositionOptimizer.cs
aborziak-ms: Added TODO to add more clear way of choosing these constants.

source/LottieData/Optimization/Experimental/MergeHelper.cs
aborziak-ms: For now probably let's keep this MergeHelper as it is. It seems that we will not implement "Merge" methods for every lottie object. But if we will, I will move them inside lottie classes in future PRs. And I'm still not sure about nice interface for this, since sometimes we need to specify these timeRanges and need to keep track of generated assets.

eliezerpMS: Alright fair enough. You can keep it in the back of your mind for another time.

source/LottieReader/Serialization/LottieCompositionReader.cs
eliezerpMS: You are correct. I mixed up the _options here.

source/WinCompData/KeyFrameAnimation.cs
```diff
@@ -65,6 +65,16 @@ void InsertKeyFrameCommon(float progress, T value, CompositionEasingFunction? ea
             _keyFrames[progress] = new ValueKeyFrame(progress, easing, value);
         }
 
+        public void InsertKeyFrame(KeyFrame keyFrame)
```
eliezerpMS, line 68: >InsertKeyFrame [](http://example.com/codeflow?start=20&length=14)

(Not sure why I did not ask before) but just curious where is this used?

eliezerpMS (dismissed):
:shipit:

eliezerpMS (approved):
:shipit:

eliezerpMS (approved):
:shipit:

PR #454, opened 2021-07-29
Enabled blur effect for PreComp layers

aborziak-ms:
Blur effect rewritten to use SurfaceBrush. Now there is no more visual glitches.
Also added CompositionMaskBrush since we need it to apply shadow color to the Visual node.

### DIFF
--- a/source/LottieToWinComp/PreComps.cs
+++ b/source/LottieToWinComp/PreComps.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-// DropShadows are currently disabled because of LayerVisual bugs.
-//#define EnableDropShadow
 #nullable enable
 
 using System;
@@ -73,16 +71,12 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                 result.Children.Add(rootNode);
             }
 
-#if EnableDropShadow
             var dropShadowEffect = context.Effects.DropShadowEffect;
 
             if (dropShadowEffect is not null)
             {
                 result = ApplyDropShadow(context, result, dropShadowEffect);
             }
-#else
-            context.Effects.EmitIssueIfDropShadow();
-#endif
 
             var gaussianBlurEffect = context.Effects.GaussianBlurEffect;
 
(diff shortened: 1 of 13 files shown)


source/UIData/Tools/GraphCompactor.cs
```diff
@@ -101,6 +101,14 @@ void Compact(ObjectGraph<Node> graph)
                             graph[child].Parent = node.Object;
                         }
 
+                        break;
+                    case CompositionObjectType.CompositionVisualSurface:
+                        Visual? source = ((CompositionVisualSurface)node.Object).SourceVisual;
+                        if (source is not null)
+                        {
+                            graph[source].IsSource = true;
```
aborziak-ms, line 109: Optimization can merge two ContainerVisuals together if they have common parent. So I decided to mark node that it is used as source of VisualSurface to not optimize it out

source/UIData/Tools/GraphCompactor.cs
```diff
@@ -1074,7 +1082,7 @@ void CoalesceContainerVisuals(ObjectGraph<Node> graph)
 
                  // The parent may have been removed already.
                  let parent = n.Node.Parent
-                 where parent is not null
+                 where parent is not null && !n.Node.IsSource
```
eliezerpMS, line 1085: >IsSource [](http://example.com/codeflow?start=53&length=8)

If this is only used here could we just do the check we do in Compact here and not have to add the IsSource?

source/LottieToWinComp/PreComps.cs
```diff
@@ -97,59 +91,81 @@ static class PreComps
         /// <summary>
         /// Applies the given <see cref="DropShadowEffect"/>.
         /// </summary>
-        static LayerVisual ApplyDropShadow(PreCompLayerContext context, Visual visual, DropShadowEffect dropShadowEffect)
+        static ContainerVisual ApplyDropShadow(PreCompLayerContext context, ContainerVisual source, DropShadowEffect dropShadowEffect)
```
eliezerpMS, line 94: >(PreCompLayerContext cont [](http://example.com/codeflow?start=46&length=25)

nit: wrap these params

source/LottieToWinComp/PreComps.cs
```diff
@@ -203,7 +227,7 @@ static LayerVisual ApplyDropShadow(PreCompLayerContext context, Visual visual, D
         static Vector3 VectorFromRotationAndDistance(double directionRadians, double distance) =>
             new Vector3(
                 x: Math.Sin(directionRadians) * distance,
-                y: Math.Cos(directionRadians) * distance,
+                y: -Math.Cos(directionRadians) * distance,
```
eliezerpMS, line 230: Why negative? Also I am having trouble remembering but I wonder why we have sin for x.... and 1 for z.... I guess I will have to re-review the code....

source/LottieToWinComp/PreComps.cs
```diff
@@ -97,59 +91,81 @@ static class PreComps
...
+            // |ColorBrush|   | VisualSurface |
+            // +----------+   +---------------+
+            GaussianBlurEffect gaussianBlurEffect = new GaussianBlurEffect(
+                name: dropShadowEffect.Name + "_blur",
+                isEnabled: true,
+                blurriness: dropShadowEffect.Softness,
+                blurDimensions: new Animatable<Enum<BlurDimension>>(BlurDimension.HorizontalAndVertical),
+                repeatEdgePixels: new Animatable<bool>(true),
+                forceGpuRendering: true
+                );
```
eliezerpMS, line 129: > ) [](http://example.com/codeflow?start=15&length=2)

nit: put on the previous line.

source/WinCompData/CompositionMaskBrush.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData.Mgce;
+
+namespace Microsoft.Toolkit.Uwp.UI.Lottie.WinCompData
+{
+    [MetaData.UapVersion(2)]
```
eliezerpMS, line 10: >UapVersion [](http://example.com/codeflow?start=14&length=10)

Version 2? [https://docs.microsoft.com/en-us/uwp/api/windows.ui.composition.compositionmaskbrush?view=winrt-20348](https://docs.microsoft.com/en-us/uwp/api/windows.ui.composition.compositionmaskbrush?view=winrt-20348)

source/LottieToWinComp/PreComps.cs
aborziak-ms: In AE 0° is up, and 90° is right. So for 0° we need (0, -1) direction vector and (1, 0) for 90°

source/UIData/Tools/GraphCompactor.cs
aborziak-ms: Hm, it is not very clear to me. We can't detect if Visual node was used as Source somewhere else if we didn't walk the graph, right?

source/WinCompData/CompositionMaskBrush.cs
aborziak-ms: Oh, missed that, right

source/LottieToWinComp/PreComps.cs
```diff
@@ -97,59 +91,81 @@ static class PreComps
...
             Debug.Assert(dropShadowEffect.IsEnabled, "Precondition");
 
-            // Create a LayerVisual so we can add a drop shadow.
-            var result = context.ObjectFactory.CreateLayerVisual();
-            result.Children.Add(visual);
+            // Shadow:
+            // +------------------+
+            // | Container Visual | -- Has the final composited result.
+            // +------------------+ <------
+            //     ^ Child                | Child
```
eliezerpMS, line 102: >Child [](http://example.com/codeflow?start=44&length=5)

I assume this is the actual object that will have the shadow? If so could you put a comment to clarify that?

source/LottieToWinComp/PreComps.cs
```diff
@@ -97,59 +91,81 @@ static class PreComps
...
+            // +----------------+           |
+            // | SpriteVisual   |           |
+            // +----------------+           |
+            //     ^ Source                 |
+            //     |                        |
+            // +--------------+             |
+            // | MaskBrush    |             |
+            // +--------------+             |
+            //     ^ Source   ^ Mask        |
+            //     |           \            V
```
eliezerpMS, line 118: >V [](http://example.com/codeflow?start=44&length=1)

Can you put a "lable" here to further clarify that this is the source visual (and maybe remove the arrow) since this is not a parent child relationship (or maybe use a line of "." instead?

source/LottieToWinComp/PreComps.cs
```diff
@@ -97,59 +91,81 @@ static class PreComps
...
+            var visualSurface = factory.CreateVisualSurface();
+            visualSurface.SourceSize = size;
+            visualSurface.SourceVisual = source;
 
-            result.Shadow = shadow;
-            shadow.SourcePolicy = CompositionDropShadowSourcePolicy.InheritFromVisualContent;
+            var maskBrush = factory.CreateMaskBrush();
 
-            var isShadowOnly = Optimizer.TrimAnimatable(context, dropShadowEffect.IsShadowOnly);
-            if (!isShadowOnly.IsAlways(true))
```
eliezerpMS, line 118: > if (!isShadowOnly.IsAlways(true)) [](http://example.com/codeflow?start=11&length=34)

We don't still want this? Do we correctly handle if this is animated?

source/LottieToWinComp/PreComps.cs
eliezerpMS: I see. That is strange....

source/UIData/Tools/GraphCompactor.cs
```diff
@@ -1443,6 +1451,8 @@ void CopyDescriptions(IDescribable from, IDescribable to)
         sealed class Node : Graph.Node<Node>
         {
             internal CompositionObject? Parent { get; set; }
+
+            internal bool IsSource { get; set; }
```
eliezerpMS, line 1455: >IsSource [](http://example.com/codeflow?start=26&length=8)

If this is only used to VisualSurface source how about caling it something like "AllowCoalesing" or something like that? So it is clearer what it is used for not really just its context.

source/WinCompData/CompositionMaskBrush.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
```
eliezerpMS, line 5: >Generic [](http://example.com/codeflow?start=25&length=7)

is this used?

source/LottieToWinComp/PreComps.cs
```diff
@@ -203,7 +227,7 @@ static LayerVisual ApplyDropShadow(PreCompLayerContext context, Visual visual, D
         static Vector3 VectorFromRotationAndDistance(double directionRadians, double distance) =>
```
eliezerpMS, line 227: >VectorFromRotationAndDistance [](http://example.com/codeflow?start=23&length=29)

So that I don't forget again would you mind putting a comment about that 0 degrees is up and not right.

source/LottieToWinComp/PreComps.cs
aborziak-ms: I think it just and Adobe convention, it is the same in other their apps
In UI it looks like this:
image

eliezerpMS: Yeah I guess that does make sense from a intuitiveness perspective...

source/LottieToWinComp/PreComps.cs
aborziak-ms: Hm, it seems that it will handle correctly only not animated scenario. Can we animate IsVisible parameter? Then I can flip true/false value of IsShadowOnly animation and apply to IsVisible property

eliezerpMS: Yeah I was thinking that is what you could do. I will leave it up to you if you want to do it for this PR or consider it in the future.

source/LottieToWinComp/PreComps.cs
eliezerpMS: :grinning: I like the diagonal taking it to the next level!

source/LottieToWinComp/PreComps.cs
aborziak-ms: Done

eliezerpMS (approved):
:shipit:

PR #455, opened 2021-07-30
Added effects (blur/shadow) support for shape layers.

aborziak-ms:
Turned out that we can pretty easily add blur/shadow support for shape layers. We can just reuse the code from PreComps.cs.

I split this change into two commits. In first one I changed the code and in second one I moved a big chunk of code from PreComps.cs to Effects.cs because now it is used by both PreComps and Shapes tranlators.


source/LottieToWinComp/PreComps.cs
```diff
@@ -218,11 +224,11 @@ static class PreComps
...
 
-            // Check if ShadowOnly can be true
-            if (!dropShadowEffect.IsShadowOnly.IsAlways(false))
+            // Check if ShadowOnly can be false
+            if (!dropShadowEffect.IsShadowOnly.IsAlways(true))
             {
-                // Check if ShadowOnly can be false
-                if (!dropShadowEffect.IsShadowOnly.IsAlways(true))
+                // Check if ShadowOnly can be true
+                if (!dropShadowEffect.IsShadowOnly.IsAlways(false))
```
aborziak-ms: Turned out that there was a bug. I mixed up IsShadowOnly and IsVisible

source/LottieToWinComp/PreComps.cs
```diff
@@ -133,7 +134,12 @@ static class PreComps
                 forceGpuRendering: true);
 
             var factory = context.ObjectFactory;
-            var size = ConvertTo.Vector2(context.Layer.Width, context.Layer.Height);
+            var size = context.CompositionContext.Size;
```
aborziak-ms: It seems that CompositionContext.Size equals to parent PreComp size. So I'm using it here since we can't use Layer.Width/Height for non-PreComp layers.

source/LottieToWinComp/PreComps.cs
eliezerpMS: I am not at all familiar with what this is in AE but from the code comments (I did not add them) on the property it sounds like it should sort of affect visibility of the caster right? What was the issue you were seeing?

source/LottieToWinComp/Effects.cs
```diff
@@ -91,5 +96,298 @@ void EmitIssueAboutUnsupportedEffect(Effect? effect, string effectName)
...
+            {
+                keyFrames.Add(new KeyFrame<bool>(keyFrame.Frame, !keyFrame.Value, keyFrame.Easing));
+            }
+
+            return new Animatable<bool>(!animatable.InitialValue, keyFrames);
+        }
+
+        /// <summary>
+        /// Applies a Gaussian blur effect to the given <paramref name="source"/> and
+        /// returns a new root. This is only designed to work on a <see cref="PreCompLayer"/>
```
eliezerpMS, line 296: >PreCompLayer [](http://example.com/codeflow?start=78&length=12)

Update to add Shape?

source/LottieToWinComp/PreComps.cs
eliezerpMS: Sorry,  it must have been late and I did not catch it last night!

source/LottieToWinComp/Effects.cs
```diff
@@ -91,5 +96,298 @@ void EmitIssueAboutUnsupportedEffect(Effect? effect, string effectName)
...
+            visualSurface.SourceSize = size;
+
+            var surfaceBrush = factory.CreateSurfaceBrush(visualSurface);
+
+            var effect = new WinCompData.Mgce.GaussianBlurEffect();
+
+            var blurriness = Optimizer.TrimAnimatable(context, gaussianBlurEffect.Blurriness);
+            if (blurriness.IsAnimated)
+            {
+                context.Issues.AnimatedLayerEffectParameters("Gaussian blur");
```
eliezerpMS, line 363: >Gaussian [](http://example.com/codeflow?start=62&length=8)

nit: lower case the G for consistency (I know it was capitalized before....)

source/LottieToWinComp/Shapes.cs
```diff
@@ -692,6 +691,20 @@ internal ShapeLayerTranslator(ShapeLayerContext context)
...
+
+                var gaussianBlurEffect = _context.Effects.GaussianBlurEffect;
+
+                if (gaussianBlurEffect is not null)
+                {
+                    rootNode = Effects.ApplyGaussianBlur(_context, rootNode, gaussianBlurEffect);
+                }
+
                 return layerHasMasks
                     ? Masks.TranslateAndApplyMasksForLayer(_context, rootNode)
```
eliezerpMS, line 709: >TranslateAndApplyMasksForLayer [](http://example.com/codeflow?start=28&length=30)

What would thhe workflow in AE be if I want to:
1) Make a bunch of shapes and then blur and/or shadow then mask the results versus 
2) Make a bunch of shapes and then mask it to make a new cool shape and then blue and/or shadow the results

I assume this arrangement below represents the first scenario and the second scenario would have an extra layer with the effects?

aborziak-ms: Hm, yeah, this order is wrong. It seems that if any layer has Effects and Masks then we should apply masks first (same as for PreComps). And it seems that there is no way in After Effects to apply Effects before Masks (unless they are on different layers)

source/LottieToWinComp/Effects.cs
aborziak-ms: done

eliezerpMS (approved):
:shipit:

Issue #431, opened 2021-03-06
LottieViewer stage and color preview should show transparency with checkerboard pattern


simeoncran:
Currently there's no way to see what is transparent on a Lottie except by changing the background color and observing what things change. This has confused people in the past (e.g. showing a Lottie with a transparent background and white shapes looks like the whole thing is white until you change the background). And it's a problem if we're going to allow alpha values to be specified in the color picker.

For an example of the checkerboard that people will expect, see an empty After Effects canvas.

PR #456, opened 2021-08-05
Lottie Viewer | Added scale slider. Added repeating pattern background and checkbox for hiding/showing solid color bg.

aborziak-ms:
Added canvas size slider right under play speed slider. It changes the maximum height of the player. We had this as a feture request.

Also now by default there is a checkerboard pattern background (like in most graphics editors) #431 

Preview:


LottieViewer/Stage.xaml.cs
```diff
@@ -32,6 +41,51 @@ public Stage()
             this.InitializeComponent();
 
             Reset();
+
+            _ = SetupBackgroundPatternAsync();
+
+            ShowSolidBackground = false;
```
raymond-wh-leung, line 47: > ShowSolidBackground = false; [](http://example.com/codeflow?start=11&length=29)

Should this be in the Reset() method?

raymond-wh-leung (approved):
:shipit:

LottieViewer/Stage.xaml.cs
aborziak-ms: Hm, I'm not sure what `Reset` is intended for initially. Right now it is used only in the constructor. My idea was that we can use `Reset` to reset the background and show default LottieViewer screen (the one that you see before I drop the animation file, with "json" word in the middle). But again, since we are not using Reset anywhere else, it is probably even better just to move this to the constructor.

raymond-wh-leung: Gotcha

raymond-wh-leung (approved):
:shipit:

LottieViewer/Stage.xaml.cs
```diff
@@ -31,7 +40,56 @@ public Stage()
...
+
+            // Do not show checkerboard pattern until lottie file is loaded.
+            _canvas.Visibility = Visibility.Collapsed;
+
+            _feedbackLottie.PlayInitialStateAnimation();
+
+            _ = SetupBackgroundPatternAsync();
+        }
+
+        // Draw repeating paatern texture on backgound canvas.
```
eliezerpMS, line 54: >paatern [](http://example.com/codeflow?start=26&length=7)

pattern

LottieViewer/Stage.xaml.cs
```diff
@@ -31,7 +40,56 @@ public Stage()
...
+        public async Task SetupBackgroundPatternAsync()
+        {
+            var compositor = ElementCompositionPreview.GetElementVisual(this).Compositor;
+            var canvasDevice = CanvasDevice.GetSharedDevice();
+            var graphicsDevice = CanvasComposition.CreateCompositionGraphicsDevice(compositor, canvasDevice);
+
+            var bitmap = await CanvasBitmap.LoadAsync(canvasDevice, @"Assets\BackgroundPattern.png");
```
eliezerpMS, line 61: >CanvasBitmap [](http://example.com/codeflow?start=31&length=12)

Just curious, why not just add this in the xaml?

LottieViewer/MainPage.xaml.cs
```diff
@@ -51,6 +51,10 @@ public MainPage()
             // Add the background to the color picker so that it can be modified by the user.
             _paletteColorPicker.PaletteEntries.Add(BackgroundColor);
 
+            // Update background visibility after checkbox clicked.
+            _paletteColorPicker.ShowSolidBackground.Click +=
```
eliezerpMS, line 55: >ShowSolidBackground.Click += [](http://example.com/codeflow?start=32&length=28)

Can we set this up in the Xaml instead?

LottieViewer/MainPage.xaml
```diff
@@ -184,6 +184,40 @@
...
+                                    SmallChange="1"
+                                    StepFrequency="1"
+                                    TickFrequency="16"
+                                    TickPlacement="TopLeft"
+                                    Value="{x:Bind _stage.PlayerContainer.MaxHeight, Mode=TwoWay}">
+                                    <Slider.Resources>
+                                        <!-- 
+                                            Override the themed slider resources so that control ignores the accent colors.
+                                            This is desirable because the accent colors may make the control hard to see
+                                            on the gray background.
```
eliezerpMS, line 210: Yeah! I like this change! :smile:

LottieViewer/Stage.xaml.cs
```diff
@@ -56,6 +129,7 @@ internal async Task<bool> TryLoadFileAsync(StorageFile file)
             {
                 // Load the Lottie composition.
                 await _playerSource.SetSourceAsync(file);
+                _canvas.Visibility = Visibility.Visible;
```
eliezerpMS, line 132: How do you know that SetupBackgroundPatternAsync completed before we get here?

aborziak-ms: Hm, I think I don't need to, since canvas is always present, and here I'm just changing it's property.

LottieViewer/MainPage.xaml.cs
aborziak-ms: hm, not sure how to do that easily, since it basically connects `_paletteColorPicker` and `_stage` and they do not know about each other

LottieViewer/Stage.xaml.cs
aborziak-ms: You mean this entire function?
I'm not sure if the approach I'm using here is the best, it is the one that I found. If there are simpler ways to make this pattern tiling texture in xaml then I can change this

eliezerpMS: Yeah it looks like you are right there is no built in tiling functionality. It looks like we have one in the toolkit [here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.media.tilesbrush?view=win-comm-toolkit-dotnet-7.0) although that might be overkill to pull that in just for this...

LottieViewer/Stage.xaml.cs
eliezerpMS: If we get here before the image loads what does the user see?

LottieViewer/MainPage.xaml.cs
eliezerpMS: True. We should definitiely keep things simpler for this app.

LottieViewer/Stage.xaml.cs
aborziak-ms: hm, I assume just an empty canvas, but since it is transparent -> background color

eliezerpMS: I guess what I am saying is that if the intent is to show the checker background when the Lottie has loaded (and not have it pop in later, which technically it probably wont do) how about awaiting the result of SetupBackgroundPatternAsync before doing "_canvas.Visibility = Visibility.Visible;"?

aborziak-ms: Ok done. 

Unfortunately it caused VSTHRD003:Avoid awaiting foreign Tasks and to resolve it asks to use JoinableTask from Microsoft.VisualStudio.Threading (that is not available for us), so I just suppressed it for this method. Is that ok?

eliezerpMS: To keep things simple how do you feel about just doing "await SetupBackgroundPatternAsync()" before the "_canvas.Visibility = Visibility.Visible;"? We could get fancy with multiple tasks and joining but I think that is overkill since we know that SetupBackgroundPatternAsync should be pretty quick. Then we can also get rid of having  a hanging task in the constructor.

aborziak-ms: Ok, agree. Done

eliezerpMS (approved):
:shipit:

PR #458, opened 2021-08-12
Added SolidColorStroke merge optimization

aborziak-ms:
Added merge optimization for SolidColorStroke. It is needed to optimize some animations that use strokes


raymond-wh-leung (approved):
:shipit:

PR #459, opened 2021-08-13
Translate shape as a separate visual with child if it has animated opacity

aborziak-ms:
Previously if shape layer had animated or non-100% opacity it generated color expression animation (`"ColorRGB(my.Opacity, r, g, b)"`). Instead of that we can create separate visual with one child - this shape layer, and then animate opacity of the visual. We already have this logic for masks and effects so I just added another condition.

### DIFF
--- a/source/LottieToWinComp/Shapes.cs
+++ b/source/LottieToWinComp/Shapes.cs
@@ -646,8 +646,8 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             // Otherwise we should use extra parent Visual (GetVisualRoot).
             //
             // Note: We can apply Masks and effects only to Visual node.
-            // Also we can reduce number of expression animations if we apply opacity dorectly to visual node
-            // instad of pushing it down to color fill that will produce color expression animation.
+            // Also we can reduce number of expression animations if we apply opacity directly to visual node
+            // instead of pushing it down to color fill that will produce color expression animation.
             internal override bool IsShape =>
                 !_context.Layer.Masks.Any() &&
                 _context.Effects.DropShadowEffect is null &&
--- a/source/LottieToWinComp/Shapes.cs
+++ b/source/LottieToWinComp/Shapes.cs
@@ -644,6 +644,10 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
             // Indicates if we can translate shape layer as a single composition shape (GetShapeRoot).
             // Otherwise we should use extra parent Visual (GetVisualRoot).
+            //
+            // Note: We can apply Masks and effects only to Visual node.
+            // Also we can reduce number of expression animations if we apply opacity dorectly to visual node
+            // instad of pushing it down to color fill that will produce color expression animation.
             internal override bool IsShape =>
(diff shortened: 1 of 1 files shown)


aborziak-ms:
I recorded some traces comparing before/after the change and it showed me some perf improvements. Not too much but at least we will have less expression animations

raymond-wh-leung:
Nice :) 

---
In reply to: [898618231](https://github.com/CommunityToolkit/Lottie-Windows/pull/459#issuecomment-898618231) [](http://example.com/codeflow?ancestors=898618231,898618231)

raymond-wh-leung (approved):
:shipit:

source/LottieToWinComp/Shapes.cs
```diff
@@ -642,10 +642,13 @@ internal ShapeLayerTranslator(ShapeLayerContext context)
                 _context = context;
             }
 
+            // Indicates if we can translate shape layer as a single composition shape (GetShapeRoot).
+            // Otherwise we should use extra parent Visual (GetVisualRoot).
```
eliezerpMS, line 646: >Otherwise we should use extra parent Visual (GetVisualRoot). [](http://example.com/codeflow?start=15&length=60)

Nit: would you mind mentioning that the opacity check is to reduce expression animations (and feel free to put more comments about the other checks here  maybe?)

eliezerpMS (approved):
:shipit:

source/LottieToWinComp/Shapes.cs
aborziak-ms: Ok

eliezerpMS: Thanks!

source/LottieToWinComp/Shapes.cs
```diff
@@ -642,10 +642,17 @@ internal ShapeLayerTranslator(ShapeLayerContext context)
...
+            //
+            // Note: We can apply Masks and effects only to Visual node.
+            // Also we can reduce number of expression animations if we apply opacity dorectly to visual node
```
eliezerpMS, line 649: >dorectly [](http://example.com/codeflow?start=86&length=8)

directly

source/LottieToWinComp/Shapes.cs
```diff
@@ -642,10 +642,17 @@ internal ShapeLayerTranslator(ShapeLayerContext context)
...
+            // instad of pushing it down to color fill that will produce color expression animation.
```
eliezerpMS, line 650: >instad [](http://example.com/codeflow?start=15&length=6)

instead

source/LottieToWinComp/Shapes.cs
eliezerpMS, line 649: >node [](http://example.com/codeflow?start=105&length=4)

nodes?

eliezerpMS (dismissed):
:shipit:

source/LottieToWinComp/Shapes.cs
aborziak-ms: omg, so many typos, sorry :) Done

eliezerpMS (approved):
:shipit:

PR #460, opened 2021-08-13
Added more merge optimizer options and moved optimizer call for LottieViewer

aborziak-ms:
- I added more merge helper methods (They are used in some animations). 
- Removed some checks since it seems that they are no longer needed. Now we can optimize more cases.
  - We should be able to merge nested LayerPreComp (Previously I noticed a bug with that so I added this check but I can no longer reproduce it). 
  - I don't know why I added `aAsset == bAsset` check since I'm making a copy of it anyway a line below (`WithTimeOffset` creates a copy).

- Turned out that previously I was calling optimizer in wrong place. It was too deep so LottieGen called it twice and even with command line option -DisableLottieMergeOptimizer. So I moved it higher in the stack.

### DIFF
--- a/source/LottieData/Optimization/Experimental/MergeHelper.cs
+++ b/source/LottieData/Optimization/Experimental/MergeHelper.cs
@@ -374,6 +374,82 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieData.Optimization
             return new MergeResult<Path>(new Path(args, a.DrawingDirection, geometryData.Value!));
         }
 
+        MergeResult<Rectangle> MergeRectangles(Rectangle a, TimeRange aRange, Rectangle b, TimeRange bRange)
+        {
+            if (a.BlendMode != b.BlendMode || a.DrawingDirection != b.DrawingDirection)
+            {
+                return MergeResult<Rectangle>.Failed;
+            }
+
+            var args = new ShapeLayerContentArgs
+            {
+                Name = $"{a.Name} {b.Name}",
+                MatchName = $"{a.MatchName}{b.MatchName}",
+                BlendMode = a.BlendMode,
+            };
+
+            var position = MergeIAnimatableVector3(a.Position, aRange, b.Position, bRange);
+            var size = MergeIAnimatableVector3(a.Size, aRange, b.Size, bRange);
+            var roundness = MergeAnimatable(a.Roundness, aRange, b.Roundness, bRange);
+
+            if (!position.Success || !size.Success || !roundness.Success)
+            {
+                return MergeResult<Rectangle>.Failed;
+            }
+
+            return new MergeResult<Rectangle>(new Rectangle(args, a.DrawingDirection, position.Value!, size.Value!, roundness.Value!));
+        }
+
+        MergeResult<TrimPath> MergeTrimPaths(TrimPath a, TimeRange aRange, TrimPath b, TimeRange bRange)
+        {
+            if (a.BlendMode != b.BlendMode || a.TrimPathType != b.TrimPathType)
+            {
+                return MergeResult<TrimPath>.Failed;
+            }
+
+            var args = new ShapeLayerContentArgs
+            {
+                Name = $"{a.Name} {b.Name}",
+                MatchName = $"{a.MatchName}{b.MatchName}",
+                BlendMode = a.BlendMode,
+            };
+
+            var start = MergeAnimatable(a.Start, aRange, b.Start, bRange);
+            var end = MergeAnimatable(a.End, aRange, b.End, bRange);
+            var offset = MergeAnimatable(a.Offset, aRange, b.Offset, bRange);
+
+            if (!start.Success || !end.Success || !offset.Success)
+            {
+                return MergeResult<TrimPath>.Failed;
+            }
+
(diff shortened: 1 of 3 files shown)


source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -836,7 +908,7 @@ MergeResult<PreCompLayer> MergePreCompLayers(PreCompLayer a, PreCompLayer b)
 
             var bAsset = GetAssetById(b.RefId);
 
-            if (aAsset is not LayerCollectionAsset || bAsset is not LayerCollectionAsset || aAsset == bAsset)
```
raymond-wh-leung, line 839: >aAsset == bAsset [](http://example.com/codeflow?start=92&length=16)

Curious what it means with this check removed.

source/LottieReader/Serialization/LottieCompositionReader.cs
```diff
@@ -288,8 +288,6 @@ LottieComposition ParseLottieComposition(ref Reader reader)
                                                     layers: new LayerCollection(layers),
                                                     markers: markers);
 
-                                result = LottieMergeOptimizer.Optimize(result);
```
eliezerpMS, line 291: > result = LottieMergeOptimizer.Optimize(result); [](http://example.com/codeflow?start=31&length=48)

Just curious, you were seeing us hit this twice for 1 lottie?

source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -374,6 +374,82 @@ MergeResult<Path> MergePaths(Path a, TimeRange aRange, Path b, TimeRange bRange)
             return new MergeResult<Path>(new Path(args, a.DrawingDirection, geometryData.Value!));
         }
 
+        MergeResult<Rectangle> MergeRectangles(Rectangle a, TimeRange aRange, Rectangle b, TimeRange bRange)
+        {
+            if (a.BlendMode != b.BlendMode || a.DrawingDirection != b.DrawingDirection)
```
eliezerpMS, line 379: >a.BlendMode != b.BlendMode  [](http://example.com/codeflow?start=16&length=27)

Looking at it now, instead of having this check repeated in a lot of methods what would it look like to have a "CanMerge" in one of the parent classes and then each class can build on its parent definition?

source/LottieData/Optimization/Experimental/MergeHelper.cs
```diff
@@ -512,6 +588,12 @@ MergeResult<ShapeLayerContent> MergeShapeLayerContents(ShapeLayerContent a, Time
                     return MergeResult<ShapeLayerContent>.From(MergeShapeGroup((ShapeGroup)a, aRange, (ShapeGroup)b, bRange));
                 case ShapeContentType.Path:
                     return MergeResult<ShapeLayerContent>.From(MergePaths((Path)a, aRange, (Path)b, bRange));
+                case ShapeContentType.TrimPath:
+                    return MergeResult<ShapeLayerContent>.From(MergeTrimPaths((TrimPath)a, aRange, (TrimPath)b, bRange));
+                case ShapeContentType.Rectangle:
+                    return MergeResult<ShapeLayerContent>.From(MergeRectangles((Rectangle)a, aRange, (Rectangle)b, bRange));
+                case ShapeContentType.RoundCorners:
+                    return MergeResult<ShapeLayerContent>.From(MergeRoundCorners((RoundCorners)a, aRange, (RoundCorners)b, bRange));
                 case ShapeContentType.Ellipse:
```
eliezerpMS, line 597: super nit: mind alphabetizing this switch statement?

source/LottieData/Optimization/Experimental/MergeHelper.cs
aborziak-ms: Animations usually contains many consecutive PreCompLayers and each of them are pointing to "asset" (content). And for some PreCompLayers they are the same (e.g. same animation for different states). So I just enabled an option to merge two equal assets into one bigger (e.g. A and A into AA)

We can probably optimize this case event better, and leave only one PreCompLayer if we have multiple PreComps that are pointing to the same asset.

source/LottieReader/Serialization/LottieCompositionReader.cs
aborziak-ms: We have this call in two places. One intended for LottieGen and can be disabled in command line, and one for LottieViewer. And turned out that we hit both of them if we use LottieGen (because this method is used by LottieGen too). So I moved this one up in stack to be called only for LottieViewer

source/LottieData/Optimization/Experimental/MergeHelper.cs
aborziak-ms: Agree, I have some ideas on how to refactor everything to even get rid of MergeHelper class. Would you mind if I will do all the refactoring in the following PR? I already started working on it

source/LottieData/Optimization/Experimental/MergeHelper.cs
aborziak-ms: Done

source/LottieData/Optimization/Experimental/MergeHelper.cs
eliezerpMS: Yep I am fine with that.

eliezerpMS (approved):
:shipit:

Issue #464, opened 2021-08-20
Generated CSharp code is not compatible with downlevel language versions


aschultz:
The code generated by LottieGen in `csharp` mode uses language features like `is not null` that are only available in C# 9.0 and above. UWP applications are still stuck on older versions of C# (7.3 I think?), so the generated code is unusable. I'm currently running into this while developing Windows support in lottie-react-native.

Please replace these constructs with more compatible ones. We don't need syntactic sugar in generated code if it's going to cause compatibility problems. The core language was plenty capable even back to C# 3 or 4.

PR #465, opened 2021-08-20
Replace 'is null' with more compatible clause and fix MSBuild bug

aschultz:
- Replace `is null` usages in generated code with downlevel equivalent to support older versions of C# in client applications
- Fix MSBuild task to send bool args correctly. The LottieGen cmdline only allows flags to be set without an associated value arg.


aschultz:
#464

aborziak-ms (approved):

PR #467, opened 2021-08-23
[LottieGen] Graph compactor fix. Do not coalesce root container visuals.

aborziak-ms:
It seems that we should not coalesce root container visual with it's child. Sometimes users of Lottie are changing height/width/scale properties of the root visual and in combination with non-null ClipRect of child it may result in wrong clipping being applied.

This started appearing after new optimization was added because new optimization often merges together immediate children of the root node, so that GraphCompactor then tries to coalesce root node and it's new single child.

### DIFF
--- a/source/UIData/Tools/GraphCompactor.cs
+++ b/source/UIData/Tools/GraphCompactor.cs
@@ -1143,12 +1143,16 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.UIData.Tools
         // affects different sets of properties then they can be combined into one.
         void CoalesceOrthogonalContainerVisuals(ObjectGraph<Node> graph)
         {
-            // If a container is not animated and has no properties set, its children can be inserted into its parent.
+            // If a container is not animated and has no properties set, its children can be inserted into its parent,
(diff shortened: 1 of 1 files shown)


source/UIData/Tools/GraphCompactor.cs
```diff
@@ -1143,12 +1143,16 @@ void CoalesceContainerVisuals(ObjectGraph<Node> graph)
         // affects different sets of properties then they can be combined into one.
         void CoalesceOrthogonalContainerVisuals(ObjectGraph<Node> graph)
         {
-            // If a container is not animated and has no properties set, its children can be inserted into its parent.
+            // If a container is not animated and has no properties set, its children can be inserted into its parent,
+            // except the root node. We should not change the root node since the user can try to change width/height/scale
+            // of the root node and in combination with non-null Clip property (or others)
```
raymond-wh-leung, line 1148: nit: odd spacing at the end compared to the previous line.

raymond-wh-leung (approved):
:shipit:

eliezerpMS (approved):
:shipit:

PR #472, opened 2021-09-17
Moved files from Optimization/Experimental folder to Optimization folder

aborziak-ms:
Since now this optimization is enabled by default, and we made enough tests to be sure that it works in all scenarios (Few weeks ago I ran a script that compared 500+ different lotties before and after the optimization, and it didn't find any difference) let's move this out of experimental folder for convenience.


eliezerpMS (approved):
:shipit:

raymond-wh-leung (approved):
:shipit:

PR #473, opened 2021-10-04
[Issues] Updated LT0030 issue to show that LottieGen actually supports linked image files

aborziak-ms:
Updated LT0030 issue text, since LottieGen actually generates the code in case of linked image file, and it correctly shows this image in Lottie animation. 

P.S. In future we should add support of this to LottieViewer too, but it requires us to enable loading zip files, not only jsons.


raymond-wh-leung (approved):
:shipit:

PR #477, opened 2021-11-11
[LottieGen] Added IAnimatedVisual2 implementation preview.

aborziak-ms:
Added implementation of IAnimatedVisual2 interface that is going to be added to xaml controls: https://github.com/microsoft/microsoft-ui-xaml/pull/6268. It is disabled by a boolean flag for now, later we will decide if we want to enable it by default (with version check) or to add a command line argument.

```
interface IAnimatedVisual2 {
    void InstantiateAnimations(); 
    void DestroyAnimations();
}
```

We also have to ensure that all the animations have zero keyframes (progress = 0.0) in order to correctly re-instantiate animation states, previously we were discarding some zero keyframes.

### DIFF
--- a/source/LottieToWinComp/Transforms.cs
+++ b/source/LottieToWinComp/Transforms.cs
@@ -4,6 +4,7 @@
 
 #nullable enable
 
+using System;
 using System.Diagnostics.CodeAnalysis;
 using Microsoft.Toolkit.Uwp.UI.Lottie.Animatables;
 using Microsoft.Toolkit.Uwp.UI.Lottie.LottieData;
@@ -228,7 +229,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                     visibilityAnimation.InsertKeyFrame(0.0f, Sn.Vector2.Zero, context.ObjectFactory.CreateHoldThenStepEasingFunction());
                 }
 
-                visibilityAnimation.InsertKeyFrame(inProgress, Sn.Vector2.One, context.ObjectFactory.CreateHoldThenStepEasingFunction());
+                visibilityAnimation.InsertKeyFrame(Math.Max(0.0f, inProgress), Sn.Vector2.One, context.ObjectFactory.CreateHoldThenStepEasingFunction());
 
                 if (outProgress < 1)
                 {
@@ -341,7 +342,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
                     visibilityAnimation.InsertKeyFrame(0.0f, false);
                 }
 
-                visibilityAnimation.InsertKeyFrame(inProgress, true);
+                visibilityAnimation.InsertKeyFrame(Math.Max(0.0f, inProgress), true);
 
                 if (outProgress < 1)
                 {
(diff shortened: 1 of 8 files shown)


source/UIDataCodeGen/CodeGen/Cx/CxInstantiatorGenerator.cs
```diff
@@ -1003,7 +1021,13 @@ static string TypeName(PropertySetValueType propertySetValueType)
                     builder.WriteLine();
                     builder.WriteLine($"if ({info.ClassName}::IsRuntimeCompatible())");
                     builder.OpenScope();
-                    builder.WriteBreakableLine($"return {_s.New(info.ClassName)}(", CommaSeparate(GetConstructorArguments(info)), ");");
+                    builder.WriteBreakableLine($"auto result = {_s.New(info.ClassName)}(", CommaSeparate(GetConstructorArguments(info)), ");");
+                    if (_implementIAnimatedVisual2)
+                    {
+                        builder.WriteLine($"result.{InstantiateAnimationsMethod}(0.0f);");
```
raymond-wh-leung, line 1027: >"result.{InstantiateAnimationsMethod}(0.0f) [](http://example.com/codeflow?start=43&length=43)

is it true that a non-zero progress will only be used inside the AnimatedVisualPlayer?

source/LottieToWinComp/Transforms.cs
```diff
@@ -336,16 +338,18 @@ public static Sn.Matrix3x2 CreateMatrixFromTransform(LayerContext context, Trans
                 {
                     // Set initial value to be non-visible.
                     visibilityNode.IsVisible = false;
-                    visibilityAnimation.InsertKeyFrame(inProgress, true);
+                    visibilityAnimation.InsertKeyFrame(0.0f, false);
```
raymond-wh-leung, line 341: >      visibilityAnimation.InsertKeyFrame(0.0f, false); [](http://example.com/codeflow?start=14&length=54)

fyi that if a keyframe at 0 is not inserted, under the cover dcomp will insert one at time 0 with the value as "this.StartingValue" which essentialy means the value the proprety is at when the animation is started.

raymond-wh-leung (dismissed):
:shipit:

source/LottieToWinComp/Transforms.cs
aborziak-ms: hm, but what will happen if for example I have a scalar animation:
kf1: progress 0.5, value 1.0f; kf2: progress: 0.8 value 2.0f; (linear easing)
and then we stop this animation in the middle on progress value 0.65. The property will have value 1.5f. What will happen when we instantiate this animation again? Will the initial value be equal to 1.5f?

raymond-wh-leung: Yea for that example, say the property value is at 1.5f when the animation is stopped. The next time if you instantiate an animation again (assuming no value is provided at keyframe 0), then the value at keyframe 0 will be 1.5f. If a specific value is needed at keyframe 0 yes you'll need to set that value explicilty.

aborziak-ms: ok I see, yeah, that what I was thinking. This behavior resulted in some visual differences from to run, so I added zero keyframes explicitly

raymond-wh-leung: Gotcha

raymond-wh-leung (approved):
:shipit:

PR #479, opened 2021-12-15
Fixed ComposeMatteLayers method to work for invisible layers

aborziak-ms:
`translatedLayer.GetVisualRoot` may return `null` if it detects that layer will not be visible (it has opacity zero, or visible time range is out of parent's time range). If this layer should be matted or should be used as a matte layer it may cause some artifacts to appear.

New implementation should be more robust and should handle most corner cases.

1. If both `mattedLayer` and `matte` are not null -> compose them into one layer
2. If `mattedLayer` is `null` -> skip both layers
3. If `matte` layers is `null` -> we can skip them too, (!) except for the case when `MatteType` is `Invert`, in this case we are inverting invisible layer which mean that it is always visible, so we should keep `mattedLayer`

### DIFF
--- a/source/LottieToWinComp/Masks.cs
+++ b/source/LottieToWinComp/Masks.cs
@@ -138,61 +138,51 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
 
         // Walk the collection of layer data and for each pair of matte layer and matted layer, compose them and return a visual
         // with the composed result. All other items are not touched.
-        public static IEnumerable<LayerTranslator> ComposeMattedLayers(CompositionContext context, IEnumerable<(LayerTranslator translatedLayer, Layer layer)> items)
+        public static IEnumerable<LayerTranslator> ComposeMattedLayers(CompositionContext context, IList<(LayerTranslator translatedLayer, Layer layer)> items)
         {
-            // Save off the visual for the layer to be matted when we encounter it. The very next
-            // layer is the matte layer.
-            Visual? mattedVisual = null;
-            Layer.MatteType matteType = Layer.MatteType.None;
-
+            // After Effects has a concept named "matte layers" which is very similar to masks.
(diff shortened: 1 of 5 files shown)


source/LottieToWinComp/Masks.cs
```diff
@@ -138,61 +138,51 @@ static class Masks
...
+                        {
+                            // Both matte and mattedLayer are not null -> both are visible, compose them into one visual.
+                            var compositedMatteVisual = Masks.TranslateMatteLayer(context, matte, mattedVisual, layer.LayerMatteType == Layer.MatteType.Invert);
+                            yield return compositedMatteVisual;
+                        }
+                        else if (mattedVisual is not null && matte is null && layer.LayerMatteType == Layer.MatteType.Invert)
+                        {
+                            // Matte layer is null, which means that it is never visible and LayerMatteType is equal to Invert.
+                            // In this case we should just return mattedVisual because matte layer is effectively a no-op.
+                            yield return new LayerTranslator.FromVisual(mattedVisual);
```
eliezerpMS, line 177: > yield return new LayerTranslator.FromVisual(mattedVisual); [](http://example.com/codeflow?start=27&length=59)

Is this what happens in After Effects? (As in if you make the matte inverted and not visible then it just shows the layer to be matted?)

Also should this produce an issue or something? Is there a case where a created would have done this intentionally?

aborziak-ms: Yes this is how it works in AE.

Actually this is a good idea I think, because that is how we discovered this bug, Lottie-Windows produced wrong result but there were no issues to see what exactly is wrong.

eliezerpMS: Cool. Thanks for adding the new issue!

I am thinking though that we probably want the issue to pop up even for the invert no-op case too right?

aborziak-ms: Agree, thx

eliezerpMS (approved):
:shipit:

PR #480, opened 2021-12-15
[LottieGen] Fixed UAP version check for Drop Shadow effect

aborziak-ms:
We need this version check for DropShadow effect since it can't be supported for UAP <=7 , so just fall back to no shadow in this case.


eliezerpMS (approved):
:shipit:

PR #483, opened 2022-01-27
[LottieGen] Fixed layer combining for Exclude mode

### LINKED ISSUE #482: lottiegen warning message
warning LP0017: Unexpected field: l.
warning LP0017: Unexpected field: ReadAnimatableVector3.l.

aborziak-ms:
It seems that AfterEffects is applying layer combination mode in reverse order. It does not affect Intersect/Union/Xor but it affects how Exclude mode works, so we need to reverse the order. `Exclude[layer1, layer2]` should be interpreted as `(layer2 - layer1)`

Note: We should do more investigation how it works  for more than 3 layers, but currently I do not have an example of animation that uses 3 layers or more with Exclude mode, but even then it should be pretty rare case. `Exclude[layer1, layer2, layer3]` may be interpreted as `(layer3 - (layer2 - layer1))` or `((layer3 - layer2) - layer1)`.

Should fix #482

### DIFF
--- a/source/LottieToWinComp/Shapes.cs
+++ b/source/LottieToWinComp/Shapes.cs
@@ -313,9 +313,20 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             return CanvasGeometryCombiner.CombineGeometries(geometries, combineMode);
 #else
                 var accumulator = geometries[0];
-                for (var i = 1; i < geometries.Length; i++)
+                if (combineMode == CanvasGeometryCombine.Exclude)
                 {
-                    accumulator = accumulator.CombineWith(geometries[i], Sn.Matrix3x2.Identity, combineMode);
+                    // TODO: investiagte how it works for 3+ layers with Exclude mode
+                    for (var i = 1; i < geometries.Length; i++)
+                    {
+                        accumulator = geometries[i].CombineWith(accumulator, Sn.Matrix3x2.Identity, combineMode);
+                    }
+                }
+                else
+                {
+                    for (var i = 1; i < geometries.Length; i++)
+                    {
+                        accumulator = accumulator.CombineWith(geometries[i], Sn.Matrix3x2.Identity, combineMode);
+                    }
                 }
 
                 return accumulator;


source/LottieToWinComp/Shapes.cs
```diff
@@ -313,9 +313,20 @@ static CompositionShape TranslateGroupShapeContent(ShapeContext context, ShapeGr
...
 #else
                 var accumulator = geometries[0];
-                for (var i = 1; i < geometries.Length; i++)
+                if (combineMode == CanvasGeometryCombine.Exclude)
                 {
-                    accumulator = accumulator.CombineWith(geometries[i], Sn.Matrix3x2.Identity, combineMode);
+                    // TODO: investiagte how it works for 3+ layers with Exclude mode
+                    for (var i = 1; i < geometries.Length; i++)
+                    {
+                        accumulator = geometries[i].CombineWith(accumulator, Sn.Matrix3x2.Identity, combineMode);
```
raymond-wh-leung, line 321: >accumulator [](http://example.com/codeflow?start=64&length=11)

Wondering if it should be reversing the for loop (i.e. for (var i = geometries.length-1; i >=0; i--) instead of using the accumulator here?

aborziak-ms: The difference is how do we interpret the "Exclude" operation for multiple layers, it can be `(layer3 - (layer2 - layer1))`, or it can be `((layer3 - layer2) - layer1)`. This code interprets it as `(layer3 - (layer2 - layer1))`, if I will do reverse loop it will be  `((layer3 - layer2) - layer1)`.

I'm trying to find the doc, or an example of it

raymond-wh-leung: Gotcha.

raymond-wh-leung (approved):
:shipit:

DominicMaas:
Is there a way I can consume a build of LottieGen using the build generated by this PR (or any pre-release nugets)? Encountered the same issue with an animation which has caused to it behave incorrectly, wanting to see if this will fix it.

aborziak-ms:
@DominicMaas LottiGen 7.1.1 should become available with the fix above. 

What kind of warning do you get?
It would be great if you can create an issue with the description of the bug, thanks.

PR #488, opened 2022-02-16
[LottieGen] Added new AnimationController API support for LottieGen

aborziak-ms:
Instead of binding Progress to each individual `AnimationController` in this change we are creating one `AnimationController` for entire animation and then using it for all animations.

Note: Lottie representation of `AnimationController` needs to have boolen flag `"IsCustom"` which displays how the `AnimationController` was created - as a custom one (using `Compositor::CreateAnimationController`), or was obtained from an object with `TryGetAnimationController`. Only custom controller can be used in `StartAnimation(property, animation, controller)` calls to make it easier to track the lifetime of `AnimationController` objects and to make it easier to generate the code.

### DIFF
--- a/source/LottieToWinComp/CompositionObjectFactory.cs
+++ b/source/LottieToWinComp/CompositionObjectFactory.cs
@@ -86,6 +86,9 @@ namespace CommunityToolkit.WinUI.Lottie.LottieToWinComp
                 case nameof(PathKeyFrameAnimation):
                     return 11;
 
+                case nameof(AnimationController):
+                    return 14;
+
                 default:
                     throw new InvalidOperationException();
             }
@@ -302,6 +305,12 @@ namespace CommunityToolkit.WinUI.Lottie.LottieToWinComp
                 $"UAP version {uapVersion} features are not available.");
 
             HighestUapVersionUsed = Math.Max(HighestUapVersionUsed, uapVersion);
+        }
+
+        internal AnimationController CreateAnimationControllerList()
+        {
+            ConsumeVersionFeature(GetUapVersionForApi(nameof(AnimationController)));
+            return _compositor.CreateAnimationController();
         }
     }
 }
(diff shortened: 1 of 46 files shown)


source/LottieToWinComp/CompositionObjectFactory.cs
```diff
@@ -86,6 +86,9 @@ internal uint GetUapVersionForApi(string apiName)
                 case nameof(PathKeyFrameAnimation):
                     return 11;
 
+                case nameof(AnimationController):
```
raymond-wh-leung, line 89: >case nameof(AnimationController): [](http://example.com/codeflow?start=16&length=33)

nit: how about add a comment that, this class was introduced in version 6 (to control one keyframeAnimator); the support for it to control multiple keyframeAnimators was added in version 14.

raymond-wh-leung (approved):
:shipit:

raymond-wh-leung (approved):
:shipit:

raymond-wh-leung (dismissed):
:shipit:

raymond-wh-leung:
btw, saw the commit description "API UAP 14", for SV2 the UAP version should be 15.

aborziak-ms:
@raymond-wh-leung oh right, I will edit it. 
Trying to fix the build now

raymond-wh-leung:
I see gotcha!

---
In reply to: [1139124839](https://github.com/CommunityToolkit/Lottie-Windows/pull/488#issuecomment-1139124839) [](http://example.com/codeflow?ancestors=1139124839,1139124839)

raymond-wh-leung:
>         <TargetPlatformVersion>10.0.18362.0</TargetPlatformVersion>

Is it by design that it is reverted from version 10.0.22558.0 from iteration #3

---
Refers to: Directory.Build.props:63 in d8a80eb. [](commit_id = d8a80eb4950bdcdeab6eb29606be078b93d840ed, deletion_comment = False)

PR #489, opened 2022-02-16
ShapeLayerTranslator fix

aborziak-ms:
Reverting the #459, it can result in wrong translation in some rare cases. Will need to investigate it more in the future.


raymond-wh-leung (approved):
:shipit:

PR #492, opened 2022-02-23
Fixed visibility property for visuals with shadow

aborziak-ms:
Fixing translation of shadow effect. If source of the shadow is not visible it does not make shadow disappear, so here I'm pushing `IsVisible` property up so that this property controls visibility of both, source and shadow visuals.

### DIFF
--- a/source/LottieToWinComp/Effects.cs
+++ b/source/LottieToWinComp/Effects.cs
@@ -245,7 +245,7 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             // We need to transfer "IsVisible" property from source to result
             // because result is now the root visual for both shadow and source.
             // Otherwise "IsVisible" property will be applied only to source.
-            MoveIsVisibleProperty(context, source, result);
+            TransferIsVisibleProperty(context, source, result);
 
             // Check if ShadowOnly can be false
             if (!dropShadowEffect.IsShadowOnly.IsAlways(true))
@@ -272,13 +272,13 @@ namespace Microsoft.Toolkit.Uwp.UI.Lottie.LottieToWinComp
             return result;
         }
 
-        static void MoveIsVisibleProperty(TranslationContext context, Visual source, Visual target)
+        static void TransferIsVisibleProperty(TranslationContext context, Visual source, Visual target)
         {
-            // Here we are moving default value of "IsVisible" to target
+            // Here we are transfering default value of "IsVisible" to target
             target.IsVisible = source.IsVisible;
             source.IsVisible = null;
 
-            // Here we are moving "isVisible" animation to target
+            // Here we are transfering "isVisible" animation to target
             foreach (var animator in source.Animators.ToList())
             {
                 if (animator.AnimatedProperty != nameof(source.IsVisible))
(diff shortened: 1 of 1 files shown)


source/LottieToWinComp/Effects.cs
```diff
@@ -267,6 +272,37 @@ void EmitIssueAboutUnsupportedEffect(Effect? effect, string effectName)
...
+                {
+                    Animate.WithKeyFrame(context, target, animator.AnimatedProperty, (KeyFrameAnimation_)animator.Animation);
+                }
+                else if (animator.Animation is ExpressionAnimation)
+                {
+                    Animate.WithExpression(target, (ExpressionAnimation)animator.Animation, animator.AnimatedProperty);
+                }
+                else
+                {
+                    throw new ArgumentException("IsVisible animation should be keyframe of expression");
```
raymond-wh-leung, line 295: >of  [](http://example.com/codeflow?start=88&length=3)

nit: or?

source/LottieToWinComp/Effects.cs
```diff
@@ -267,6 +272,37 @@ void EmitIssueAboutUnsupportedEffect(Effect? effect, string effectName)
...
+                }
+
+                source.StopAnimation(animator.AnimatedProperty);
+            }
+
+            // Here we are moving default value of "IsVisible" to target
+            target.IsVisible = source.IsVisible;
```
raymond-wh-leung, line 302: > target.IsVisible = source.IsVisible; [](http://example.com/codeflow?start=11&length=37)

If the Animate.WithKeyFrame() or WithExpression() calls StartAnimation() that targets IsVisibile, this set here (if the the value gets changed), will stop the animation above, maybe this needs to be set first before the Animate.WithKeyFrame() calls above?

source/LottieToWinComp/Effects.cs
aborziak-ms: fixed

source/LottieToWinComp/Effects.cs
```diff
@@ -242,6 +242,11 @@ void EmitIssueAboutUnsupportedEffect(Effect? effect, string effectName)
             result.Size = size;
             result.Children.Add(blurResult);
 
+            // We need to transfer "IsVisible" property from source to result
+            // because result is now the root visual for both shadow and source.
+            // Otherwise "IsVisible" property will be applied only to source.
+            MoveIsVisibleProperty(context, source, result);
```
raymond-wh-leung, line 248: >MoveIsVisibleProperty [](http://example.com/codeflow?start=12&length=21)

nit: what do you think about TransferIsVisible or PropagateIsVisible?

source/LottieToWinComp/Effects.cs
aborziak-ms: This class is just a "mock class" of a real WinComp class (Lottie-Windows\source\WinCompData\Visual.cs) so it just stores the data in it, after all the translation is done this class is used to instantiate real WinComp instances. So it should be fine, but I reordered it anyway

source/LottieToWinComp/Effects.cs
aborziak-ms: Transfer sounds good, done

source/LottieToWinComp/Effects.cs
raymond-wh-leung: Gotcha.

raymond-wh-leung (approved):
:shipit:

PR #494, opened 2022-03-26
Fixed previous progress with linear cubic bezier

FrayxRulez:
It seems that the previous code was placing a "neutral" keyframe with the previous keyframe value right before the current one instead of right after the previous one, causing the current keyframe to be skipped.


aborziak-ms:
@FrayxRulez do you have an example of animation where this bug reproduces?

aborziak-ms (dismissed):

aborziak-ms (approved):

Issue #493, opened 2022-03-04
Lottie animation plays in Mobile and Web but not for Desktop (UWP)


MarkIvanDev:
This lottie animation plays in Mobile (using LottieFiles in Android) and Web (using LottieFiles.com) but it does not play using the Lottie Viewer in the Microsoft Store.
[Lottie Animation.zip](https://github.com/CommunityToolkit/Lottie-Windows/files/8182417/Lottie.Animation.zip)

It shows this when viewing in Lottie Viewer:
![UWP Lottie Viewer](https://user-images.githubusercontent.com/16188252/156683586-d7284215-abba-41b8-95e4-c32b129447b1.png)

It shows this in LottieFiles.com:
![LottieFiles.com](https://user-images.githubusercontent.com/16188252/156683712-2adbd002-5c60-4a88-934f-fa2b2d1c06c1.png)

It shows this in LottieFiles in Android:
![LottieFiles in Android](https://user-images.githubusercontent.com/16188252/156684037-10590265-30b2-4ede-a19e-828e2621b941.jpg)

It doesn't show also when using the `LottieVisualSource` in UWP projects. And it also does not  show the fallback content when using this lottie file.

I am using the following versions:
Microsoft.Toolkit.Uwp.UI.Lottie v7.1.0
Lottie Viewer v7.1.0

aborziak-ms:
@MarkIvanDev the problem with this specific animation is that it uses a keyframe path animation with a different number of segments between keyframes. Unfortunately current Windows Composition API does not support this, so LottieViewer fails to present the animation on the screen. I will try to find a workaround for this

MarkIvanDev:
Thank you. Is this fix in the current version of NuGet package (v7.1.1)?

PR #495, opened 2022-04-20
[LottieGen] Added workaround to make path animations playable even with different number of segments in keyframes

aborziak-ms:
Some animations can't be played in Lottie-Windows because of WinComp Api limitation: number of segments in path animation should be the same for all keyframes, but After Effects supports animation with any number of segments.

Added a workaround to make number of segments equal for all the keyframes by adding 0-length segments to the path at the end. It may result in a little bit different animation from what you see in After Effects.

#493


raymond-wh-leung (dismissed):
:shipit:

aborziak-ms:
Added warning if this state is detected

raymond-wh-leung (dismissed):
:shipit:

raymond-wh-leung (approved):
:shipit:

raymond-wh-leung (approved):
:shipit:

PR #497, opened 2022-05-03
Fixed gradient brush offset for rectangle geometry

aborziak-ms:
We've already have a fix for this, but turned out that it had some flaws: https://github.com/CommunityToolkit/Lottie-Windows/pull/450

- Current implementation of WinComp API behaves in a different way for Rounded Rectangle geometry and regular Rectangle Geometry. So I added an additional check for this. 
- Previous implementation didn't take into account position of the rectangle.
- Radial gradient behaves in a different way than linear gradient, so it doesn't need the offset at all.


raymond-wh-leung (approved):
:shipit:

PR #500, opened 2022-05-24
Namespace migration and WinUI3 package MicrosoftToolkit.WinUI.Lottie

aborziak-ms:
Added project for WinUI3 package. Since there are a lot of namespaces changes also added preprocessor flag to distinguish WinUI 3 package from UWP package.


source/Lottie/LottieVisualOptions.cs
```diff
@@ -4,7 +4,11 @@
 
 using System;
 
+#if Lottie_Windows_WinUI3
+namespace MicrosoftToolkit.WinUI.Lottie
```
raymond-wh-leung, line 8: >MicrosoftToolkit [](http://example.com/codeflow?start=10&length=16)

Wondering why MicrosoftToollKit instead of Microsoft.TooKit?

aborziak-ms: Oh, sure, it is a new naming convention for **WinUI3** packages in **CommunityToolkit** organization:

"_As part of this release, we’re introducing a new Package Root for the Windows Community Toolkit on NuGet: CommunityToolkit.*. We’ve updated our namespaces and packages from **Microsoft.Toolkit.Uwp.*** to now be **CommunityToolkit.WinUI.***. We’re making this migration now as part of our WinUI 3 update, as having our package identity allows us to better maintain our packages and their ownership with the community. In the future, it also allows us more flexibility to work with other groups and technologies to share branding and ownership of packages._"

https://devblogs.microsoft.com/ifdef-windows/windows-community-toolkit-for-project-reunion-0-5/

raymond-wh-leung: Gotcha, good to know. Thx

raymond-wh-leung (dismissed):
:shipit:

source/Lottie/LottieVisualOptions.cs
michael-hawker: @aborziak-ms yes, but the new name is 'CommunityToolkit.WinUI.*` not `MicrosoftToolkit`. Our plan for 8.0 will be to have both the UWP and WinUI packages use the same common namespace, so in your case `CommunityToolkit.WinUI.Lottie`. If you plan to ship a 7.x series it may make sense to keep the split to be consistent. But since this is a new package, you may want to just keep everything consistent now.

Goal with keeping a single namespace across UWP and WinUI 3 is that it makes it easier to consume in XAML and across any projects that want to multi-target both UWP and WinUI 3 simultaneously.

We haven't tried publishing two packages for both UWP and WinUI out of the same branch before, but since you setup a new project file, then it should be fine. We'll have to see in the DevOps feed first to test.

Lottie-Windows-WinUI3/Lottie-Windows-WinUI3.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<OutputType>Library</OutputType>
+		<RootNamespace>MicrosoftToolkit.WinUI.Lottie</RootNamespace>
```
michael-hawker, line 5:
```suggestion
		<RootNamespace>CommunityToolkit.WinUI.Lottie</RootNamespace>
```

`MicrosoftToolkit` is not our root namespace.

source/Lottie/ImageAssetHandler.cs
```diff
@@ -5,17 +5,25 @@
...
 using System;
+
+#if Lottie_Windows_WinUI3
+using Microsoft.UI.Composition;
+#else
 using Windows.UI.Composition;
+#endif
 
+#if Lottie_Windows_WinUI3
+namespace MicrosoftToolkit.WinUI.Lottie
```
michael-hawker, line 16:
```suggestion
namespace CommunityToolkit.WinUI.Lottie
```

source/Lottie/ImageAssetHandler.cs
```diff
@@ -5,17 +5,25 @@
...
+#else
 namespace Microsoft.Toolkit.Uwp.UI.Lottie
```
michael-hawker, line 18: If you want to do a 7.x release, probably want to keep the UWP namespace the same. If you want to jump on the 8.0 train early you could just use the same common namespace. (CommunityToolkit.WinUI.Lottie)

The main Windows Community Toolkit 8.0 release that's going to adopt this though probably won't come out until near the end of this year though. We're just setting up our new process in Toolkit Labs for Windows which we then have to adopt and merge all our Toolkit forks together to support UWP and WinUI in one codebase like you're doing here.

Lottie-Windows-WinUI3/Lottie-Windows-WinUI3.csproj
```diff
@@ -0,0 +1,40 @@
...
+		<TargetFramework>net5.0-windows10.0.18362.0</TargetFramework>
+
+		<Nullable>enable</Nullable>
+		<CsWinRTIncludes>Microsoft</CsWinRTIncludes>
+		<Platforms>x64;x86</Platforms>
+
+		<DefineConstants>Lottie_Windows_WinUI3</DefineConstants>
```
michael-hawker, line 12: We're adopting `WINAPPSDK` as our constant in Toolkit Labs for Windows, so if you want to align, you could change this.

michael-hawker (changes requested):
Couple of comments and info about where we're heading with the WCT.

Also not sure if you've done an audit for `Dispatcher` usage and `XamlRoot` for the general migration guidance. If things work in XAML Islands today, you should be fine. But always good to check if you missed a move to `DispatcherQueue`.

Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <Authors>Microsoft.Toolkit</Authors>
+    <Authors>CommunityToolkit</Authors>
```
michael-hawker, line 4: This should stay the same `Microsoft.Toolkit` is our NuGet identity still (just got caught in the mass rename).

aborziak-ms: Oh, ok. Fixed

README.md
```diff
@@ -24,7 +24,7 @@ This repo also contains source code for **[samples](/LottieViewer)**.
 ## Build Status
 | Target | Branch | Status | Recommended NuGet package |
 | ------ | ------ | ------ | ------ |
-| 7.1.0 release | main | [![Build Status](https://dev.azure.com/dotnet/WindowsCommunityToolkit/_apis/build/status/Microsoft.Toolkit.Uwp.UI.Lottie?branchName=main)](https://dev.azure.com/dotnet/WindowsCommunityToolkit/_build/latest?definitionId=61&branchName=main) | [![NuGet](https://img.shields.io/nuget/v/Microsoft.Toolkit.Uwp.UI.Lottie.svg)](https://www.nuget.org/packages/Microsoft.Toolkit.Uwp.UI.Lottie/) |
+| 7.1.0 release | main | [![Build Status](https://dev.azure.com/dotnet/WindowsCommunityToolkit/_apis/build/status/CommunityToolkit.WinUI.Lottie?branchName=main)](https://dev.azure.com/dotnet/WindowsCommunityToolkit/_build/latest?definitionId=61&branchName=main) | [![NuGet](https://img.shields.io/nuget/v/CommunityToolkit.WinUI.Lottie.svg)](https://www.nuget.org/packages/CommunityToolkit.WinUI.Lottie/) |
```
michael-hawker, line 27: This devops link may not change: https://dev.azure.com/dotnet/WindowsCommunityToolkit/_apis/build/status/Microsoft.Toolkit.Uwp.UI.Lottie?branchName=main

Lottie-Windows-WinUI3/Lottie-Windows-WinUI3.csproj
```diff
@@ -0,0 +1,40 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<OutputType>Library</OutputType>
+		<RootNamespace>CommunityToolkit.WinUI.Lottie</RootNamespace>
```
michael-hawker, line 5: `AssemblyName` missing here as well? (That's also the fallback for PackageId, so I'm not sure what happens without either.)

Lottie-Windows/Lottie-Windows.csproj
```diff
@@ -9,21 +9,21 @@
     <Description>This library provides the LottieVisualSource which is consumed by the Microsoft.UI.Xaml.Controls.AnimatedVisualPlayer to render Lottie JSON files.</Description>
     <PackageTags>UWP Toolkit Windows Animations Lottie XAML</PackageTags>
 
-    <RootNamespace>Microsoft.Toolkit.Uwp.UI.Lottie</RootNamespace>
-    <AssemblyName>Microsoft.Toolkit.Uwp.UI.Lottie</AssemblyName>
+    <RootNamespace>CommunityToolkit.WinUI.Lottie</RootNamespace>
+    <AssemblyName>CommunityToolkit.WinUI.Lottie</AssemblyName>
```
michael-hawker, line 13:
```suggestion
    <AssemblyName>CommunityToolkit.WinUI.Lottie</AssemblyName>
    <PackageId>CommunityToolkit.Uwp.Lottie</PackageId>
```

I think it's fine to have the Assembly Name match the namespace still, but we'll want to set the PackageId as otherwise the package names will conflict with the new WinUI 3 one.

Lottie-Windows-WinUI3/Lottie-Windows-WinUI3.csproj
aborziak-ms: done

PR #504, opened 2022-05-26
Refactoring

aborziak-ms:
Refactored Lottie-Windows Uwp and WinUI3 projects (csproj), moved common tags out to the Lottie-Windows.props file.
Moved LottieSampels folder out of samples folder for consistency with other projects.


raymond-wh-leung (approved):
:shipit:

PR #507, opened 2022-06-08
[LottieGen] Refactored effects, added cache for factories

aborziak-ms:
Lottie-Windows was previously generating one factory per each effect. I refactored the code a little bit so that we can reference same factory for multiple EffectBrushes which will reduce number of factories being created. OS does in fact have it's own cache for effects, so this change does not increase performance, just decreases generated code size.

Also small fix: fixed blur + shadow radius scaling, After Effects blur of radius `R` corresponds to Windows blur of radius `R / 3.33`

### DIFF
--- a/source/WinCompData/Mgce/CompositeEffect.cs
+++ b/source/WinCompData/Mgce/CompositeEffect.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Linq;
 using CommunityToolkit.WinUI.Lottie.WinCompData.Mgc;
 
 namespace CommunityToolkit.WinUI.Lottie.WinCompData.Mgce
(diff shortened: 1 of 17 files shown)


source/WinCompData/Mgce/CompositeEffect.cs
```diff
@@ -12,10 +13,56 @@ namespace CommunityToolkit.WinUI.Lottie.WinCompData.Mgce
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        public override int GetHashCode()
+        {
+            int hashCode = -1419366798;
```
raymond-wh-leung, line 62: curious why this value?

aborziak-ms: it was auto generated by Visual Studio so decided to keep it

raymond-wh-leung: Gotcha

raymond-wh-leung (approved):
:shipit:

PR #509, opened 2022-06-09
Fixed bezier curve IsLinear check

aborziak-ms:
Previous implementation had too many corner cases, for instance it was not working proper with zero vectors. Replaced it with vector-based check. Also added an epsilon so that if we are getting bezier curve that is really close to a line, we will translate it to a line, since linear animation is much cheaper in runtime.


raymond-wh-leung (approved):
:shipit:

PR #519, opened 2022-08-16
Use base64 string instead of array of bytes for CSharp code gen.

aborziak-ms:
This change should reduce size of embeded images that are exported as bytes arrays currently by a factor of 3, by exporting them as base64. For c#  only for now, because it is unclear which base64 API to use for winrt and cxx


raymond-wh-leung (approved):
:shipit:

PR #520, opened 2022-08-16
Fixed codegen to produce correct WinUI3 code

aborziak-ms:
There were some problems with generating code for WinUI3, a lot of namespaces were wrong (Microsoft vs Windows) and also some types are not available for WinUI3. Fixed so that LottieGen now work correctly for WinUI3 as well.


raymond-wh-leung (approved):
:shipit:

PR #524, opened 2022-11-16
Fixed IAnimatedVisual impl line for WinUI 2.8 (+fixed 2.7 -> 2.8)

aborziak-ms:
For CppWinRT we should declare all the interfaces that we want to implement. Previously we declared IAnimatedVisual2 without IAnimatedVisual which caused animations to not show up.
Also changed min version of WinUI from 2.7 to 2.8 for IAnimatedVisual2 because it became finally available only in 2.8.


raymond-wh-leung (approved):
:shipit:

PR #530, opened 2023-04-17
Updated Cake build system to newer version, and updated source to net7.0.

### LINKED ISSUE #478: WinUI3 support
Hi. I'm already using WinUI 3 in personal projects and production apps. And as I see - Lottie is not supported on this platform. There is no package `CommunityToolkit.WinUI.Lottie`.
And current package `Microsoft.UI.Xaml.UWP.Lottie` targets the UWP platform, that's why it's impossible to use Lottie in WinUI. Are there any plans for WinUI 3 support?

azchohfi:
This also fixes #478, as this makes sure we are shipping the MSIL version of CommunityToolkit.WinUI.Lottie, which works on all archs (x86, x64, and ARM64).
Only thing that I believe should be discussed here is the versioning. This is a breaking change only due to the bump in support for net7.0, and WAS 1.3.


aborziak-ms (commented):
Thanks for this change!

aborziak-ms (approved):

aborziak-ms:
> Only thing that I believe should be discussed here is the versioning. This is a breaking change only due to the bump in support for net7.0, and WAS 1.3.

I think we can release it as Lottie 8.0 if it is a breaking change, especially that it fixes quite a big issue.

Soap-141:
Thanks!

michael-hawker:
I think if we're only updating dependencies we don't have to do a major version bump, as per the [paragraph in SemVer here](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api):

> What should I do if I update my own dependencies without changing the public API?
That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. We would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

Upping to 8.0 could get confusing with all the package identity and namespace unification stuff we're doing for the rest of the WCT.

aborziak-ms:
> 

I see, ok. We did namespace migration for lottie as well, so I think it aligns with it anyway

michael-hawker:
> > 
> 
> I see, ok. We did namespace migration for lottie as well, so I think it aligns with it anyway

Oh, are both UWP and WinUI 3 packages using the same namespace now? Our current thinking and setup is [described here](https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/4488#issuecomment-1458638688) where we have two separate packages but a common namespace to aid in transitioning between UWP and WinUI 3 or building components that work on both from the same codebase (without needing forks).

We've been discussing if we move to '8.0' or switch to a more date-based schema per package as well. Right now we're just shipping things on our DevOps feeds as we continue to port things, so we have room to change decisions before we push to NuGet. I'll be posting more info on this in the next day or so and can share that for more background on where we've been at with the rest of the WCT.

PR #533, opened 2023-08-27
Fix wrong variable name with GenerateDependencyObject parameter

antoniohou:
According to source\UIDataCodeGen\CodeGen\CSharp\CSharpInstantiatorGenerator.cs line 376

```c#
            // Add fields for each of the theme properties.
            // Not needed if generating a DependencyObject - the values will be stored in DependencyPropertys.
            if (!SourceInfo.GenerateDependencyObject)
            {
                foreach (var prop in SourceInfo.SourceMetadata.PropertyBindings)
                {
                    var defaultValue = GetDefaultPropertyBindingValue(prop);

                    WriteInitializedField(builder, ExposedType(prop), $"_theme{prop.BindingName}", _s.VariableInitialization($"c_theme{prop.BindingName}"));
                }
            }
```

If used GenerateDependencyObject parameter there is no variable call "_theme{prop.BindingName}", when write the SetColorProperty function, it should change the variable name


antoniohou:
@dotnet-policy-service agree

aborziak-ms (approved):

PR #558, opened 2024-05-08
Add UIA support

adpa-ms:
Adds an automation provider for a Lottie Island + automation infrastructure so that it can be easily extended in the future.


jeffstall (approved):
:shipit:

PR #559, opened 2024-05-15
Update IsChildOf graph walk in LayersGraph.cs

getrou:
It looks like the recursive graph walk in LayersGraph.cs was intended to use the visited set of nodes as part of its traversal. In a quick test, I saw a Lottie json file load time go from 2 minutes down to 15 seconds.

Discovered by running the Visual Studio profiler over the Lottie Viewer

### DIFF
--- a/source/LottieData/Optimization/LayersGraph.cs
+++ b/source/LottieData/Optimization/LayersGraph.cs
@@ -11,6 +11,18 @@ namespace CommunityToolkit.WinUI.Lottie.LottieData.Optimization
     /// <summary>
     /// Represents directed acyclic graph of layer groups.
     /// Node1 is child of Node2 iff they have time ranges that intersect and Node1 goes after Node2 in z-order.
+    ///
+    ///                         +Z
+    /// |---------------------------------------------------|
+    ///      |--Node1--|
+    ///             |---Node2---|                              Time -->
+    ///                                |----Node3----|
+    /// |---------------------------------------------------|
+    ///                         -Z
+    ///
+    /// In this example Node1 is a child of Node2, but not of Node3.
+    /// Nodes can have multiple parents. Optimizations can be made to graphs that don't overlap in
+    /// time, which is often the case when a single Lottie file contains multiple animations.
     /// </summary>
 #if PUBLIC_LottieData
     public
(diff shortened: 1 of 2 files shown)


jeffstall:
>             /// If not null then this node was merged with another node and wee should consider them as one node now.

Fix spelling while we're here.

---
Refers to: source/LottieData/Optimization/LayersGraph.cs:30 in d0c6ca8. [](commit_id = d0c6ca822a1eb6f67162ffacbaf1e9c82f2a79ca, deletion_comment = False)

jeffstall:
> #if PUBLIC_LottieData

It would be helpful to have some text art to show what this looks like. It's a little trouble to visualize. It's making it difficult to understand if there's better optimizations so that we don't need to call IsChildOf() at all.

---
Refers to: source/LottieData/Optimization/LayersGraph.cs:15 in d0c6ca8. [](commit_id = d0c6ca822a1eb6f67162ffacbaf1e9c82f2a79ca, deletion_comment = False)

jeffstall (approved):
:shipit:

source/LottieData/Optimization/LayersGraph.cs
```diff
@@ -66,7 +66,7 @@ private bool IsChildOf(GraphNode node, HashSet<GraphNode> visited)
 
                 foreach (var parent in Parents)
                 {
-                    if (parent.Equals(node) || parent.IsChildOf(node))
+                    if (parent.Equals(node) || parent.IsChildOf(node, visited))
```
aborziak-ms, line 69: I think this is a correct fix for this line, but looking at lines above it seems that visited set should contain `this` nodes, instead of `node`'s. As you can see the parameter `node` is not changing, but `this` is what changes between the calls of this function. It looks like when I wrote this code I was writing it with regular DFS in mind like `dfs(current_node)`, but because of naming the function turned out to be `current_node->dfs(target_node)`. So it seems like code above should be changed to this:
```
if (visited.Contains(this))
{
    return false;
}
visited.Add(this);
```
Then it makes sense, as it is a cache of visited nodes, and `this` is the currently visited node.

jeffstall (approved):
Thanks for the updates!

aborziak-ms (approved):

PR #565, opened 2024-10-30
Fixed the version of WinUI3 in the command

HotCakeX:
Fixed the version of WinUI3 in the command. It needs to be `-WinUIVersion 3.0` instead of `-WinUIVersion 3`

<br>

image

image


HotCakeX:
@dotnet-policy-service agree

aborziak-ms (approved):

PR #566, opened 2024-12-19
Don't use fallback APIs based on OS version when running on WinAppSDK

getrou:
Lottie detects the version of the OS, and uses fallback codepaths when it detects that the needed API does not exist. This was good for UWP Lottie, but is unnecessary for WinAppSDK Lottie, since WinAppSDK bundles that code into the app and is not dependent on OS version.

Particularly, this was an issue when running on old OS versions that did not have proper pathing support. Lottie would detect the old OS and fall back to a type of pathing that does not work correctly, when that is unnecessary because WinAppSDK has the latest pathing updates.

This change hard-codes the contract version for WinAppSDK Lottie.


source/Lottie/Loader.cs
```diff
@@ -217,6 +217,11 @@ static IReadOnlyList<Issue> ToIssues(IEnumerable<TranslationIssue> issues)
         /// <returns>The highest UAP version supported by the current process.</returns>
         static uint GetCurrentUapVersion()
         {
+#if WINAPPSDK
```
jeffstall, line 220: Should the WinAppSDK version just not have this function? i.e., remove the call to GetCurrentUapVersion at the callsite instead?

jeffstall (approved):
Recommend a change, and still signed off.

source/Lottie/Loader.cs
getrou: Fixed!

aborziak-ms (approved):

PR #576, opened 2025-05-22
Switch UWPNET assembly name to CommunityToolkit.Uwp.Lottie

mink-ms:
If a consumer wants to use both UWPNET and WinUI3, the DLL names conflict since they're both "CommunityToolkit.WinUI.Lottie.dll". So rename the UWPNET one to CommunityToolkit.**Uwp**.Lottie.dll.

```<CsWinRTComponent>true</CsWinRTComponent>``` prevents the renaming, and it's not required for C# consumers so drop this as well.

### DIFF
--- a/Lottie-Windows/Lottie-Windows-Uwp/Lottie-Windows-Uwp.csproj
+++ b/Lottie-Windows/Lottie-Windows-Uwp/Lottie-Windows-Uwp.csproj
@@ -4,7 +4,7 @@
     <TargetFramework>uap10.0.16299</TargetFramework>
     <OutputType>winmdobj</OutputType>
 
-
+    <AssemblyName>CommunityToolkit.WinUI.Lottie</AssemblyName>
     <PackageId>CommunityToolkit.Uwp.Lottie</PackageId>
     <PackageTags>UWP Toolkit Windows Animations Lottie XAML</PackageTags>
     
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Lottie-Windows/Lottie-Windows-Uwp/Lottie-Windows-Uwp.csproj [hunk 1]
```diff
@@ -4,7 +4,7 @@
     <TargetFramework>uap10.0.16299</TargetFramework>
     <OutputType>winmdobj</OutputType>
 
-
+    <AssemblyName>CommunityToolkit.WinUI.Lottie</AssemblyName>
```
aborziak-ms, line 7: should it be `Uwp` here as well?

mink-ms: to match UwpNET yes, but this was already ```WinUI``` and I only really need the UwpNET one to have a different assembly name so I opted to keep this as-is. Could still switch to ```Uwp``` though

aborziak-ms: Ah I see, then it is ok, let's keep it like that for now, to not change the namespace for the package that is already there.

aborziak-ms (approved):

